PR #6, opened 2022-09-12
Named modules

lypwig:
Here is a PR for named modules, that can be used like this:

```py
from build123d import common as bc, generic as bg, part as bp, sketch as bs, line as bl
```

See vase example (file in this PR, screenshot below) for more details.

Screenshot_20220912_035333


gumyr:
As implemented, this change reverts f586174963ea54b4c761c1b60878b2d8b881671b which was seen very favourably by the community. As it was, and as it would be here, the user needs to know all of the modules that compose build123d which is a large burden to put on new users.

Is there something that can be done in the `__init__.py` file to achieve both styles of import? Is there a way to add something like?
```py
sketch = [
    "Rectangle",
    "Circle",
    ...,
]
```

PR #10, opened 2022-09-23
Support extrude next/last to multiple disjoint faces

fpq473:


gumyr:
It seems like the cq_warehouse `Face.projectToShape` would be useful here (https://github.com/gumyr/cq_warehouse/blob/main/src/cq_warehouse/extensions.py#L1520) - we have the shape, the direction and the face to project. The resulting list of faces (trim_faces) could be make into shells with the same code `trim_shells = Shell.makeShell(trim_faces).Shells()` where the `surface_index = surface_dirs.index(max(surface_dirs))` would change to a `surface_indices` which would select the faces pointing in the correct direction. The `trim_faces = trim_shells[surface_index].Faces()` would need to also change to exact faces from multiple shells.

By using project (which is an OCCT implementation) the result should be faster and better tested than a custom algorithm.

By the way, there is no need to import `List` from typing. `list` is recommended as of python 3.9.

gumyr:
This PR isn't valid anymore as there is new direct_api; however, the general ideas presented here are now implemented in `Solid.extrude_until` and `BuildPart.Extrude`.  The following test case illustrates this request:
```py
from build123d import *

stairs = Compound.make_compound(
    [
        Solid.make_box(10, 3.5, 0.5).locate(Location((0, o, o + 5)))
        for o in range(0, 12, 3)
    ]
)
to_extrude = Face.make_rect(12, 8).locate(Location((5, 6.25, 0)))

under = Solid.extrude_until(to_extrude, stairs, (0, 0, 1))

show_object(stairs, "stairs", options={"color": "blue", "alpha": 0.8})
show_object(to_extrude, "face to extrude")
show_object(under, "under", options={"color": "red", "alpha": 0.8})
```
image

Here, a simple rectangle is extruded into four disjoint "steps".

PR #27, opened 2022-10-10
Group by

bernhard-42:
Created an initial PR for `group_by`:

- I looked further into overloading boolean operators and would really vote against (hence, 2 of the commits remove the overloading of `>` and `<`. But of course the commit can be removed again). The bitwise operators are usually not used in CAD programs, so I think this could be OK and not so confusing.
- As you suggested, I have assigned `>>` and `<<` with `groupby`.

The approach works as discussed: `(b.faces() >> Axis.Z) >> Axis.X`

The code currently uses `round(key, tol_digits)` to properly group the rounded floats. Usually tolerances are give as `1e-6`, however, using `tol_digits` was much easier.

There are no tests at the moment, I first wanted this approach get reviewed.


bernhard-42:
I was playing around more with it. I think it was a mistake to redefine the operators now: the samples don't work any more. I guess we should revert the last three commits.

bernhard-42:
Adding my discord proposal about overloading

- sort_by: `b.faces() ^ Axis.Z` returning a list of faces
- sort_by reversed: not needed: Use `[-index]` for element access, else reversed() (returning a generator). Or define `~` as reversed: `~(b.faces() ^ Axis.Z)` (returning a list)
- sort_by min: not needed: Use `(b.faces() ^ Axis.Z)[0]`
- sort_by max: not needed: Use `(b.faces() ^ Axis.Z)[-1]`
- group_by min: `b.faces() << Axis.Z` returning list of faces
- group_by max: `b.faces() >> Axis.Z` returning list of faces
- filter_by_type: `b.faces() % GeomType.CIRCLE`
- filter_by_axis: `b.faces() | Axis.Z`

Minor changes, but completely avoids overloading the boolean operators

bernhard-42:
After having played more with build123d and the discussion in discord, I think the only change in this PR should be the addition of `group_by`. I really think it helps in selecting objects (especially after I understood the local var pattern as a substitute for CQ tags)

I would leave it to you to decide on whether and which operators overrides.

bernhard-42:
Given that you've changed a lot in the last merge, I will close this PR and raise a new one based on the new code base.
That's easier than trying to adapt ...

gumyr:
Yes, code is changing frequently.  I definitely want the `group_by` functionality and assigning it the ">>" operator seems reasonable (the examples will be changed to remove the sort operators anyway).  `sort_by` and `filter_by` (merging of `filter_by_axis` and `filter_by_type` - just like `sort_by`) would get operators (I'm inclined to go with `>` and `|` to somewhat match cadquery's string selectors).

PR #63, opened 2022-11-30
Add a Rotate method to Plane

jdegenstein:
Please review this as I am really not sure I did everything correctly. I am particularly unsure about returning self and the function annotation.

```
In [2]: a=Plane.XY

In [3]: a
Out[3]: Plane(o=(0.00, 0.00, 0.00), x=(1.00, 0.00, 0.00), z=(0.00, 0.00, 1.00))

In [4]: a.rotate((0,90,0))
Out[4]: Plane(o=(0.00, 0.00, 0.00), x=(0.00, 0.00, -1.00), z=(1.00, 0.00, 0.00))

In [5]: a
Out[5]: Plane(o=(0.00, 0.00, 0.00), x=(0.00, 0.00, -1.00), z=(1.00, 0.00, 0.00))

In [6]: Plane.XY.rotate((0,90,0))
Out[6]: Plane(o=(0.00, 0.00, 0.00), x=(0.00, 0.00, -1.00), z=(1.00, 0.00, 0.00))
```

### DIFF
--- a/docs/introductory_examples.rst
+++ b/docs/introductory_examples.rst
@@ -242,7 +242,7 @@ Additionally the '@' operator is used to simplify the line segment commands.
 ---------------------------------------------------
 
 Mirror can also be used with BuildPart (and BuildSketch) to mirror 3D objects. The ``Plane.offset()``
-method shifts the plane in the normal direction (+ve or -ve).
+method shifts the plane in the normal direction (positive or negative).
 
 .. image:: assets/general_ex16.svg
     :align: center
@@ -327,7 +327,7 @@ It is also possible to create a rotated workplane, building upon some of the con
 example with the ``rotated()`` method.
 
 GridLocations places 4 Circles on 4 points on this rotated workplane, and then the Circles are
-extruded in the "both" (+ve and -ve) normal direction.
+extruded in the "both" (positive and negative) normal direction.
 
 .. image:: assets/general_ex22.svg
     :align: center


bernhard-42:
Isn't that the same as `a = a.rotated(0, 90, 0)` which already exists?

jdegenstein:
Yes, it is the same as that, I discussed with gumyr and he seemed open to adding this (similar to what is in CQ). There may be a way of adding with fewer lines of code.

gumyr:
@jdegenstein, the code here is how I've been doing rotate but @bernhard-42 has a better way - here is his code from `Location.__init__`:
```py
            if isinstance(args[0], (Vector, tuple)):
                if isinstance(args[1], (Vector, tuple)):
                    rotation = [radians(a) for a in args[1]]
                    q = gp_Quaternion()
                    q.SetEulerAngles(gp_EulerSequence.gp_Intrinsic_XYZ, *rotation)
                    transform.SetRotation(q)
                elif isinstance(args[0], (Vector, tuple)) and isinstance(
                    args[1], (int, float)
                ):
                    angle = radians(args[1])
                    q = gp_Quaternion()
                    q.SetEulerAngles(gp_EulerSequence.gp_Intrinsic_XYZ, 0, 0, angle)
                    transform.SetRotation(q)

                # set translation part after setting rotation (if exists)
                transform.SetTranslationPart(Vector(args[0]).wrapped)
```
I think we should be consistent here and use this the gp_Quaternion solution with Intrinsic angles in both `Plane.rotate` and `Plane.rotated` - @bernhard-42 do you agree?

bernhard-42:
While Euler angles are often easier to read when used in `__repl__`, for the location calculations I usually use quaternions.
And, I like when the creation of the object and printing the object uses related methods (here `SetEulerAngles`and `GetEulerAngles`) This usually avoids to introduce misinterpretations.
So, yes, I personally would go quaternions

PR #69, opened 2022-12-01
added further initializers to Plane

bernhard-42:
This is not a finished PR, since we should discuss the implementation of Compound (keep it, drop it, change it)
Currently it supports sketches with one single face in order to get the correct direction of this single face. We could also use the underlying workplane of the sketch or drop Compounds

What do you think?

When we have agreed, I'll finish the PR and add tests and streamline class Workplanes (the logic moved to Plane inits now)

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -627,17 +627,7 @@ class Workplanes(WorkplaneList):
             if isinstance(obj, Plane):
                 self.workplanes.append(obj)
             elif isinstance(obj, (Location, Face)):
-                if isinstance(obj, Face):
-                    face = obj
-                    origin = face.center()
-                else:
-                    face = Face.make_rect(1, 1).move(obj)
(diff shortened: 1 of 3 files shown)


bernhard-42:
Decided to remove Compound from Plane initializers: it is not unique (should i use the workplane location or a singel face location - which could be a different one). If someone wants the workplane location of a compound, `Plane(compound.location)` is the easy solution

bernhard-42:
`Workplanes` is also changed and uses the Plane initializers for `Face` and `Location`

gumyr:
This all looks good to me.  I agree that `Plane(<Compound>)` doesn't need to be supported - BuildSketch is a very special case where the Faces are all on the same plane but that generally isn't true.  Is there anything more that you wanted to do here or is it ready to be merged?

bernhard-42:
Nothing I want to add. Feel free to merge.
Thanks!

PR #85, opened 2022-12-30
Fixes for cheat_sheet

jdegenstein:


gumyr:
Thanks for the fixes.  Note that sphinx requires that the spacing in the table to be perfect so I had to make some minor changes.

jdegenstein:
Ok, I will pay better attention to that in the future, thanks for fixing it

PR #90, opened 2023-01-05
min, max, min_group and max_group for ShapeList

bernhard-42:
As discussed.
Given the minimum footprint of this, I did not create test cases.


gumyr:
As per the discussion on Discord, it seems like having `first` and `last` properties are sufficient - at least for now.  So I'm going to close this PR without merging.  Thank you for the submission and the good discussion.

PR #115, opened 2023-02-03
add subtraction test that currently fails due to OCCT clean issues

jdegenstein:
```
FAIL: test_mode_subtract2 (test_build_part.BuildPartTests)
----------------------------------------------------------------------
  File "\sourcedir\tests\test_build_part.py", line 112, in test_mode_subtract2
    self.assertAlmostEqual(
AssertionError: 1.8846104014209444 != 3.874630939427411 within 0 places (1.9900205380064668 difference)
```
The part volume is supposed to be twice as high, or ~3.872. Note that there are also discrepancies between some of the volumes reported by build123d:
```
In [79]: test.part.volume
Out[79]: 1.8846104014209444

In [80]: test.part.volume*2
Out[80]: 3.7692208028418888

In [81]: test.solids()[1].volume
Out[81]: 1.9362315146088336

In [82]: test.solids()[1].volume*2
Out[82]: 3.872463029217667
```

### DIFF
--- a/tests/test_build_part.py
+++ b/tests/test_build_part.py
@@ -104,6 +104,16 @@ class BuildPartTests(unittest.TestCase):
         self.assertTrue(isinstance(test._obj, Compound))
         self.assertAlmostEqual(test.part.volume, 8000 - (4000 / 3) * pi, 5)
 
+    def test_mode_subtract2(self):
+        """Note that this is known to fail due to OCCT clean issues"""
+        with BuildPart() as test:
+            Sphere(1)
+            Box(0.1, 3, 3, mode=Mode.SUBTRACT)
+        self.assertTrue(isinstance(test._obj, Compound))
+        self.assertAlmostEqual(
+            test.part.volume, 4 / 3 * pi * 1**3 - pi * 1**2 * 0.1, 0
+        )
+
     def test_mode_intersect(self):
         """Note that a negative volume is created"""
         with BuildPart() as test:


gumyr:
May I suggest that this test be changed to be something like that following as the problem isn't clear from the test provided:

```py
sphere = Solid.make_sphere(1)
divider = Solid.make_box(0.1, 3, 3, Plane(origin=(-0.05, -1.5, -1.5)))
positive_half, negative_half = [s.clean() for s in sphere.cut(divider).solids()]
assert positive_half.volume != negative_half.volume
```
The problem appears to be the extra Edge on the sphere messing clean up (I've seen weird behaviour when using a sphere to dish the top of a key cap - if that weird Edge is rotated away the problem disappears). I don't think this problem is actually related to cut/subtract. Checking for not equal will allow the tests to run without errors but alert us to any changes in behaviour. The comment you've provided is essential here - this is a weird negative test.

image

gumyr:
I added the following test to `test_direct_api.py`:
```py
    def test_clean_error(self):
        """Note that this test is here to alert build123d to changes in bad OCCT clean behavior
        with spheres or hemispheres. The extra edge in a sphere seems to be the cause of this.
        """
        sphere = Solid.make_sphere(1)
        divider = Solid.make_box(0.1, 3, 3, Plane(origin=(-0.05, -1.5, -1.5)))
        positive_half, negative_half = [s.clean() for s in sphere.cut(divider).solids()]
        self.assertGreater(abs(positive_half.volume - negative_half.volume), 0, 1)
```
It passes now with the bad behaviour (so the tests run clean) but will fail if this bad behaviour ever changes.

PR #126, opened 2023-02-11
Add ci

hiaselhans:
This adds the CI with the following workflows:

- Mypy Type-checking
  - excluded dependencies
  - allow implicit None
  - This currently fails with a lot of type issues.
- unittests
  - uses pytest 

I also changed the dependency of cadquery-ocp to >= 7.6.3 because 7.7.0 is just around the corner...
changes in the egg-info resulted from installing. should we maybe exclude this directory?

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,20 @@
+name: 'Setup'
+inputs:
+  python-version:  # id of input
+    description: 'Python version'
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: install requirements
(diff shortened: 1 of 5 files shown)


gumyr:
Wow, this is great - thank you!  I upgraded to OCP 7.7.0 this morning which should resolve that question.  It seems like the docs have changed as well (removed the environment setup from install as you've done here) - so if you resync everything should be good to go.

Now that the pip install is with a pyproject.toml file, does there even need to be an `.egg-info` directory?

hiaselhans:
great, thank you!
The branch is rebased now.

To be honest, it's the first time that i see an egg-info being checked out in the source. I don't think it's needed.
We could just move the build123d directory to the root of the repository?

gumyr:
The egg-info probably got added by mistake but I'm not 100% what is required or not.

Issue #26, opened 2022-10-10
bad IDE integration


lypwig:
Some classes are incorrectly recognized as modules:

![image](https://user-images.githubusercontent.com/1665542/194806651-9c273e2f-31f6-4545-909b-7c8bb55f6746.png)

Same for enumerations:
![image](https://user-images.githubusercontent.com/1665542/194806766-921d98c0-11b3-4fd1-aa27-61991566bba2.png)

![image](https://user-images.githubusercontent.com/1665542/194808739-fedeb488-6c55-45f3-8929-9330688b0704.png)

I must type `bd.Mode.Mode` to have a correct auto-completion:
![image](https://user-images.githubusercontent.com/1665542/194806883-1af742b0-b42b-4d5f-b2ee-7aca37c7b049.png)

gumyr:
I know this generally isn't the best python practice but when imported as `from build123d import *` all of these problems go away (at least in my environment).  It will take some research to figure out why alternative imports cause this issue.

gumyr:
With commit 2b830ea472b267d41e3b6878d181ec679e2e6483, this problem seems to be solved. The issue was caused by importing from the top level of the project instead of from the appropriate module.

![image](https://user-images.githubusercontent.com/85638518/218333836-1255c933-3639-46a1-8492-30ed4841c4de.png)

![image](https://user-images.githubusercontent.com/85638518/218333849-15126d80-cbb0-4e5e-9505-892342d6f5c9.png)

![image](https://user-images.githubusercontent.com/85638518/218333885-182ab01b-6efd-4f3c-a2ab-fa71c81372ed.png)

PR #128, opened 2023-02-11
don't import from top-level

hiaselhans:
could be related to #26 

changes the imports to use the corresponding sub-packages

### DIFF
--- a/src/build123d/build_generic.py
+++ b/src/build123d/build_generic.py
@@ -48,14 +48,10 @@ from build123d.direct_api import (
     Solid,
     Axis,
 )
-from build123d import (
-    BuildLine,
-    BuildSketch,
-    BuildPart,
-    Builder,
-    LocationList,
-    WorkplaneList,
-)
+from build123d.build_line import BuildLine
+from build123d.build_sketch import BuildSketch
+from build123d.build_part import BuildPart
+from build123d.build_common import Builder, LocationList, WorkplaneList
 
 logging.getLogger("build123d").addHandler(logging.NullHandler())
 logger = logging.getLogger("build123d")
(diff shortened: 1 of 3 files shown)


jdegenstein:
I had the best results during testing with `from build123d import Shape` in `jupyter_tools.py` which affects the ipython console in CQ-editor. That being said, I could be wrong about it.

I was having some type checking issues when importing from the sub-packages where `isinstance(obj, Shape)` as returning False unexpectedly.

gumyr:
What is the reason behind not importing from the top level?  The changes seem equivalent to me but I'm missing something important I fear.

Thanks by the way!

hiaselhans:
> I had the best results during testing with `from build123d import Shape` in `jupyter_tools.py` which affects the ipython console in CQ-editor. That being said, I could be wrong about it.
> 
> I was having some type checking issues when importing from the sub-packages where `isinstance(obj, Shape)` as returning False unexpectedly.

That's a very weird bug indeed. With this PR being merged already, how does the latest dev behave in cq-editor? Is there a way we could write a test for it?



> What is the reason behind not importing from the top level? The changes seem equivalent to me but I'm missing something important I fear.
> 
> Thanks by the way!

It's not a hard requirement, but in my opinion it's easier to navigate and it avoids circular imports a bit.
Just thinking about an example module "abmodule" where you have file a defining class A and file b defining class B:

__init__.py:
```python
from abmodule.a import *
from abmodule.b import *
```

a.py:
```python
import B from abmodule
>>>ImportError: cannot import name 'B' from partially initialized module 'abmodule' (most likely due to a circular import)
```

Where `from abmodule.b import B` should still work

jdegenstein:
I just tested again against the latest build123d and (my fork of) CQ-editor and I am NOT seeing that type checking problem. I will keep an eye on it, as I recall the problem was occasionally intermittent.

PR #129, opened 2023-02-11
fix some typing issues

hiaselhans:
this is a first attempt to make mypy a little more happy.

There is a ton of mypy errors coming from the fact that for axis and planes there are "class-properties" defined (e.g. Axis.X). It looks like that is [deprecated in python 3.11](https://docs.python.org/3/whatsnew/3.11.html#language-builtins) relevant issue:  python/cpython#89519.

I think there are three ways to move forward with this:
- setting the values directly on the class (without property). The downside is that those values could be mutable and cause some troubles when modified by the user
- removing the property. Downside: added verbosity
- creating a custom getter class with `__call__()`.
- using some metaclass magic

I can take a look at how to implement one of the later two if changing the signature is not an option for you.

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -112,8 +112,8 @@ class Builder(ABC):
         self.workplanes = workplanes
         self._reset_tok = None
         self.builder_parent = None
-        self.last_vertices = []
-        self.last_edges = []
+        self.last_vertices: list[Vertex] = []
+        self.last_edges: list[Edge] = []
         self.workplanes_context = None
         self.active: bool = False  # is the builder context active
         self.exit_workplanes = None
@@ -156,20 +156,22 @@ class Builder(ABC):
 
         logger.info("Exiting %s", type(self).__name__)
 
-    @abstractstaticmethod
+    @staticmethod
+    @abstractmethod
     def _tag() -> str:
         """Class (possibly subclass) name"""
-        return NotImplementedError  # pragma: no cover
+        raise NotImplementedError  # pragma: no cover
 
+    @property
     @abstractmethod
     def _obj(self) -> Shape:
         """Object to pass to parent"""
-        return NotImplementedError  # pragma: no cover
+        raise NotImplementedError  # pragma: no cover
 
     @abstractmethod
     def _obj_name(self) -> str:
         """Name of object to pass to parent"""
-        return NotImplementedError  # pragma: no cover
+        raise NotImplementedError  # pragma: no cover
 
     @property
     def max_dimension(self) -> float:
@@ -221,7 +223,7 @@ class Builder(ABC):
         Returns:
             VertexList[Vertex]: Vertices extracted
         """
-        vertex_list = []
+        vertex_list: list[Vertex] = []
         if select == Select.ALL:
             for edge in self._obj.edges():
                 vertex_list.extend(edge.vertices())
@@ -297,7 +299,7 @@ class Builder(ABC):
             solid_list = self.last_solids
         return ShapeList(solid_list)
 
-    def validate_inputs(self, validating_class, objects: Shape = None):
+    def validate_inputs(self, validating_class, objects: Iterable[Shape] = None):
         """Validate that objects/operations and parameters apply"""
 
         if not objects:
@@ -441,7 +443,7 @@ class HexLocations(LocationList):
         self.align = align
 
         # Generate the raw coordinates relative to bottom left point
-        points = ShapeList()
+        points = ShapeList[Vector]()
         for x_val in range(0, x_count, 2):
             for y_val in range(y_count):
                 points.append(
@@ -614,7 +616,7 @@ class GridLocations(LocationList):
         align_offset = []
         for i in range(2):
             if align[i] == Align.MIN:
-                align_offset.append(0)
+                align_offset.append(0.)
             elif align[i] == Align.CENTER:
                 align_offset.append(-size[i] / 2)
             elif align[i] == Align.MAX:
@@ -633,7 +635,7 @@ class GridLocations(LocationList):
             )
 
         self.local_locations = Locations._move_to_existing(local_locations)
-        self.planes = []
+        self.planes: list[Plane] = []
         super().__init__(self.local_locations)
 
 
@@ -701,7 +703,7 @@ class WorkplaneList:
     @classmethod
     def localize(
         cls, *points: VectorLike
-    ) -> Union[list[list(Vector)], list[Vector], Vector]:
+    ) -> Union[list[list[Vector]], list[Vector], Vector]:
         """Localize a sequence of points to the active workplanes
 
         The return value is conditional:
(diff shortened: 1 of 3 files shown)


src/build123d/build_common.py
```diff
@@ -701,7 +703,7 @@ def _get_context(cls):
     @classmethod
     def localize(
         cls, *points: VectorLike
-    ) -> Union[list[list(Vector)], list[Vector], Vector]:
+    ) -> Union[list[list[Vector]], list[Vector], Vector]:
```
hiaselhans, line 706: this output type looks hard to use.. :)

src/build123d/build_generic.py
```diff
@@ -94,13 +94,14 @@ def __init__(
...
-            rotate = (
-                Rotation(*rotation_value)
-                if isinstance(rotation_value, tuple)
-                else rotation
-            )
-            objects = [obj.moved(rotate) for obj in objects]
+            if rotation is None:
+                rotation = Rotation(0, 0, 0)
+            elif isinstance(rotation, float):
+                raise NotImplementedError()
```
hiaselhans: please correct me if i am wrong, but it looks like the the float option wasn't handled in the prior version too.
obj.moved takes a Location object and uses Location.wrapped which would not exist on floats. What should the desired behavior be here?

src/build123d/build_common.py
```diff
@@ -277,6 +278,7 @@ def faces(self, select: Select = Select.ALL) -> ShapeList[Face]:
         if select == Select.ALL:
             face_list = self._obj.faces()
         elif select == Select.LAST:
+            # TODO: move to dedicated builder?
```
hiaselhans: problem here is that this class does not have a last_faces and last_solids attribute

src/build123d/build_common.py
```diff
@@ -112,8 +112,8 @@ def __init__(
         self.workplanes = workplanes
         self._reset_tok = None
         self.builder_parent = None
-        self.last_vertices = []
-        self.last_edges = []
+        self.last_vertices: list[Vertex] = []
+        self.last_edges: list[Vertex] = []
```
gumyr: This is a list of Edge: list[Edge]

src/build123d/direct_api.py
```diff
@@ -4842,7 +4843,7 @@ def intersect(self, *to_intersect: Shape) -> Compound:
         return tcast(Compound, self._bool_op(self, to_intersect, intersect_op))
 
     def get_type(
-        self, obj_type: Union[Edge, Face, Shell, Solid, Wire]
+        self, obj_type: Type[Edge] | Type[Face] | Type[Shell] | Type[Solid] | Type[Wire]
```
gumyr: Is this supported in Python 3.9?  I want to support 3.9 at least for a while.

src/build123d/build_common.py
```diff
@@ -241,10 +243,9 @@ def edges(self, select: Select = Select.ALL) -> ShapeList[Edge]:
             ShapeList[Edge]: Edges extracted
         """
         if select == Select.ALL:
-            edge_list = self._obj.edges()
+            return self._obj.edges()
```
gumyr: IIRC pylint doesn't like multiple returns, that's why there's an intermediate variable.

src/build123d/build_common.py
gumyr: True but some of the derived classes do.  Those that don't override this method with one that raises NotImplementedErrror. IIRC this was done to help with consistency in documentation.

src/build123d.egg-info/PKG-INFO
```diff
@@ -19,6 +19,8 @@ Build123d could be considered as an evolution of [CadQuery](https://cadquery.rea
 
 The documentation for **build123d** can found at [readthedocs](https://build123d.readthedocs.io/en/latest/index.html).
 
+There is a [***Discord***](https://discord.com/invite/Bj9AQPsCfx) server (shared with CadQuery) where you ask for help in the development channel.
+
 To install **build123d** from github:
 ```
 python3 -m pip install git+https://github.com/gumyr/build123d.git@pyproject.toml
```
gumyr: The install can be shortened: `python3 -m pip install git+https://github.com/gumyr/build123d`

gumyr: As the .egg-info isn't in GitHub - I guess this change isn't really useful anymore.

src/build123d/build_common.py
```diff
@@ -156,20 +156,22 @@ def __exit__(self, exception_type, exception_value, traceback):
 
         logger.info("Exiting %s", type(self).__name__)
 
-    @abstractstaticmethod
```
gumyr, line 159: I see abstractstaticmethod has be deprecated - thanks for pointing that out.

src/build123d/build_common.py
```diff
@@ -156,20 +156,22 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+    @staticmethod
+    @abstractmethod
     def _tag() -> str:
         """Class (possibly subclass) name"""
-        return NotImplementedError  # pragma: no cover
+        raise NotImplementedError  # pragma: no cover
```
gumyr, line 163: Thanks, I thought I fixed these - so irritating!

src/build123d/build_generic.py
gumyr: The float option only applies to the BuildSketch section below. In general there isn't much input validation done in any of the APIs but there probably should be a check here, something like:
```py
            if rotation is None:
                rotation = Rotation(0, 0, 0)
            elif isinstance(rotation, float):
                raise ValueError("Float values of rotation are not valid for BuildPart")

            rotate = (
                Rotation(*rotation)
                if isinstance(rotation_value, tuple)
                else rotation
            )
            objects = [obj.moved(rotate) for obj in objects]
```

src/build123d/direct_api.py
```diff
@@ -3618,7 +3619,7 @@ def project_text(
 
 
 # This TypeVar allows IDEs to see the type of objects within the ShapeList
-T = TypeVar("T", bound=Shape)
+T = TypeVar("T", Shape, Vector)
```
gumyr, line 3631: This is all magic to me, it was better than nothing but didn't work perfectly - hopefully this resolves it.

gumyr (changes requested):
Thanks for doing all this work - I really appreciate you making the code better.

All of the .egg-info files have been removed from GitHub.

gumyr:
The loss of class properties would be a very negative change for this project. The ability to use `Axis.X` and `Plane.YZ` is very valuable and expresses the concepts in basically mathematical notation. CadQuery used "named" planes like `Plane.named("XY")` which kinda sucked. If you have any ideas of how to maintain this functionality it would be greatly appreciated.

Jojain:
@gumyr, If the decorator in the stdlib is deprecated nothing stops you to make your own version in b123d and not rely on the python one;

src/build123d/build_common.py
hiaselhans: thanks!

gumyr:
@hiaselhans Another request if you don't mind: would you mind breaking the unittests and static type checking into two separate workflows? This way there can be a badge for each one - most folks are only going to be interested in the tests passing.

(I'd normally do this myself but I don't want to cause more merge issues)

src/build123d/direct_api.py
hiaselhans: true, i guess that came in python 3.10

hiaselhans:
@gumyr thanks for the prompt replies! I enjoy being able to work on the tools i use :)

did a squash and addressed your issues.
That workflow split is coming as a separate pr.

PR #133, opened 2023-02-13
build123d.*Joint: Make all joints reversibly constructable

ezrec:
Joint: 
- Add an abstract helper function 'relative_to()' that returns the relative position of 'other''s joint to self.
- Make 'connect_to()` concrete.

*Joint: Rework to provide 'relative_to()' instead of overriding 'connect_to()'

### DIFF
--- a/src/build123d/direct_api.py
+++ b/src/build123d/direct_api.py
@@ -7771,11 +7771,25 @@ class Joint(ABC):
         self.parent = parent
         self.connected_to: Joint = None
 
-    # pylint doesn't see this as an abstract method and warns about different arguments in
-    # derived classes
-    @abstractmethod
     def connect_to(self, other: Joint, *args, **kwargs):  # pragma: no cover
         """Connect Joint self by repositioning other"""
+
+        if not isinstance(other, Joint):
+            raise TypeError(f"other must of type Joint not {type(other)}")
+
+        relative_location = None
+        try:
+            relative_location = self.relative_to(other, *args, **kwargs)
+        except TypeError:
+            relative_location = other.relative_to(self, *args, **kwargs).inverse()
+
+        other.parent.locate(self.parent.location * relative_location)
+
+        self.connected_to = other
+
+    @abstractmethod
+    def relative_to(self, other:Joint, *args, **kwargs) -> Location:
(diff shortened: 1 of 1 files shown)


src/build123d/direct_api.py
```diff
@@ -7771,11 +7771,25 @@ def __init__(self, label: str, parent: Union[Solid, Compound]):
...
-    # derived classes
-    @abstractmethod
     def connect_to(self, other: Joint, *args, **kwargs):  # pragma: no cover
         """Connect Joint self by repositioning other"""
+
+        if not isinstance(other, Joint):
+            raise TypeError(f"other must of type Joint not {type(other)}")
+
+        relative_location = None
+        try:
```
gumyr, line 7781: My only concern is here.  Auto flipping the order of the connection seems like it will potentially cause confusion for the user.  How about just raising an error and suggesting to the user that they try flipping the order manually?

gumyr (changes requested):
I like what you've done here - thank you.  With the exception of the one comment about not auto flipping the joint, I think it's good to go.

src/build123d/direct_api.py
ezrec: 'Auto flipping the connection' actually ends up doing _exactly_ what the user intends.

For example, if I was to connect:

```
chair_leg.joints["rigid_mount"].connect_to(roller_ball.joints["ball_mount"])
```

The RigidJoint would reject the BallJoint as an 'other' in it's `.relative_to()` method. The `connect_to()` then reverses the connection attempt (which succeeds - `BallJoint().relative_to(RigiJoint())`), then moves `other` based on the _inverse_ of that relative location.

This is _logically_ the same as the user intent (connect chair leg to ball mount), and the `chair_leg.connected_to` is now `roller_ball`, as was the user intent.

gumyr: I'm concerned about follow on consequences of flipping the connection. If one follows the chain of joints from the root object to the leaf object wouldn't there be a "kink" in the middle that could cause problems?  I'd need to try this out to be sure.

However, I think I'm going to accept it as is - if there are problems they can be fixed later.

gumyr:
Thanks for the PR @ezrec !

PR #134, opened 2023-02-14
clean up typing errors (pt.2)

hiaselhans:
This is part 2 to improve on typing issues.

### classproperty

~~This PR adds a `classproperty` class which is used for Plane.XY / Axis.X properties.~~
However, I am not very happy with this solution for these reasons:

- typing can't really be done well (but it works and detects the proper type)
- Given that a new instance is created in the property, a simple classmethod `Plane.XY()` would make that more obvious and clear.
- Docstrings aren't great:
``` python
 |  ----------------------------------------------------------------------
 |  Data and other attributes defined here:
 |  
 |  XY = Plane(o=(0.00, 0.00, 0.00), x=(1.00, 0.00, 0.00), z=(0.00, 0.00, ...
 |  
```

### direct_api.py

There are still a lot of typing issues in direct_api.py.
That is a really really large file and I think it's inherited from cadquery? Do you think I can go ahead and split it up in smaller chunks?

### localize

And then there is this:
https://github.com/gumyr/build123d/blob/c40a53782dae3d5f5ebf7a0162bde69075f7b8fb/src/build123d/build_common.py#L694-L723

In my experience this is impossible to type and it also adds a lot of special cases to handle it right.
I suggest that at least we ignore the special case of having just one workplane active and have a function signature like that:

``` python
VectorInput = TypeVar("VectorInput", bound=Union[list[VectorLike], VectorLike])
def localize(pts: VectorInput) -> list[VectorInput]:
   ...
```

I think that would improve a lot on readability and maintainability. And it would solve around 100 mypy issues at once.


src/build123d/build_line.py
```diff
@@ -748,28 +734,26 @@ class Spline(Edge):
     def __init__(
         self,
         *pts: VectorLike,
-        tangents: Iterable[VectorLike] = None,
-        tangent_scalars: Iterable[float] = None,
+        tangents: Collection[VectorLike] = None,
```
Jojain, line 737: I'm not a typing expert but isn't `Collection` more restrictive than `Iterable` ?

src/build123d/build_common.py
```diff
@@ -180,22 +189,30 @@ def _add_to_context(
     ):
         """Integrate a sequence of objects into existing builder object"""
         return NotImplementedError  # pragma: no cover
-
+    
     @classmethod
-    def _get_context(cls, caller=None):
-        """Return the instance of the current builder
+    def _get_context_or_raise(cls, caller=None) -> Self:
```
Jojain, line 194: Is there a reason to rename this method other than more lisibility ? I personally find it a bit weird to add `_or_raise` in the method name

gumyr:
@hiaselhans it would be great if joined the Discord so we could discuss this stuff more interactively (there is a link on the README).

While I appreciate all the work you're doing here, this PR is way too big.  Although `mypy` is cool and something that seems like it would add a lot of value, after seeing all of the difficulties and changes required to get it to be happy I'm not convinced that this is the best place to put (your) and my time.  Getting the test coverage to at or near 100% is a much higher priority as well as development of a few other features required before creating a release (hopefully in the next couple months).

With respect to the size of `direct_api.py`, I would like to find a way to split this file up but I was unable due to circular imports.  You'll find that the classes all depend on each other which left me with no alternative to this enormous file.  If you have any suggestions as to how to split the file up I'd love to hear them.

With respect to the changes to the class properties, pylint doesn't like the solution you proposed:
```
src/build123d/direct_api.py:3987:4: E0213: Method 'XY' should have "self" as first argument (no-self-argument)
```
more work needs to be done here.

I agree that the `localize` method is a bit weird.  It's just used internally so I should really take a second look and simplify it.

I know this is a very negative review - I'm sorry about that. It's just I'm one guy and having such large changes made to my code for a goal I'm not 100% behind is just too much at this point.

src/build123d/build_line.py
hiaselhans: Yes indeed Collection is more restrictive in that it provides a __len__() dunder method that is used a few lines below.

src/build123d/build_common.py
hiaselhans: the reason is that in `__enter__` _get_context is used to set the parent with an Optional result, while in the rest of the codebase the _get_context function is used as if it could never return None

hiaselhans:
> @hiaselhans it would be great if joined the Discord so we could discuss this stuff more interactively (there is a link on the README).
> 
> While I appreciate all the work you're doing here, this PR is way too big. Although `mypy` is cool and something that seems like it would add a lot of value, after seeing all of the difficulties and changes required to get it to be happy I'm not convinced that this is the best place to put (your) and my time. Getting the test coverage to at or near 100% is a much higher priority as well as development of a few other features required before creating a release (hopefully in the next couple months).
> 
> With respect to the size of `direct_api.py`, I would like to find a way to split this file up but I was unable due to circular imports. You'll find that the classes all depend on each other which left me with no alternative to this enormous file. If you have any suggestions as to how to split the file up I'd love to hear them.
> 
> With respect to the changes to the class properties, pylint doesn't like the solution you proposed:
> 
> ```
> src/build123d/direct_api.py:3987:4: E0213: Method 'XY' should have "self" as first argument (no-self-argument)
> ```
> 
> more work needs to be done here.
> 
> I agree that the `localize` method is a bit weird. It's just used internally so I should really take a second look and simplify it.
> 
> I know this is a very negative review - I'm sorry about that. It's just I'm one guy and having such large changes made to my code for a goal I'm not 100% behind is just too much at this point.

I appreciate your time to review my PR's and i can definitely split this up into pieces, but I don't want to give up so fast:

Typing done right can prevent a lot of bugs.
So when you make me choose between pylint and a static type checker, i would always choose the latter. Can you please elaborate a bit on the difficulties and changes?
Here are two bugfixes from this PR that mypy revealed:
https://github.com/gumyr/build123d/blob/c40a53782dae3d5f5ebf7a0162bde69075f7b8fb/src/build123d/build_sketch.py#L255

https://github.com/gumyr/build123d/blob/c40a53782dae3d5f5ebf7a0162bde69075f7b8fb/src/build123d/build_sketch.py#L281

Regarding the class properties:
I made my arguments against them, and it's not my fault they won't be supported in python 3.11+. i can rename the arguments to "self", but thats even more misleading and what is one line in a 200lines pylint output? I think the best solution is to use plain classmethods.

gumyr:
It's not that I'm against typing or typed languages, I've been using them for something like 40 years now.  My concern is the time I'd need to spend to fully review and understand the implication of all the changes you've made and the changes it seems must be made purely to keep the tool happy. Although, the `mode=mode` bug is a good find (setting mode is clearly not being tested very well) I'm almost 100% sure there are real functional bugs in the actual CAD operations that are still to be found. mypy is a priority for me, just not the top priority.

With respect to the class properties, I'm not blaming you for anything, in-fact I'm thanking you for bring this problem to my attention. But this is a future problem while there are many today problems to fix.

hiaselhans:
Clearly your point on years of experience. And I don't want to just come across and mess around with your code or complain about anything. On the contrary I am grateful that you created this library and I'd like to help.

I removed the part concerning the class properties, so all that is left is mostly fixes to existing type annotations.
It should be doable in 10-15minutes. If you want me to strip it down further i can split it up.
But I see that you already disabled the type checking workflow and my guess is that you'd rather not merge these changes at all?
If so, that's indeed quite a bummer for me. My experience is that proper typing does a great help with IDE's and gets me moving faster.

gumyr:
This PR is still too big. Many changes are not necessary, like changing method and variable names that just makes it more scary.  Let's leave this change for now - at some point in the future I'll take it up again and get mypy to pass.

I disabled the static type checking workflow as I received something like 40 email notifications yesterday and it looks like it's not possible to turn off notifications for a single workflow.  

My experience and that of the b3d users is that the IDE integration is good as is - other than the one caused by top level imports I've heard no complaints.

gumyr:
I've tried to run mypy and have fixed a few things but I won't be using it until it matures.  This is the error that decided it for me:
```
build_line.py:596: error: Argument 1 to "radians" has incompatible type "Optional[float]"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
```
It thinks passing a `float` into the `math.radians` method is an error.  Sorry but no...

hiaselhans:
Hi @gumyr .
Are you sure you got that right?

https://github.com/gumyr/build123d/blob/e14379ec21e5e9270b22aa246b142c9912037c97/src/build123d/build_line.py#L585-L596

angle is Optional[float] just like mypy says. So that's indeed a good catch by mypy.

``` python
>>> import math
>>> math.radians(None)
  File "<stdin>", line 1, in <module>
TypeError: must be real number, not NoneType
>>> 

```

gumyr:
@hiaselhans you are correct - this is a real problem and I have fixed it - thank you for being persistent and pointing this out.

For now I still won't be using mypy widely.  Here is an example of why:
```py
"""Test of mypy handling of Union'd outputs"""
from typing import Union


def two_times(*values: float) -> Union[float, list[float]]:
    """Return either a single value or list of values"""
    values_times_two = [value * 2.0 for value in values]
    if len(values_times_two) == 1:
        return_value = values_times_two[0]
    else:
        return_value = values_times_two
    return return_value


print(two_times(1))
print(two_times(1, 2))
```
```
2.0
[2.0, 4.0]
```
When I run `pylint` on this code I get a score of 10.0/10.0.  However mypy gives:
```
mypy_test.py:11: error: Incompatible types in assignment (expression has type "List[float]", variable has type "float")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

Here is an alternative version:
```py
@overload
def alternate_two_times(value: float) -> float:
    """A single input"""


@overload
def alternate_two_times(*value: float) -> list[float]:
    """multiple inputs"""


def alternate_two_times(*args):
    """Return either a single value or list of values"""
    values_times_two = [value * 2.0 for value in args]
    if len(values_times_two) == 1:
        return_value = values_times_two[0]
    else:
        return_value = values_times_two
    return return_value
```
which gives:
```
mypy_test.py:16: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]
```

If this function only returned `list[float]`, the user would have to write `x = two_times(1)[0]` which I think is inferior code.

The problem with class properties is another major headache.  It looks like this will have to be changed sometimes in the future but is perfectly valid in 3.9 and 3.10 yet causes mypy to produce a large number of errors.  I found that I could turn of the requirement for explicit `Optional` off so that isn't an issue for me anymore (although I see that the recommendation is to force users to explicitly use `Optional`, this make the code worse in my opinion and I will not be doing this). 

Do you think it's possible to create a mypy.ini such that only "real" problems are found?

PR #136, opened 2023-02-15
Adding notes about pip upgrade

bgolinvaux:
Following [our discussion](https://discord.com/channels/964330484911972403/964330484911972406/1075449958649569361) on Discord, I have added a note about upgrading pip, at both locations where that made sense.

I could have put a single big note on top of the page but I am afraid that some people only read the specific instructions they need. Please tell me if you think the duplicate note is not pretty.


gumyr:
Thank you!

PR #143, opened 2023-02-20
Add dev (SCM) versioning to pip install

jdegenstein:
CadQuery installed from pip+git gives a version tag:
```
>>> import cadquery as cq
>>> cq.__version__
'2.2.1.dev8+g6c8030d'
```
This page discusses more about how to set this up:
https://github.com/pypa/setuptools_scm

It would be nice to have in build123d too. This requires PEP518.


pyproject.toml
```diff
@@ -2,12 +2,15 @@
 requires = [
     "setuptools>=42",
     "wheel",
+    "setuptools>=45",
```
gumyr, line 4: Should setuptools be listed twice here?

gumyr (changes requested):

pyproject.toml
jdegenstein: No it should not

jdegenstein:
The tests are failing due to missing `_version.py`, I added it but I am not sure where else it is supposed to be.

src/build123d/_version.py
```diff
@@ -0,0 +1 @@
+
```
gumyr: According to https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package, there should be a `__version__ = "0.0.1"` here. Could this be why the builds are failing?

gumyr (changes requested):

src/build123d/_version.py
jdegenstein: added __version__ = "0.0.1" as requested, and imported in `init.py`

PR #144, opened 2023-02-22
setuptools_scm versioning to pip install

jdegenstein:
Decided to delete the previous PR, as it had a lot of incorrect changes. I tested this PR locally and it worked properly.

The only thing I am not sure of was if this will be properly ignored by tests.

The goal of this is to add something similar to what CadQuery has:
```py
>>> import cadquery as cq
>>> cq.__version__
'2.2.1.dev8+g6c8030d'
```

### DIFF
--- a/src/build123d/__init__.py
+++ b/src/build123d/__init__.py
@@ -17,7 +17,6 @@ __all__ = [
     "Align",
     "AngularDirection",
     "CenterOf",
-    "Direction",
     "FontStyle",
     "FrameMethod",
     "GeomType",
--- a/src/build123d/__init__.py
+++ b/src/build123d/__init__.py
@@ -1,4 +1,4 @@
-from ._version import __version__
+from .version import version as __version__
 from build123d.build_common import *
 from build123d.build_line import *
 from build123d.build_sketch import *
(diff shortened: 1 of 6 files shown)


src/build123d/__init__.py
```diff
@@ -16,6 +17,7 @@
     "Align",
     "AngularDirection",
     "CenterOf",
+    "Direction",
```
gumyr: The enum `Direction` has been eliminated, why is it added back here? It causes the tests to fail.

gumyr (changes requested):

src/build123d/__init__.py
jdegenstein: Apologies, this must have been carried over from an older version. I have removed it again and it looks like the tests are doing better again.

jdegenstein:
Here's a demo of this actually working:
```
$ pip install git+https://github.com/jdegenstein/build123d@scm_versioning
...
```
```
$ python
Python 3.9.15 
>>> import build123d
>>> build123d.__version__
'0.1.dev494+g04b14c7'
```

gumyr:
Thank you @jdegenstein!

PR #147, opened 2023-02-23
Exclude package using full name

fpq473:
@jdegenstein As discussed in chat.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
 [tool.setuptools.packages.find]
 where = ["src"]
 # exclude build123d._dev from wheels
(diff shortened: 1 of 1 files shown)


gumyr:
@fpq473  and @jdegenstein Now that I've merged this change I get the following warning:
```
/home/roger/Documents/build123d/src/build123d/version.py:9: UserWarning: could not determine build123d package version; this indicates a broken installation
```
Should I revert the change?

gumyr:
Never mind. I reinstalled and this error went away.

jdegenstein:
Let us know if you see any more weirdness, I would like this to be able to handle most (if not all) methods of installation

PR #148, opened 2023-02-23
PolarLocations does not respect rotation for start_angle other than zero

jdegenstein:
I believe this is correct, but will create an example case to demonstrate. I noticed that PolarLocations did not properly respect rotation for a start_angle other than zero.

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -512,7 +512,7 @@ class PolarLocations(LocationList):
                 Location(
                     Vector(radius, 0).rotate(Axis.Z, start_angle + angle_step * i),
                     Vector(0, 0, 1),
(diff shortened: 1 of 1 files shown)


jdegenstein:
Consider the following code:
```py
from build123d import *

with BuildPart(Plane(origin=(0,0,0),x_dir=(1,.5,0))) as cp:
    Box(9,9,9)

show_object(cp.part,options=rand_color())

with BuildPart() as mainp:
    with PolarLocations(radius=25,count=2,start_angle=30,stop_angle=360+30):
        Add(cp.part)

show_object(mainp.part)
```
Before this PR the rotation is NOT properly modified by start_angle (green and yellow boxes are at same angle):
image

With this PR the rotation IS properly modified by start_angle:
image

PR #150, opened 2023-02-26
Added notes about poetry installs

bgolinvaux:
Many versions of poetry currently installed contain a bug that makes them checkout the `master` branch by default.

I added a note in the docs that advises users to suffix their poetry commands with `@dev` when required.


gumyr:
Thank you!

PR #170, opened 2023-03-11
improve `Vector.add/sub` monkey-patching

snoyer:
Currently the monkey-patching of `Vector.add` and `Vector.sub` is a complete re-implementation of the original operators.
This lead to either code and code-path duplication, or dead code as the monkey-patching is not conditional and happens even when not using the builder API (so the original code in `Vector` never runs).

This proposal uses a wrapper function instead, it will localize the operand as needed and then call the original operator.

A minor implementation detail is to `try` and `localize` without checking the context beforehand, handling the no-context case with `except` (thus avoiding calling `_get_context()` twice). That mechanism could probably be generalized in other places by raising a custom error when no context is found.

A possible further improvement would be to avoid monkey-patching altogether if only the direct API is being used but I don't have a solution for that right now.


src/build123d/build_common.py
```diff
@@ -776,33 +776,19 @@ def __init__(self, *objs: Union[Face, Plane, Location]):
...
-        raise ValueError("Only Vectors or tuples can be subtracted from Vectors")
-
-    return result
-
-
-Vector.add = _vector_add  # type: ignore[assignment]
-Vector.sub = _vector_sub  # type: ignore[assignment]
+
+def _vector_add_sub_wrapper(original_op: Callable[[Vector, VectorLike], Vector]):
+    def wrapped(self: Vector, vec: VectorLike):
```
gumyr: Could a different name be used here as `wrapped` is the name of a Vector property?  Otherwise, looks good.

gumyr (changes requested):

src/build123d/build_common.py
snoyer: I don't think the name matters, does it show up anywhere? Either way I can change it, any suggestion?

gumyr: How about `wrapper`?

gumyr:
Thank you

PR #172, opened 2023-03-11
Direct API tests cleanup

snoyer:
Some tidying up of `test_direct_api.py`:
- removed `*` imports
- remove monkey-patching by subclassing
- added `assertVectorAlmostEquals` to compare `Vector`s and remove a bunch of `.to_tuple()`
- simplified `show_topology` tests by refactoring the boilerplate

Hopefully the noise reduction can make adding more tests in the future a bit better.

Should the tests involving file io use temporary files instead of hardcoded filenames in the current directory?

### DIFF
--- a/tests/test_direct_api.py
+++ b/tests/test_direct_api.py
@@ -243,9 +243,7 @@ class TestAxis(DirectApiTestCase):
         )
 
     def test_axis_reverse(self):
-        self.assertTupleAlmostEquals(
-            Axis.X.reverse().direction.to_tuple(), (-1, 0, 0), 5
-        )
+        self.assertVectorAlmostEquals(Axis.X.reverse().direction, (-1, 0, 0), 5)
 
     def test_axis_reverse_op(self):
         axis = -Axis.X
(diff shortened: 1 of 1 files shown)


gumyr:
Thank you for that!  Test cases tend to not get too much love so this is appreciated.

> Should the tests involving file io use temporary files instead of hardcoded filenames in the current directory?

The tests run on three different OS's so messing around with temporary files seemed like unnecessary work.  However, I've done this before:
```py
import tempfile

    temp_directory = tempfile.TemporaryDirectory()
    temp_directory_path = temp_directory.name
    original_directory_path = os.path.join(temp_directory_path, "original")
```
for a script that was creating patches and it worked on all three OSs.

snoyer:
Cleanest/safest you can go I believe is:
```py
with tempfile.NamedTemporaryFile('w', suffix='.svg') as f:
    f.write('<svg><path d="M 0,1 L 2,3"/></svg>')
    f.flush()
    result = import_svg(f.name)
```
If you want I can refactor this way and throw it at the CI to see how bad that goes. Let me know and I can to it tomorrow.

PR #177, opened 2023-03-14
callable predicates for `filter_by` and `group_by`

snoyer:
Proposal to add callable predicates to `ShapeList`'s `filter_by` and `group_by`.

- `filter_by` accepts an arbitrary `Callable[[Shape], bool]` type in additon  to `Axis` and `GeomType`.
- `group_by` accepts an arbitrary `Callable[[Shape], K]` (where `K` is the grouping key) type in addition to `Axis` and `SortBy`.
- additionally `group_by` returns a `GroupBy` object instead of a `list`. this object is still subscriptable by index, but also allows to retrieve the groups by key or by shape.
- internally the old argument types are converted to callable, this should simplify the code and facilitate future additions.

The PR is not merge-ready (I haven't updated the docstrings and there may be a few typing tweaks to do) but here it is for discussion.


snoyer:
Interestingly a [single test failed](https://github.com/gumyr/build123d/actions/runs/4418290801/jobs/7745209470#step:4:81) on only one of the 6 runners. @gumyr is that a thing known to happen, or do you think I broke the code in an non-obvious way?

gumyr:
I've seen a single test fail like that a few times.  I'm used to this in very large real-time systems which are difficult to make fully deterministic but to see it here is a little worrisome. I haven't really spent any time investigating but at first glance I didn't see a pattern.

snoyer:
> to see it here is a little worrisome. I haven't really spent any time investigating but at first glance I didn't see a pattern.

A few commits later the single seemingly unrelated test (`test_regular_polygon_matches_polar`) that failed the first time only on `3.10, windows` is now passing. Should we blame cosmic rays at the github datacenter and move on? (unless more intermittent failures pop up in the future)

tests/test_direct_api.py
```diff
@@ -2275,6 +2292,43 @@ def test_group_by(self):
         with self.assertRaises(ValueError):
             boxes.solids().group_by("AREA")
 
+    def test_group_by_callable_predicate(self):
+        boxesA = [Solid.make_box(1, 1, 1) for _ in range(3)]
```
gumyr, line 2347: `[Solid.make_box(1, 1, 1)] * 3` ?

gumyr (changes requested):
This is fancy stuff - thanks. It looks like others on the discord have similar ideas so although I hadn't thought to use the filters this way there is certainly value.  As you say, a few docstrings are missing but otherwise it looks good.

tests/test_direct_api.py
snoyer: That would put the same instance 3 times in the list which wasn't the intent. However in this one test it looks like it would work either way. Let me know.

gumyr:
Thank you

PR #178, opened 2023-03-14
Add Vector.distance_from_plane.

jrmobley:
I added this new function instead of adding implementation to `Vector.distance_to_plane`.
`distance_to_plane` says `raise NotImplementedError("Have not needed this yet, but OCCT supports it!")` which implied that the implementation would be passed down to OCCT, but I did not see an appropriate function to call upon.  Also, it was not clear to me if `distance_to_plane` was meant to be an unsigned distance.  For those reasons, I left it alone and added `distance_from_plane` that returns a signed distance.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -449,9 +449,9 @@ class Vector:
         """Minimum unsigned distance between vector and plane"""
         return plane.wrapped.Distance(self.to_pnt())
 
-    def signed_distance_from_plane(point: Vector, plane: Plane) -> float:
+    def signed_distance_from_plane(self, plane: Plane) -> float:
         """Signed distance from plane to point vector."""
-        return (point - plane.origin).dot(plane.z_dir)
+        return (self - plane.origin).dot(plane.z_dir)
     
     def project_to_plane(self, plane: Plane) -> Vector:
         """Vector is projected onto the plane provided as input.
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -449,6 +449,10 @@ class Vector:
         """Minimum distance between vector and plane"""
         raise NotImplementedError("Have not needed this yet, but OCCT supports it!")
 
+    def distance_from_plane(point: Vector, plane: Plane) -> float:
+        """Signed distance from plane to point vector."""
+        return (point - plane.origin).dot(plane.z_dir)
+    
     def project_to_plane(self, plane: Plane) -> Vector:
         """Vector is projected onto the plane provided as input.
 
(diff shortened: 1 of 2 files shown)


jrmobley:
Oh, now I see that the OCCT implementation in question is a method of `gp_Pln` not `gp_Vec` where I was looking.  I will test that it gives the same results and revise this PR.

jrmobley:
I implemented `Vector.distance_to_plane` in terms of `gp_Pln.Distance(gp_Pnt)` and revised the doc string to indicate that it returns an unsigned distance.  I renamed `Vector.distance_from_plane` to `Vector.signed_distance_from_plane` and left it in, because the signed distance is what I needed.

src/build123d/geometry.py
```diff
@@ -445,10 +445,14 @@ def project_to_line(self, line: Vector) -> Vector:
...
 
-    def distance_to_plane(self):
-        """Minimum distance between vector and plane"""
-        raise NotImplementedError("Have not needed this yet, but OCCT supports it!")
-
+    def distance_to_plane(self, plane: Plane) -> float:
+        """Minimum unsigned distance between vector and plane"""
+        return plane.wrapped.Distance(self.to_pnt())
+
+    def signed_distance_from_plane(point: Vector, plane: Plane) -> float:
```
```suggestion
    def signed_distance_from_plane(self, plane: Plane) -> float:
```

jrmobley: Wow, Python let me get away with that.  I've fixed the param name in the declaration and the body.

snoyer: As far as Python is concerned if `u` and `v` are `Vector` then `u.add(v)` is the same as `Vector.add(u, v)` so yeah you could write the method either way, but `self` with implicit typing is the prescribed style.

gumyr:
There is an existing test in `test_direct_api.py` is happily failing now:
```py
    def test_vector_not_implemented(self):
        v = Vector(1, 2, 3)
        with self.assertRaises(NotImplementedError):
            v.distance_to_plane()
```
would you remove this test as part of this PR?

jrmobley:
I see that it is failing a test that should have been removed.  I will fix the test.  I will also try to determine why the tests passed locally.

jrmobley:
Test were not passing locally.  I must not have run them after the later changes.  Thanks, CI, you da best!

jrmobley:
> There is an existing test in `test_direct_api.py` is happily failing now:
> ...
> would you remove this test as part of this PR?

Haha, sorry for the awkward flow.  I didn't see your comment until later.  I swear it wasn't there!  🤣

gumyr:
Thank you! You make 10 contributors - I'll have to check to see if we get a coupon now ;-)

PR #181, opened 2023-03-15
add optional x_dir argument to Plane(face)

jrmobley:
Because the default behavior of `Plane(face)` produces unpredictable `x_dir`s, I added an optional parameter to control it when needed.

Because I added even more complexity to the argument checking, I refactored it with the intention of making it more clear, robust, and DRY.  Included more tests to exercise the argument checks.  There remains some weakness regarding `VectorLike` arguments in that any `TypeError` raised in the `Vector` constructor is propagated without change, so the context of the error will be less clear to the user.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1478,7 +1478,7 @@ class Plane:
         """Return a plane from a OCCT gp_pln"""
 
     @overload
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #186, opened 2023-03-19
add `Shape.relocate`

snoyer:
Add a method to change a `Shape`s location while keeping it geometrically similar.

Changes include adding `Location.__eq__` and some minor unrelated formatting.

Not sure about the phrasing of the docstrings, let me know if you have anything better.


src/build123d/geometry.py
```diff
@@ -1145,6 +1145,13 @@ def __mul__(self, other: Location) -> Location:
     def __pow__(self, exponent: int) -> Location:
         return Location(self.wrapped.Powered(exponent))
 
+    def __eq__(self, other: Location):
+        return (
+            isinstance(other, Location)
+            and self.position == other.position
+            and self.orientation == other.orientation
```
gumyr, line 1070: This doesn't cover all the possibilities since the orientations are angles so mod 360 applies but more significantly different combinations result in the same end angle.  It seems safer to compare quaternion's 
```py
        quaternion1 = gp_Quaternion()
        quaternion.SetEulerAngles(
            gp_EulerSequence.gp_Intrinsic_XYZ,
            radians(x1),
            radians(y1),
            radians(z1),
        )
        quaternion2 = ...
        ... quaternion1.IsEqual(quaternion2)
```

```

gumyr (changes requested):
Hopefully users don't require the use of this method often but it's a bit of handy functionality to have just in case.

`relocate` looks good but the Location equivalency test isn't complete IIRC my Euler angle math. See the comment.

gumyr:
@snoyer I wasn't able to merge with the changes that have been made in the dev branch since this PR was created.  If you rebase I'll merge what's here and modify the orientation check to use gp_Quaternion. Thanks.

snoyer:
I clicked the "sync branch" button and blindly trusted Github to do its thing, CI seems happy so I assume nothing broke. Let me know if there's more for me to fix manually

gumyr:
Thanks - that worked.  I've updated the Location compare to use quaternions which should be safer so it all looks good.

Issue #56, opened 2022-11-22
SVG import - generate a set of BuildLine instances, one for each path.


gumyr:
To preserve more of the data, I think it is worth considering generating a list of list of edges, where each top-level list represents a separate path in the SVG file.

snoyer:
I'd be interested in contributing to the SVG import functionality.
First, a few observations:
1. the current dependency on `svgpathtools` would be enough to create wires form path strings (`M ... L ... Z ... M ....`)
2. however to parse whole svg documents reliably we would need to use something like [`svgelements`](https://github.com/meerk40t/svgelements) in order to also handle inherited transforms and css properties.

@gumyr, What's your take on the way forward?

gumyr:
There are quite a few aspects of how SVGs are handled that should be improved.  I'll list them here and we can discuss further:

1. When writing SVG files, all curves are converted to straight line segments. I've had laser cut parts made from the results of this algorithm but it seems very problematic in that any post processing that's looking for holes as circles (ellipses I guess) would not work.  The DXF exporter translates curves into appropriate curves for ezdxf - the same should be done for SVG.  One extra complication is that as far as I know all splines need to be converted to cubic or quadratic bezier curves - there are tools in OpenCascade that will help with the conversion but there would likely be effort required convert bezier with up to 26 knots (the OpenCascade limit IIRC) to something compatible with SVG.
2. All SVG exports are projections while DXF exports are just edges or wires.  It seems as though the projection functionality should be extracted such that it's independent and able to be used for both types of exports. The current export APIs would have to change but I think this could be made to work without too much churn from the current API.
3. SVG (and DXF IIRC) really only work with plain edges, wires & paths not faces other than extracting wires. Being able to export/import edges/faces with fill and stroke(color & patterns) would be a big step forward. build123d already has a `color` attribute for all objects, adding `stroke` for edges and wires might be useful.
4. As you mention, there very well could be fundamental parts of SVGs that are not being handled (other than what I've already described). I'm certainly not a SVG expert so your expertise is appreciated here.
5. Being able to do perspective projections seems desirable - this is a new feature for CadQuery and is already supported by OpenCascade so shouldn't be much work to implement.

If `svgelements` should replace `svgpathtools` then maybe it makes sense to start there but I don't know enough about either to really appreciate the difference.

Give all of the above - next steps?

snoyer:
[...]

PR #190, opened 2023-03-24
2d export

jrmobley:
I have refactored the SVG and DXF export into three classes.  The `Drawing` class creates an XY planar projection of a shape and provides the hidden and visible edges.  The `ExportDXF` and `ExportSVG` classes provide multi-layer export of XY planar shapes to DXF and SVG files respectively.

All b-splines are reduced to at most 3rd degree.  B-Splines are converted to poly-bézier curves for SVG export.  Conics *other than* circles and ellipses are exported as b-splines/béziers.  Bézier curves are exported to DXF as b-splines.  Otherwise edges are exported in their original form.

The interfaces of `ExportDXF` and `ExportSVG` are both modeled after features of DXF as exposed by the `ezdxf` library.  I have attempted to make the SVG output match DXF output as closely as possible.

I took the liberty of making some biggish decisions in this PR, so I wanted to get it posted for discussion and feedback.

### DIFF
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -1,0 +1,268 @@
+# pylint has trouble with the OCP imports
+# pylint: disable=no-name-in-module, import-error
+
+from build123d import *
+from build123d import Shape
+from OCP.GeomConvert import GeomConvert_BSplineCurveToBezierCurve #type: ignore
+from OCP.Geom import Geom_BSplineCurve #type: ignore
+from OCP.gp import gp_Dir, gp_Pnt #type: ignore
+from math import degrees
+from typing import Callable, List, Any
+from typing_extensions import Self
+from svgpathtools import CubicBezier, Path
+from xml.etree.ElementTree import Element, ElementTree, indent
+from enum import Enum, auto
+import ezdxf
+
+# ---------------------------------------------------------------------------
+# ---------------------------------------------------------------------------
+
+class Drawing(object):
+
+    def __init__(self):
+        pass
+    
+# ---------------------------------------------------------------------------
+# ---------------------------------------------------------------------------
+
+class AutoNameEnum(Enum):
+    def _generate_next_value_(name, start, count, last_values):
+        return name
+    
+class LineType(AutoNameEnum):
+    CONTINUOUS = auto()
+    CENTERX2 = auto()
+    CENTER2 = auto()
+    DASHED = auto()
+    DASHEDX2 = auto()
+    DASHED2 = auto()
+    PHANTOM = auto()
+    PHANTOMX2 = auto()
+    PHANTOM2 = auto()
+    DASHDOT = auto()
+    DASHDOTX2 = auto()
+    DASHDOT2 = auto()
+    DOT = auto()
+    DOTX2 = auto()
+    DOT2 = auto()
+    DIVIDE = auto()
+    DIVIDEX2 = auto()
+    DIVIDE2 = auto()
+
+# ---------------------------------------------------------------------------
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


gumyr:
This is excellent work, a huge thank you. Please remove "draft" so it can be merged and we can play with it.

The user would do something like:
```py
svg_file = ExportSVG(unit=Unit.MILLIMETER, line_weight=0.5)
svg_file.add_layer(name="template", line_weight=.., line_type=..)
svg_file.add_shape(template_edges, layer="template")
svg_file.add_layer(name="main", ...)
svg_file.add_shape(my_objects, layer="main")
svg_file.write(...)
```
?
Currently there is a simple `cad_objects.export_svg(...)` interface, it doesn't even seem worth keeping this.

@jdegenstein you've done a lot of SVG - what do you think?

jrmobley:
Ok, I'll mark this ready for review.

Here is one of my usage examples:
```py
        svg = ExportSVG(margin=5, line_weight=0.13)
        svg.add_shape(dwg.visible_lines)
        svg.add_layer("hidden", color=ColorIndex.LIGHT_GRAY, line_weight=0.09, line_type=LineType.ISO_DASH_SPACE)
        svg.add_shape(dwg.hidden_lines, layer="hidden")
        svg.write("dev-drawing.svg")
```

Shapes added without a `layer` argument are added to the default layer.  The default layer properties are supplied you construct the exporter.

Here is how I make a section for export.  The exporter expects all topology to be in the XY plane.
```py
with BuildPart() as x:
    Add(pcb)
    Section(m3_plane)
section: Compound = m3_plane.to_local_coords(x.part)
```

Regarding the `cad_objects.export_svg(...)` interfaces, you could ditch them or just re-implement them with the new exporters.  I'm ambivalent about that one.

jdegenstein:
Looks very interesting! I haven't tried it yet, but I would like to create e.g. the build123d logo without using inkscape. I want to be able to control line width/color and fill color. Is this possible with the new changes?

jrmobley:
> Looks very interesting! I haven't tried it yet, but I would like to create e.g. the build123d logo without using inkscape. I want to be able to control line width/color and fill color. Is this possible with the new changes?

I'm afraid it isn't really possible right now.  With the Drawing in particular, I don't think the underlying OCCT hidden-line removal algorithm maintains or reconstructs any information about faces.  So the output has no closed shapes to fill, it's just a line soup.  Maybe there is a completely different algorithm available that would be more suitable?  Research required.

For 2D sections, it is also currently processed as a line soup, but there does exist the possibility of traversing down the topology tree and organizing the output into fillable shapes.  But as the implementation stands currently, no the user cannot get filled output.

Regarding per-line colors and filled output, @gumyr I know you mention this in [issue 56](https://github.com/gumyr/build123d/issues/56#issuecomment-1464071083), and I was thinking about implementation.  The primary obstacle I see is that below the level of Compound in the topology tree, you have to rely upon OCCT to enumerate the child objects, and so you get OCCT objects without the b3d wrappers, so it seems like color assignments would be lost.  AFAICT.

gumyr:
@jrmobley by "you have to rely upon OCCT to enumerate the child objects" is this being done with OCCT methods or `edges()`. I ask because (in the algebra branch) I've made modifications such that `edges()` assigns a new `topo_parent` instance variable (this was done so the parent shape doesn't need to be provided in a fillet). Thus is would be easy to find a parent's `color` and use that.

jrmobley:
The exporters right now are using `edges()` so that `topo_parent` could enable some stuff.  I was picturing a tree traversal like you do in `show_topology()` in order to keep the structure intact when exporting SVG, but maybe that is overkill?

PR #197, opened 2023-03-31
re-fix PolarLine elif angle is not None

jdegenstein:
not sure if the prior change from a few days ago was intentionally reverted, but here it is again applied to objects_curve.py


gumyr:
Thanks - don't know how this got reverted.

PR #203, opened 2023-04-06
[docstring] fix sort_by defaults to Axis.Z not SortBy.Z

jdegenstein:


Jojain:
Real question why would there be a default for sort_by ? It's highly arbitrary to choose Z axis as a default no?

gumyr:
Thank you.  @Jojain is correct that sorting by anything is arbitrary but otherwise the order is undefined and by Axis.Z seems to be the most common use case.

PR #205, opened 2023-04-07
[docs] Intro Ex10 improved and minor doc fixes

jdegenstein:
Ex10 was not properly illustrating Select.LAST since it did not impact the final part creation. The new version better illustrates what Select.LAST does. Updated builder and algebra API versions of Ex10.


gumyr:
Thank you.

PR #206, opened 2023-04-07
Fixing links in documentation to work with the new algebra big rename

elliejs:
This is untested due to me not knowing how to set up mocked RST docs, so please test.


gumyr:
Thank you.  There were a lot of things to fix here as the code has undergone quite a bit of restructuring lately.  Much appreciated.

Issue #215, opened 2023-04-14
sweep over CenterArc does not behave as expected


OvisMaximus:
Given the following code
```py
from build123d import *

ring_diameter = 110

with BuildPart(Plane.XZ) as part_builder:
    with BuildLine(Plane.XY) as sweep_path:
        a_clamp_width = 12
        CenterArc((-ring_diameter/2, 0), ring_diameter / 2, -a_clamp_width / 2, a_clamp_width)
        #Line((0, -5), (0, 5))

    with BuildSketch(Plane.XZ) as sketch_builder:
        Rectangle(4.0, 8.0, align=(Align.CENTER,Align.MIN))

    sweep()

show_object(part_builder.part.wrapped, name="part")
show_object(sweep_path.line.wrapped, name="sweep_path")
show_object(sketch_builder.sketch.wrapped, name="sketch")
```
As a reference for my expectation, uncomment the Line statement and disable the CenterArc line. this is the rendered part:
![image](https://user-images.githubusercontent.com/5701377/231987837-1f448159-de54-4884-b048-c5b082eb3a99.png)

and that's the sketch it is based upon, also matching my expectation:
![image](https://user-images.githubusercontent.com/5701377/231965932-acdcdc3a-2ced-4f04-81ba-21dc4fcf7c2d.png)

The unmodified code above should create a similar solid with two deltas, first the angle of the arc is not identical to the extruded width with < 5% delta and secondly the solids Y edges should not be straight but slightly bend. Instead we get:
![image](https://user-images.githubusercontent.com/5701377/231985611-56352d43-ee11-49fe-928d-a65f3e12855f.png)

gumyr:
This issue is caused by a confusing feature, the ability to relocated the results of the operations.  As you have `Plane.XZ` as the workplane for BuildPart there was a relocation happening which clearly is unexpected.  This feature was too difficult to understand and has been removed.  The result is now:
![image](https://user-images.githubusercontent.com/85638518/232328886-4cb7d2f9-10bd-4aa5-8a14-f02df79a6409.png)

PR #217, opened 2023-04-19
align extruded solids according to the PartBuilder

hiaselhans:
this addresses the following example:

``` python
import build123d as bd
import jupyter_cadquery as jcq

with bd.BuildPart(bd.Plane.XZ) as bp:
    with bd.BuildSketch(bd.Plane.XZ):
        rect = bd.Rectangle(50,50)
        rect2 = bd.Rectangle(40,40)

    bd.extrude(amount=5)
    bd.extrude(rect2, amount=15)

bp
```

before:
image

after:
image

with the change, calls to extrude align the solids according to the workplanes in the context

### DIFF
--- a/src/build123d/operations_part.py
+++ b/src/build123d/operations_part.py
@@ -153,6 +153,16 @@ def extrude(
                     )
                 )
 
+
+    # Align the extrusions to the current workplanes
+    if to_extrude is not None:
+        if workplanelist := WorkplaneList._get_context():
+            if workplanes := workplanelist.workplanes:
+                new_solids = sum([
+                    [wp * solid for wp in workplanes]
+                    for solid in new_solids
+                ], start=[])
+
     if context is not None:
         context._add_to_context(*new_solids, clean=clean, mode=mode)
     elif clean:


gumyr:
In this case the repositioning makes sense but it quickly gets complex for the user to understand. Ironically, I recently removed the last of this repositioning logic from the operations.  See the comment in the docs here: https://build123d.readthedocs.io/en/latest/operations.html
> It’s important to note that objects created by operations are not affected by Locations, meaning their position is determined solely by the input objects used in the operation.

I guess I should expand this warning to workplanes as well.

Here are a couple ways to do the same:
```py
import build123d as bd

with bd.BuildPart(bd.Plane.XZ) as bp:
    with bd.BuildSketch(bd.Plane.XZ):
        rect = bd.Rectangle(50, 50)
    bd.extrude(amount=5)
    with bd.BuildSketch(bd.Plane.XZ):
        rect2 = bd.Rectangle(40, 40)
    bd.extrude(amount=15)

algebra = bd.Plane.XZ * bd.extrude(bd.Rectangle(50, 50), 5) + bd.Plane.XZ * bd.extrude(
    bd.Rectangle(40, 40), 15
)
```
One extra line in the builder solution or do the whole thing in one line with algebra.

hiaselhans:
Personally, i find it much more confusing to have an operation behave so differently when it is called with the explicit argument.
One time, the context matters, the other time it doesn't

How does it get complex with repositioning? and what were the reasons to remove the repositioning logic?

gumyr:
Here was the last Issue on this: https://github.com/gumyr/build123d/issues/215

The rule that operations are only a function of their parameters and not workplane or location contexts seems like the least complex for users to understand.  I think the problem behind this specific issue is that the `rect` and `rect2` are actually on `Plane.XY` which is not what you're expecting - isn't that true?  If so, this is a common misunderstanding of how BuildSketch works which I've tried to remedy with some documentation.  Not a great solution I admit.

jdegenstein:
I currently believe that using an object outside of the "scope" it is created in should not be promoted as a typical use pattern. By "using" I mean applying an operation from the object (such as extrude).

hiaselhans:
On the contrary, I am totally fine with 2D object being unaligned.
The confusing part is that it is hard to understand when context applies and when it doesn't.

What i would have expected:

- 2D objects / operations are defined in 2-dimensional space and thus only have a 2D but no 3D location inherent.
- 3D objects / operations are have a 3D location/alignment. Whenever a 3D object is created it is aligned to the current context

context would consist of:

- current 3D locations / orientations (Workplanes)
- current 2D locations (e.g. GridLocations)
- current Builder (3D/2D)


In this sense, i would expect the following to be valid:

``` python
with bd.BuildPart(bd.Plane.XZ) as part:
    # 3d-objects are being captured and oriented in the XZ plane in this context
    circle = bd.Circle(2)  # creates a 2D primitive -> not captured
    bd.extrude(circle, amount=10)  # extrudes the 2D primitive to a 3d Primitive, which is captured and aligned
```
Lifting the limitations in `Builder.validate_inputs` would be a way to make #218 work.

I would even go further and say that contexts can be nested and there is no "inertial system", so that in nested context the inner orientation is relative to the outer one.
In the following:

```
with bd.BuildPart(bd.Plane.XZ) as outer_part:
    with bd.BuildPart(bd.Plane.XY) as inner_part:
        bd.Cone(5, 1, 10)

jcq.show(outer_part, inner_part)
```

The cone in outer_part would point in the Y direction, while in inner_part it points towards the Z direction.
To my understanding that would remove the necessity of mode=Private in partbuilders and would pave the way for "library components".

gumyr:
Although these design decisions are valid, they aren't what I've decided to implement for build123d. Everyone has a different perspective of what makes sense to them; for example most users just expect to extrude along the Face normal even though that isn't really a very stable reference to decide on an extrusion direction (the normal can be flipped or even change across the Face if it's non-planar).

The current rule that operations only depend on their inputs and not the workplane/locations around them is my attempt to make a simple rule that applies everywhere and most users seem to find that it makes sense.  It seems as though you'd be more comfortable with the algebra api where you have full control of everything which is great, there is no "best" api.

gumyr:
Thanks for identifying this requirement. It is now implemented on dev by the builder-scope refactor (https://github.com/gumyr/build123d/commit/acaec742c7ed30e5dcc0d4e29d0652251a8bf4f3).

Workplane placement is now applied consistently to builder output across operations, rather than being handled as a special case for extrude. Closing this older PR as superseded.

PR #241, opened 2023-05-21
introduce short_sagitta parameter to RadiusArc to select between both possible cases

bernhard-42:
As discussed on Discord.


gumyr:
Thanks @bernhard-42 !

PR #254, opened 2023-06-10
Fix import_svg with folders 

### LINKED ISSUE #253: Using folders on import_svg breaks builder_name
**Steps to reproduce**

```python
svg_code, builder_name = import_svg_as_buildline_code("./assets/file.svg")
```

Results in the following generated code, which gives an error:
```python
from build123d import *
with BuildLine() as :
    Line((39760.0, 75303.0),(39889.0, 75206.0))
```

diesphink:
Fix #253


gumyr:
Thanks for fixing this!  I hope you find svg import useful.

PR #261, opened 2023-06-23
add annotations for ShapeList.__getitem__()

hiaselhans:
Without these annotations my IDE (vscode) is unable to resolve the type for xy.wires()[0]


gumyr:
Thank you!  This fixes a irritation I've had but it never got to the point where I tried to fix it.

PR #268, opened 2023-07-02
More ExportSVG features:

jrmobley:
Wires are output as a single path element.
Support for fill color.
Support of RGB colors.
Beginnings of unit tests.

See https://github.com/gumyr/build123d/discussions/260

### DIFF
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -3,6 +3,8 @@
 
 from build123d import *
 from build123d import Shape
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!  I know a lot of work has gone into this and it will be a great improvement to the existing SVG/DXG exporters.

gumyr:
Project coverage just went up 5.58% - woo hoo

Issue #272, opened 2023-07-05
Attempting to use zero as a value for Color red/green/blue results in one instead


voneiden:
Doing something like

```python
color = Color(red=0, green=0, blue=0)
```
will actually result in a `Color(1, 1, 1)` due to the way the code checks kwargs:

https://github.com/gumyr/build123d/blob/dev/src/build123d/geometry.py#L807

0 will cause the if to be skipped so the default of 1 is used instead.

Will submit a PR to address this shortly.

gumyr:
Closed with cdfeb7e3b65cb679a9c31920105ced25c127350d

Thanks

PR #273, opened 2023-07-05
Fix Color not handling kwargs with a value of zero

voneiden:
Ref #272


gumyr:
Thank you!  A difficult bug to find.

PR #274, opened 2023-07-05
Make build123d default planes match OCP CAD Viewer

jdegenstein:
Make the default planes in build123d match the behavior of OCP CAD Viewer and many other CAD software. In CQ-editor the "front" plane is the only one that is currently different (although may modify in my fork).


gumyr:
Thanks for fixing this!

jdegenstein:
You are very welcome, happy to help where I can

PR #310, opened 2023-08-31
add new mass constants to __init__.py

jdegenstein:
This should allow the mass constants to be automatically available from an import.


jdegenstein:
That worked, thank you! 
image

PR #313, opened 2023-09-06
Add dependencies for Apple Silicon

johnlamb:
Two extra adds to make installation work for Apple silicon, need to manually install py_lib3mf from GitHub repo and pip install python_utils.


gumyr:
Thank you! I think we'll all be happy when this custom install business gets resolved.

PR #316, opened 2023-09-10
Update key_concepts_algebra.rst fix a whitespace

42sol-eu:
only one whitespace to fix display in algebra key features


gumyr:
Thanks!

Issue #271, opened 2023-07-05
Upgrade OCP to 7.7.1.0


gumyr:
See https://github.com/CadQuery/OCP

jdegenstein:
I believe this issue can now be closed (one fewer issue for release 1.0.0)!

gumyr:
Complete

PR #325, opened 2023-09-20
Upgrade OCP to 7.7.1

jdegenstein:
As per issue #271 I tested on Win with conda/python=3.10 and it worked as expected

Only outstanding issue is that OCP-stubs are still on version = 7.7.0 (no release for 7.7.1 yet). I did not see OCP-stubs being used anywhere, but I could be wrong.


gumyr:
Thanks.  I don't think there was a change in the OCP API so maybe that's why there was no update of the stubs.

PR #331, opened 2023-09-25
Fix doc links to drafting objects and add to cheat sheet

jdegenstein:
links were to e.g. `objects_sketch.Arrow` instead of `drafting.Arrow`


gumyr:
Thanks.  I moved the code around a bit and didn't update everything.

PR #337, opened 2023-10-04
Fix property classmethods

Ruudjhuu:
Class methods wrapping other descriptors is deprecated in python 3.11. [docs classmethod](https://docs.python.org/3.11/library/functions.html#classmethod). I changed the few occurrences where this construction is used.

Mypy is also not able to check the call type in python < 3.11; https://github.com/python/mypy/issues/11619. Which results in always having a mypy error if one of these properties is used. 

I did also include some small changes to the ignore file:
* custom python enviroments were not ignored
* ignore vscode ide settings.
* The tests also generated a .3mf file which was not yet ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,12 @@ docs/_build/
 #mypy cache
 .mypy_cache
 build
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!  The problem with @classmethod/@property for the Plane and Axis classes is longstanding - I'm happy to have this fixed.

PR #340, opened 2023-10-09
Add angle parameter to chamfer

### LINKED ISSUE #339: Add ability to chamfer by angle
Sometimes asymmetric chamfers are specified by an angle and the length of one side, rather than the lengths of both sides.

Example:

image

Rather than the user having to calculate the length of the second side, chamfer should accept an angle parameter and calculate the correct chamfer length.

Desired capability:
```
result = extrude(Circle(125/2), -175)
result = chamfer(result.edges().sort_by(Axis.Z).first, length=175-90, angle=35/2)
```

mayhem-64:
Fixes #339

### DIFF
--- a/src/build123d/operations_generic.py
+++ b/src/build123d/operations_generic.py
@@ -28,6 +28,7 @@ license:
 """
 import copy
 import logging
+from math import radians, tan
 from typing import Union, Iterable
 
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you - this is a nice feature to have.

PR #345, opened 2023-10-12
add direction argument to chamfer.

Ruudjhuu:
When performing an asymmetrical chamfer on a 3d object, the implementation picks the first face it can find where the edge is part of and uses this as the direction of the first argument `length`. This issue results in inconstant chamfers in complex objects (usually on an object where conditionally the number of edges are altered).

I added a face argument to the chamfer functions. Now the users can control the direction of asymmetrical chamfers.

Let me know what you think.

### DIFF
--- a/src/build123d/operations_generic.py
+++ b/src/build123d/operations_generic.py
@@ -278,6 +278,8 @@ def chamfer(
         ValueError: no objects provided
         ValueError: objects must be Edges
         ValueError: objects must be Vertices
+        ValueError: Only one of length2 or angle should be provided
+        ValueError: Face can only be used in conjunction with length2 or angle
     """
     context: Builder = Builder._get_context("chamfer")
     if length2 and angle:
--- a/src/build123d/operations_generic.py
+++ b/src/build123d/operations_generic.py
@@ -258,6 +258,7 @@ def chamfer(
     length: float,
     length2: float = None,
     angle: float = None,
+    face: Face = None,
 ) -> Union[Sketch, Part]:
     """Generic Operation: chamfer
 
@@ -270,6 +271,8 @@ def chamfer(
         length (float): chamfer size
         length2 (float, optional): asymmetric chamfer size. Defaults to None.
         angle (float, optional): chamfer angle in degrees. Defaults to None.
+        face (Face): identifies the side where length is measured. The edge(s) must
+            be part of the face
 
     Raises:
         ValueError: no objects provided
@@ -283,6 +286,9 @@ def chamfer(
     if angle:
         length2 = length * tan(radians(angle))
 
+    if face and not (length2 or angle):
+        raise ValueError("Face can only be used in conjunction with length2 or angle")
+
     length2 = length if length2 is None else length2
 
     if (objects is None and context is None) or (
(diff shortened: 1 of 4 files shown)


jdegenstein:
Can you provide an example in which the behavior is unexpected? I am trying to understand if a simpler solution might be possible, e.g. a bool, `flip=True` (better name pending).

EDIT: Also, we are trying to keep test coverage above 95% before the 1.0 release, so tests that cover new functionality are greatly appreciated (if not mandatory at this point).

Ruudjhuu:
Of course,

This is a minimal example of the behavior I am talking about.

## Without changes
```python
from build123d import BuildPart, Box, fillet, Axis, chamfer, BasePartObject, Locations

def example(b_fillet: bool, b_chamfer: bool) -> BasePartObject:
    with BuildPart() as part:
        Box(20, 20, 20)

        if b_fillet:
            bottom_face = part.faces().sort_by(Axis.Z)[0]
            bottom_edge = bottom_face.edges().sort_by(Axis.Y)[0]
            fillet(objects=bottom_edge, radius=5)

        if b_chamfer:
            top_face = part.faces().sort_by(Axis.Z)[-1]
            top_edge = top_face.edges().sort_by(Axis.Y)[-1]
            chamfer(objects=top_edge, length=3, length2=10)

    return BasePartObject(part.part)


with BuildPart() as obj1:
    example(True, True)

with BuildPart() as obj2:
    with Locations((30, 0)):
        example(False, True)
```

Which results in:
image

## With Changes
```python
from build123d import BuildPart, Box, fillet, Axis, chamfer, BasePartObject, Locations

def example(b_fillet: bool, b_chamfer: bool) -> BasePartObject:
    with BuildPart() as part:
        Box(20, 20, 20)

        if b_fillet:
            bottom_face = part.faces().sort_by(Axis.Z)[0]
            bottom_edge = bottom_face.edges().sort_by(Axis.Y)[0]
            fillet(objects=bottom_edge, radius=5)

        if b_chamfer:
            top_face = part.faces().sort_by(Axis.Z)[-1]
            top_edge = top_face.edges().sort_by(Axis.Y)[-1]
            chamfer(objects=top_edge, length=3, length2=10, face=top_face)

    return BasePartObject(part.part)


with BuildPart() as obj1:
    example(True, True)

with BuildPart() as obj2:
    with Locations((30, 0)):
        example(False, True)
```

image


## Test
Regarding the tests, I'll add bad weather scenarios if you think there is need for this solution. When I do not know how to reproduce an exception, is it fine to ignore it with `# pragma: no cover` or just ignore it without skipping the line for measurements?

gumyr:
Thanks for such a thorough exploration of this problem.  

I like the idea of being able to control an asymmetric chamfer (I raised a CadQuery issue on this quite a long time ago). As @jdegenstein points out a flip (or something similar) parameter could be used; however, that approach isn't all that deterministic as directly specifying one of the Shapes as done here. One could use a selector to ensure the chamfer is correct independent of the internal workings of the OCP chamfer algorithms.

In `topology.py` there are several different chamfer methods:
- `Mixin3D.chamfer()`
- `Face.chamfer_2d()`, and
- `Wire.chamfer_2d()`.
in addition to the `chamfer` function/operation.

How about extending the improvement in this PR to 2d as well? In the `chamfer` function, the `face` parameter could be renamed to `reference` (or something like that) which could either be an `Edge` or `Face`.  The `Mixin3D` method would have a reference `Face` as you've done here, while the `Face` and `Wire` methods would have a reference `Edge`.

How does that sound?

gumyr:
Thank you!  This is a very nice enhancement to the chamfer functionality.

Issue #300, opened 2023-08-15
svg import from solvespace not working as expected


bernhard-42:
I was playing around with SolveSpace and created this sketch:
<img width="540" alt="image" src="https://github.com/gumyr/build123d/assets/954408/e69de8d9-519b-4907-88f2-95ac74e83d29">

Affinity Designer imports it as 
<img width="566" alt="image" src="https://github.com/gumyr/build123d/assets/954408/4fe0c8bf-7c9c-455c-b05e-ee37993b361e">

However, importing into build123d with

```python
p = import_svg("profile.svg")
show(p)
```

results in 

<img width="407" alt="image" src="https://github.com/gumyr/build123d/assets/954408/e53f5fac-5c4d-4d3c-a410-f8301e4a5220">

Two issues:
1) the whole sketch is flipped vertically
2) one of the arcs goes in the wrong direction

The SolveSpace file and the svg can be found on Discord (https://discord.com/channels/964330484911972403/1074840524181217452/1141086336124719165)

gumyr:
With commit dd4f923f6cd62d25c45dc4188bf223778047e619 both problems are fixed - the arc is correct and there is a parameter to flip y which defaults to True.
![image](https://github.com/gumyr/build123d/assets/85638518/2de4c3ab-053e-4f20-ae85-09f7c9e5cf60)

PR #346, opened 2023-10-15
Fix SVG import Y-axis and arc degrees.

petejohanson:
I believe this fixes issue #300 as noted there with SVG imports being "flipped" and having arcs that are incorrect. I had similar issues with trying to import an SVG exported from KiCad, where using current version results in:

Screenshot from 2023-10-15 11-24-11

And with these fixes:

Screenshot from 2023-10-15 11-23-16

I'm happy to contribute a test for this, but the SVG import tests don't offer much in terms of the direction you'd want to go with such a test. I'm open to ideas.

Thanks!

### DIFF
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -183,12 +183,12 @@ def import_svg_as_buildline_code(file_name: str) -> tuple[str, str]:
             class_name = type(curve).__name__
             if class_name == "Arc":
                 values = [
-                    (curve.__dict__["center"].real, curve.__dict__["center"].imag)
+                    (curve.__dict__["center"].real, -curve.__dict__["center"].imag)
                 ]
                 values.append(curve.__dict__["radius"].real)
                 values.append(curve.__dict__["radius"].imag)
-                values.append(curve.__dict__["theta"])
-                values.append(curve.__dict__["theta"] + curve.__dict__["delta"])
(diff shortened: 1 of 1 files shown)


gumyr:
Thanks for looking into this. There are two separate issues here - the inversion of the arcs and the flipping of the y axis. 

1. I've looked into the arc problem and it appears that the problem is that the start and end angles are reversed.  Commit [97a4b66671529301c4c64cb6ade65969793d5791 ](https://github.com/gumyr/build123d/commit/97a4b66671529301c4c64cb6ade65969793d5791) attempts to fix this by sorting these values which does resolve the problem with the svg file referenced in the issue 300.  Please try your import again to see if this was the same issue you were seeing.
2. Flipping the y-axis (svg files have an inverted y-axis) should be an optional parameter.  This will match the new svg importing that is coming which is much more capable.

gumyr:
The svg import functionality has been replaced with this commit: dd4f923f6cd62d25c45dc4188bf223778047e619. This change should both fix the arcs and provides a flip_y option to correct for svg's different orientation.  These changes make this PR invalid (as the core code isn't in the build123d repo anymore) so this PR is being closed.

petejohanson:
@gumyr Thanks! I'll give the updated version a try when I have a few minutes.

PR #354, opened 2023-10-24
Update installation.rst to fix Apple Silicon instructions

jdegenstein:
ocpsvg was in the wrong spot (needs the --no-deps flag)


gumyr:
Thanks.  It sucks that I can't verify the Apple silicon builds - I get them wrong almost every time :(

PR #361, opened 2023-10-29
fix axis equality

### LINKED ISSUE #358: Axis equality broken
`Axis.X == Axis.X` evaluates to `False`.

This invalidates the tests `TestShapeList.test_filter_by` which would fail if the equality worked.

Ruudjhuu:
fixes #358

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -460,24 +460,6 @@ class Axis(metaclass=Axis_meta):
 
     _dim = 1
 
-    @classmethod
-    @property
-    def X(cls) -> Axis:
-        """X Axis"""
-        return Axis((0, 0, 0), (1, 0, 0))
-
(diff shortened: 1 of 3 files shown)


src/build123d/geometry.py
```diff
@@ -460,24 +460,6 @@ class Axis(metaclass=Axis_meta):
...
-    def Y(cls) -> Axis:
-        """Y Axis"""
-        return Axis((0, 0, 0), (0, 1, 0))
-
-    @classmethod
-    @property
-    def Z(cls) -> Axis:
-        """Z Axis"""
-        return Axis((0, 0, 0), (0, 0, 1))
-
```
Ruudjhuu, line 480: This is cleanup as they are located in the metaclass few lines above.

gumyr:
Thank you!

PR #367, opened 2023-11-05
Avoid type warnings when adding tuples to vectors

yeicor:
This simple typing fix avoids this warning from the tea cup example:

Screenshot_20231105_121154


gumyr:
Thank you!

PR #374, opened 2023-11-09
pack(): introduced a new helper for packing parts into a 2D footprint.

fischman:
Also added documentation for the new facility as well as a nascent CONTRIBUTING.md to collect pointers to the tools used by developers of this project.


gumyr:
Very nice, thank you!

As an extension, have you considered making `pack` a little more general? As is it's defined as:
```py
def pack(objects: Collection[Part], padding: float) -> Collection[Part]:
```
which implies only `Parts` are valid but it seems as though there is nothing here that wouldn't work for all `Shapes` - i.e. the user might have a collection of `Faces` they would like to pack for display as in `show(pack(my_face_list, 10))`

fischman:
Thanks for the speedy review+merge!

> As an extension, have you considered making `pack` a little more general? As is it's defined as:

Good idea: done in https://github.com/gumyr/build123d/pull/377.

PR #387, opened 2023-11-15
De-lint topology.py, joints.py, and operations_part.py

fischman:
Reduces pylint findings count from 94 to 52.

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -36,7 +36,6 @@ from __future__ import annotations
 #   too-many-statements, too-many-instance-attributes, too-many-branches
 import copy
 import itertools
-import logging
 import os
 import platform
 import sys
@@ -564,6 +563,7 @@ class Mixin1D:
         Returns:
             Union[None, Plane]: Either the common plane or None
         """
+        # pylint: disable=too-many-locals
         # BRepLib_FindSurface could help here
         points: list[Vector] = []
         all_lines: list[Edge, Wire] = [
@@ -816,6 +816,7 @@ class Mixin1D:
         Returns:
             Wire: offset wire
         """
+        # pylint: disable=too-many-branches, too-many-locals, too-many-statements
         kind_dict = {
             Kind.ARC: GeomAbs_JoinType.GeomAbs_Arc,
             Kind.INTERSECTION: GeomAbs_JoinType.GeomAbs_Intersection,
@@ -1391,6 +1392,7 @@ class Shape(NodeMixin):
         topo_parent (Shape): assembly parent of this object
 
     """
+    # pylint: disable=too-many-instance-attributes, too-many-public-methods
 
     _dim = None
 
@@ -1763,7 +1765,7 @@ class Shape(NodeMixin):
         try:
             upgrader.Build()
             self.wrapped = downcast(upgrader.Shape())
-        except:
+        except: # pylint: disable=bare-except
             warnings.warn(f"Unable to clean {self}")
         return self
 
@@ -2573,6 +2575,7 @@ class Shape(NodeMixin):
         Returns:
             Shape: Resulting object may be of a different class than self
         """
+        # pylint: disable=too-many-branches
 
         # def ocp_section(this: Shape, that: Shape) -> (list[Vertex], list[Edge]):
         #     # Create a BRepAlgoAPI_Section object
@@ -3039,6 +3042,7 @@ class Shape(NodeMixin):
             The projected faces
 
         """
+        # pylint: disable=too-many-locals
         path_length = path.length
         # The derived classes of Shape implement center
         shape_center = self.center()  # pylint: disable=no-member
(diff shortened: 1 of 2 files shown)


src/build123d/joints.py
```diff
@@ -87,28 +87,29 @@ def __init__(
         super().__init__(label, to_part)
 
     @overload
-    def connect_to(self, other: BallJoint, *, angles: RotationLike = None, **kwargs):
+    def connect_to(self, other: BallJoint, *args, angles: RotationLike = None, **kwargs):
```
MatthiasJ1: `*` is not equivalent to `*args`

fischman: Thanks @MatthiasJ1. Can you say more about what problem you see here?
(I'm newly returning to python work so wouldn't be shocked if I'm missing something silly here)

snoyer: `*,` forces subsequent argument to be explicitly passed as keyword arguments. In this case it would make `a.connect_to(b, foo)` invalid and would require `a.connect_to(b, angles=foo)`. see https://peps.python.org/pep-3102/
That _may_ still be the right choice here but that's a matter of API design not simply cleanup refactoring.

fischman: Thanks @snoyer. IIUC `a.connect_to(b, foo)` is invalid both in the "before" and "after" versions of the code being pointed at here, and `a.connect_to(b, angles=foo)` is valid in both. If that's true, then the change seems like a no-op. 
Just to verify my understanding I wrote https://colab.research.google.com/drive/1fAbilkOI2QNnYZ6-2JaQipCkxH9fJt_R which seems to show identical behavior for simplified versions of the before/after code. 
Do you have an example of a call that is valid before and invalid after, or invalid before (and where catching the invalidity is useful, I suppose) but silently "valid" after?

snoyer: good point, my example was poorly chosen. I guess the `*args,` version would accept unnamed (unused?) positional arguments that the `*, ` version wouldn't (eg. `g(1,2,b=3)` is ok but `f(1,2,b=3)` isn't).
Again, whether that's good/better or bad/worse is up to API considerations. Just splitting hairs about it not being a 100% equivalent refactor here :)

fischman: Thanks again @snoyer. I definitely appreciate the double- and triple-checking here.
If the API change is judged ok by @gumyr I'm happy to reword the commit/PR descriptions to reflect this.
Will flip the PR to "draft" mode pending that judgement.

gumyr: I'd like the `*,` to remain thus forcing only kwargs.  Joints parameters are already kinda complex and I expect them to get more so over time so I want the clarity that comes with kwargs.

src/build123d/topology.py
```diff
@@ -2941,7 +2944,7 @@ def to_arcs(self, tolerance: float = 1e-3) -> Face:
     def _repr_javascript_(self):
         """Jupyter 3D representation support"""
 
-        from .jupyter_tools import display
+        from .jupyter_tools import display # pylint: disable=import-outside-toplevel
```
gumyr: I'd like this one to remain.  I expected this import would work as a normal import at the top of the module but a quick test failed and I didn't have time to investigate why.

src/build123d/topology.py
```diff
@@ -7303,7 +7317,7 @@ def _connect_to(self, other: Joint, **kwargs):  # pragma: no cover
         self.connected_to = other
 
     @abstractmethod
-    def connect_to(self, other: Joint, **kwargs):
+    def connect_to(self, other: Joint, *args, **kwargs):
```
gumyr: Same comment as above.  Should this be `def connect_to(self, other: Joint, *, **kwargs):`? Seems redundant.

gumyr (changes requested):

src/build123d/joints.py
fischman: AFAICT `*, **kwargs` in succession isn't valid Python, and an intervening (named) keyword parameter is required, so the best I've been able to come up with so far is https://github.com/gumyr/build123d/commit/85ee55fabd5b981894bb0d0cd1ca90f8c0dabeb1
@snoyer @MatthiasJ1: do you see any issues with that approach?
@gumyr Please LMK if that looks good to you and I'll rework this PR on top of that commit.

gumyr: I don't have time right now to properly investigate but I don't like:
```py
def connect_to(self, other: Joint, *, ignored: int = 0, **kwargs):
``` 
even though this might be the best solution at this point.

At this point, how about separating the Joint changes for everything else so we can close the others off?

gumyr: I just tried removing the arguments from the abstract methods and this seems to work:
```py
    @abstractmethod
    def connect_to(self, other: Joint):
        """All derived classes must provide a connect_to method"""
        raise NotImplementedError

    @abstractmethod
    def relative_to(self, other: Joint) -> Location:
        """Return relative location to another joint"""
        raise NotImplementedError
```
Is that all that needs to be done?

fischman: :facepalm: yes, done.

src/build123d/topology.py
fischman: Resolved by dropping existing variadic per other comment thread instead of adding bare star.

src/build123d/topology.py
fischman: Done.

gumyr:
Thank you.

PR #390, opened 2023-11-19
add posibility to compare vertices

Ruudjhuu:
Add possibility to compare Vertices.

After these changes I have a problem with 1 testcase: test_basic_functions 

```python
    def test_basic_functions(self):
        """Test creating a line and returning properties and methods"""
        with BuildLine() as test:
            l1 = Line((0, 0), (1, 1))
            TangentArc((1, 1), (2, 0), tangent=l1 % 1)
            self.assertEqual(len(test.vertices()), 3)
            self.assertEqual(len(test.edges()), 2)
            self.assertEqual(len(test.vertices(Select.LAST)), 2)
            self.assertEqual(len(test.edges(Select.LAST)), 1)
            self.assertEqual(len(test.edges(Select.ALL)), 2)
```

`test.vertices(Select.LAST)` is expected to return 2 new later created points, but technically point (1.1) is already created by the `Line` while  `TangentArc` only adds point (2.0).

My proposal is to adapt this test case so `test.vertices(Select.LAST)` is expected to return 1 point. The problem is: I am not familiar enough with the code base to fully understand the impact of this behavior change. Is this proposal OK for you?

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -6632,6 +6632,15 @@ class Vertex(Shape):
             raise StopIteration
         return value
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Vertex):
+            return NotImplemented
+
+        return self.to_tuple() == other.to_tuple()
+
+    def __hash__(self) -> int:
+        return hash(self.to_tuple())
+
 
 class Wire(Mixin1D, Shape):
     """A series of connected, ordered edges, that typically bounds a Face"""
--- a/tests/test_build_line.py
+++ b/tests/test_build_line.py
@@ -49,7 +49,7 @@ class BuildLineTests(unittest.TestCase):
             TangentArc((1, 1), (2, 0), tangent=l1 % 1)
             self.assertEqual(len(test.vertices()), 3)
             self.assertEqual(len(test.edges()), 2)
-            self.assertEqual(len(test.vertices(Select.LAST)), 2)
+            self.assertEqual(len(test.vertices(Select.LAST)), 1)
             self.assertEqual(len(test.edges(Select.LAST)), 1)
             self.assertEqual(len(test.edges(Select.ALL)), 2)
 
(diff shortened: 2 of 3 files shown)


gumyr:
This change would break the ability to determine if two Edges share the same Vertex object which is used to navigate the topology of a larger object.  For example, if you had a Wire with many Edges and you want to find which Edges are connected you can currently find the Edges that share the same Vertex object (not just Vertices at the same point in space). 

One can determine if two Vertices are in the same location by using `Vector(vertex1))==Vector(vertex2))`.

Ruudjhuu:
I do not know the exact implementation of the behavior you are talking about, but to my knowledge if you want to compare two instances and check if it is the exact same instance, the python `is` keyword can be used. if you want to compare by value (different instances but same content) you should use python `==` syntax. 

example after my changes:

```python
>>> from build123d import Vector
>>> vec_a = Vector(1,2,3)
>>> vec_b = Vector(1,2,3)
>>> vec_a is vec_b
False
>>> vec_a == vec_b
True
>>> vec_b = vec_a
>>> vec_a is vec_b
True
```

Do you think it might be valuable to look into this? Currently I am satisfied with the workaround you provided. 

I'm also a little bit worried that not more testcases fail if this change breaks that behavior.

gumyr:
You are correct in that `is` should really be used for this purpose, unfortunately the `is` operator can't be reassigned and only compares the Python wrapper objects not the underlying OpenCascade topological objects.  Consider this:
```py
box = Box(1, 1, 1).solid()
top_corner_vertex = (
    box.vertices().group_by(Axis.Z)[-1].group_by(Axis.X)[-1].sort_by(Axis.Y)[0]
)
all_edges = box.edges()

corner_edges1 = all_edges.filter_by(lambda e: top_corner_vertex in e.vertices())
print(len(corner_edges1))

corner_edges2 = all_edges.filter_by(
    lambda e: any(top_corner_vertex is v for v in e.vertices())
)
print(len(corner_edges2))

corner_edges3 = all_edges.filter_by(
    lambda e: any(top_corner_vertex.wrapped.IsSame(v.wrapped) for v in e.vertices())
)
print(len(corner_edges3))
```
```
3
0
3
```
The `==` operator is mapped to `IsSame` as shown above. If you have an ideas on how this could be improved without loosing the ability to compare the topological objects please let me know.

image

Ruudjhuu:
I do understand the issue, I see FreeCad also struggling with this and it's python api. I think the current solution is acceptable. Going to close this issue.

PR #395, opened 2023-11-26
update polygon doc

MatthiasJ1:


gumyr:
There is no restriction to the order of the points, here is a clockwise example:
```py
p = Polygon((0, 0), (0, 1), (1, 1), (1, 0), (0, 0), align=None)
```
image
Is the concern that the object is up-side-down?

MatthiasJ1:
Yes. I clarified the docs.

gumyr:
I don't understand why you're requesting this change. Both CW and CCW orders generate polygons and the code does not attempt to enforce any order. Is there something specific that is causing a problem here?

MatthiasJ1:
Yes, that is why I specified it is the winding order which only determines the normal, not the shape.
The code does enforce an order when it comes to the polygon's normal which is what I am documenting.

gumyr:
How about this instead?
```py
    Add polygon(s) defined by given sequence of points to sketch. 

    Note that the order of the points define the normal of the Face that is created in Algebra mode, 
    where counter clockwise order creates Faces with their normal being up while a clockwise order 
    will have a normal that is down.  In Builder mode, all Faces are up.
```
with no change to the description of `pts`.

MatthiasJ1:
Sounds good to me. I had to figure out the winding order by testing when lofting in algebra mode so this would address that.

gumyr:
Fixed with commit f622815708a3a15b171e4532ba8715646cc619e1

Issue #270, opened 2023-07-04
Loft should allow Vertices as the first and last elements


gumyr:

jdegenstein:
Did some prototyping work to add this functionality to `make_loft` in hindsight maybe more of the input validation needs to go into `loft` instead, but a lot of the logic should be reusable:
```py
from typing import Union
from OCP.BRepOffsetAPI import (
    BRepOffsetAPI_MakeFilling,
    BRepOffsetAPI_MakeOffset,
    BRepOffsetAPI_MakePipe,
    BRepOffsetAPI_MakePipeShell,
    BRepOffsetAPI_MakeThickSolid,
    BRepOffsetAPI_ThruSections,
)

# def make_loft2(cls, objs: Union[list[Wire], list[Vertex, Wire]], ruled: bool = False) -> Solid:
def make_loft2(objs: Union[list[Wire], list[Vertex, Wire]], ruled: bool = False) -> Solid:
    """make loft

    Makes a loft from a list of wires.

    Args:
        wires (list[Wire]): section perimeters
        ruled (bool, optional): stepped or smooth. Defaults to False (smooth).

    Raises:
        ValueError: Too few wires

    Returns:
        Solid: Lofted object
    """

    if len(objs) < 2:
        raise ValueError("More than one wire, or a wire and a vertex is required")

    if all(isinstance(obj, Wire) for obj in objs):
        pass
    elif any(isinstance(obj, Vertex) for obj in objs) and any(isinstance(obj, Wire) for obj in objs):
        if len(objs) == 2:
            pass
        elif isinstance(objs[0],Vertex) and isinstance(objs[-1],Vertex):
            pass
        else:
            raise ValueError("Vertices must be the first, last, or first and last elements")
    elif all(isinstance(obj, Vertex) for obj in objs):
        raise ValueError("At least one wire is required if vertices are the first, last, or first and last elements")
    
    # the True flag requests building a solid instead of a shell.
    loft_builder = BRepOffsetAPI_ThruSections(True, ruled)

    for obj in objs:
        if isinstance(obj,Vertex):
            loft_builder.AddVertex(obj.wrapped)
        elif isinstance(obj,Wire):
            loft_builder.AddWire(obj.wrapped)

    loft_builder.Build()

    # return cls(loft_builder.Shape())
    return loft_builder.Shape()

Solid.make_loft2 = make_loft2

a = RegularPolygon(1,3).wires()[0]
v1 = Vertex(0,0,-1)
[...]

PR #397, opened 2023-11-27
Add vertices support to loft

jdegenstein:
should address issue https://github.com/gumyr/build123d/issues/270 to add vertex support to loft. This is a good way of creating polygonal shapes such as a rectangular pyramid (using ruled=True) that is not very easy otherwise in build123d.

I also added `Sketch` as an acceptable input type, which I believe was a missing feature for `loft`

The formatter also caught some other whitespace issues elsewhere, hopefully that is OK.


gumyr:
Thank you, this is a nice addition to loft.

shayded-exe:
This is beautiful

PR #400, opened 2023-11-28
PyPI Publish GH Action

jdegenstein:
This is ready for testing, it is setup to publish to the PyPI testing target (not any real PyPI target). There may need to be some changes made to `pyproject.toml` but I have not done that yet. Further info about `pyproject.toml` is here:

https://cibuildwheel.readthedocs.io/en/stable/options/

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,80 @@
+name: Wheel building and publishing
+
+on:
+  # schedule:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
+  test_wheel_building:
+    #
+    permissions:
+      contents: none
+    uses: someactionyml.yml #TODO: replace with wheel build action
+    if: (github.event_name == 'push') # || github.event_name == 'pull_request')
+    with:
+      upload_to_pypi: false
+      targets: |
+        - cp311-manylinux_x86_64
+
+    secrets:
+      pypi_token: ${{ secrets.pypi_token }} # consider using alternate strategy
+
+  build_wheel:
+    # This does the actual wheel building or if triggered manually via the workflow dispatch, or for a tag.
+    # this job does NOT publish the wheels
+    permissions:
+      contents: none
+    uses: someactionyml.yml #TODO: replace with wheel build action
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,84 @@
+name: Wheel building and publishing
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main  #check branch
```
gumyr: There is no main branch yet, everything goes into the `dev` branch currently.  When we get to a release the `main` branch will be created and it should be used here.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, windows-2022, macos-11]
+    if: (github.repository == 'gumyr/build123d' && ( startsWith(github.ref, 'refs/tags/v') || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || contains(github.event.pull_request.labels.*.name, 'Build all wheels')))
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.16.2
```
gumyr, line 42: Does `cibuildwheel` need to be installed?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,84 @@
...
+  #     - uses: actions/upload-artifact@v3
+  #       with:
+  #         path: dist/*.tar.gz
+      
+  upload_pypi:
+    needs: [build_wheels] #, build_sdist]
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/build123d
```
gumyr, line 67: Is `https://pypi.org/p/build123d` short for `https://pypi.org/project/build123d`? Do either work?

gumyr (commented):
If you change the branch I think the PR should be merged and we can start getting it to work.  Unfortunately, I don't know much about this so I can't provide very useful feedback.

.github/workflows/publish.yml
jdegenstein: I could be wrong, but not to my knowledge. `uses: pypa/cibuildwheel@v2.16.2` should install `cibuildwheel` within the GH action runner.

.github/workflows/publish.yml
jdegenstein: Yes, that is a shortcut and either will work (from PyPI docs)

PR #409, opened 2023-12-01
Fix Location.* typing

MatthiasJ1:
Fixes typing for `Location.__mul__`. Below is now typed correctly and fixes autocomplete.

```
Pos(10,0,0) * Sphere(1)
```


gumyr:
Thank you!

PR #411, opened 2023-12-01
pyproject.toml: remove OCP-stubs as dependency

jdegenstein:
preparing for PyPI release that does not allow depending on GH repos


jdegenstein:
Probably best to wait on this until the other GH deps are on PyPI, but updated to remove OCP-stubs as a dep altogether.

PR #444, opened 2023-12-22
Fix Plane move and add moved

MatthiasJ1:
`Plane.move` contradicts the documentation on `move` vs `moved` ([here](https://build123d.readthedocs.io/en/latest/key_concepts.html#location)) and is missing `moved`, therefore operations like `Pos() * Plane` will fail.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -2045,6 +2045,23 @@ class Plane(metaclass=PlaneMeta):
             loc (Location): relative change
 
         Returns:
+            Plane: relocated plane (self)
+        """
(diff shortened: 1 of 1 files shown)


gumyr:
Thank you!  The test for Plane.move isn't complete:
```py
    def test_move(self):
        pln = Plane.XY.move(Location((1, 2, 3)))
        self.assertVectorAlmostEquals(pln.origin, (1, 2, 3), 5)
```
Would you correct this test by checking that the instance of the Plane hasn't changed and introduce a new test for `moved` that checks that the instances are different?

Issue #432, opened 2023-12-11
Bug with the 3MF exporter


KarimJedda:
Hello! Thanks for the amazing library, it's a lot of fun :) 

Here's an issue I'm struggling with. The following, doesn't work: 

```python
from build123d import *

with BuildPart() as funky_box:
    Box(3, 3, 3)
    with BuildSketch(*funky_box.faces()):
        Rectangle(1, 2, rotation=45)
    extrude(amount=10)


outset = funky_box.part
outset.color = Color("blue")  

exporter = Mesher(unit=Unit.CM)
exporter.add_shape(outset)
exporter.write("single.3mf")
```
with the following error message: 
```
Traceback (most recent call last):
  File "/home/potato/tests/build123d/third.py", line 14, in <module>
    exporter.add_shape(outset)
  File "/home/potato/tests/build123d/venv/lib/python3.10/site-packages/build123d/mesher.py", line 422, in add_shape
    raise RuntimeError("3mf mesh is invalid")
RuntimeError: 3mf mesh is invalid
```

whereas the following, does work:

```python
from build123d import *

with BuildPart() as funky_box:
    Box(3, 3, 3)
    with BuildSketch(*funky_box.faces()):
        Rectangle(2, 2, rotation=45) # changed here
    extrude(amount=10)


outset = funky_box.part
outset.color = Color("blue")  

exporter = Mesher(unit=Unit.CM)
exporter.add_shape(outset)
exporter.write("single.3mf")
```

An `STL` export works in both cases without errors, so I suspect this to be an issue with `py_lib3mf`. Any help would be appreciated, meanwhile I'll stick to `STL`.

jdegenstein:
This may be related to https://github.com/gumyr/build123d/issues/350 and https://github.com/gumyr/build123d/issues/311

gumyr:
Thanks for raising the issue.  I was able to isolate the problem further based on your test:
```py
with BuildPart() as funky_box:
    Box(3, 3, 3)
    with BuildSketch(funky_box.faces().sort_by(Axis.Z)[-1]):
        Rectangle(1, 2)
    extrude(amount=1)
```
![image](https://github.com/gumyr/build123d/assets/85638518/02a5d6a1-de3b-49e5-9ab7-c69b9eee6cef)

works, but
```py
with BuildPart() as funky_box:
    Box(3, 3, 3)
    with BuildSketch(funky_box.faces().sort_by(Axis.Z)[-1]):
[...]

Issue #311, opened 2023-09-01
Certain Fonts/Characters Break the Lib3MF Mesher()


jdegenstein:
When using the Lib3MF Mesher() it was noticed that the Lib3MF library sometimes returns that the mesh is invalid. After further investigation I noticed that certain combinations of fonts/characters seem to be the "root cause" (as was the case in the originally found example).

this works fine:
```py
with BuildPart() as p:
    Box(25,10,10)
    with Locations((0,20)):
        Box(10,10,10)
    with BuildSketch() as s:
        Text('B',8,font="Arial")
    extrude(amount=15,mode=Mode.SUBTRACT)

exporter = Mesher()
exporter.add_shape(p.part)
exporter.write('3MF_test_2boxes_BCtext.3mf')
```
and this works too:
```py
with BuildPart() as p:
    Box(25,10,10)
    with Locations((0,20)):
        Box(10,10,10)
    with BuildSketch() as s:
        Text('B',8,font="Arial") #this works with below
        with Locations((5,0)): 
           Text('C',8,font="Arial") #this works with above
    extrude(amount=15,mode=Mode.SUBTRACT)
#omitted same exporter statements as above
```
This does NOT work:
```py
with BuildPart() as p:
    Box(25,10,10)
    with Locations((0,20)):
        Box(10,10,10)
    with BuildSketch() as s:
        Text('BC',8,font="Arial") #this does NOT work alone
    extrude(amount=15,mode=Mode.SUBTRACT)
#omitted same exporter statements as above
```
Neither does this:
```py
with BuildPart() as p:
    Box(25,10,10)
    with Locations((0,20)):
        Box(10,10,10)
    with BuildSketch() as s:
        Text('B C',8,font="Arial") #this does NOT work alone
    extrude(amount=15,mode=Mode.SUBTRACT)
#omitted same exporter statements as above
```

gumyr:
@jdegenstein would you please try this out again?  It works for me now with commit (cd834d1a16211e8b8613514deef7ecf8fbf791b7) but since fonts are platform dependent it still might be a problem.

jdegenstein:
Tested again on latest commit and can confirm this is resolved! Thanks everyone who contributed to the fix!

PR #445, opened 2023-12-23
An attempt at #432

KarimJedda:
Alright this might be a little bit of a stretch, but I believe the issue being floating point shenanigans. Since this required quite a bit of trial and error, I'll share my solution. If it is something acceptable, I can refine the PR, lint the code a bit and add a few tests, otherwise I'd be happy to see how you will solve it. Let me know what you prefer.

Debugging the two pieces of code mentioned in #432 , gives significantly different coordinates for the vertices, as well as creating quite a few more as you pointed out. 

To solve this, I've done three things:
- Used a `cKDTree` for all the spatial data structure stuff
- Introduced a tolerance level at which it's appropriate to merge two vertices
- Introduced a precision factor for the floating point coordinates 

The gist of the change involved reworking `_mesh_shape`, into shape (pun intended). 

There might certainly be a better solution to this, all tests are green so I think there should be no regressions. This also solves #311 

Here's the beautiful object, at last:
Screenshot from 2023-12-23 23-51-42

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Development install
 ```
 git clone https://github.com/gumyr/build123d.git
 cd build123d
-python3 -m pip install -e .
+python3 -m pip install -r requirements.txt
 ```
 
 Further installation instructions are available (e.g. Poetry, Apple Silicon) see the [installation section on readthedocs](https://build123d.readthedocs.io/en/latest/installation.html).
--- a/src/build123d/mesher.py
+++ b/src/build123d/mesher.py
@@ -101,6 +101,7 @@ import OCP.TopAbs as ta
 from OCP.TopLoc import TopLoc_Location
 
 from py_lib3mf import Lib3MF
+from scipy.spatial import cKDTree
 from build123d.build_enums import MeshType, Unit
 from build123d.geometry import Color, Vector
 from build123d.topology import Compound, Shape, Shell, Solid, downcast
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Development install
 ```
 git clone https://github.com/gumyr/build123d.git
 cd build123d
-python3 -m pip install -e .
+python3 -m pip install -r requirements.txt
```
jdegenstein, line 43: Can you revert this change? I just want to keep this PR focused on one issue.

jdegenstein:
Nice work!

Couple other things here: can you make sure to run through the black formatter? It looked to me like there were a few formatting differences before/after black. Not directly relevant to this PR, but we might want to add a section to the docs regarding contributions to set out our standard practices.

Also, I wonder what the appropriate tolerance for this is? Currently you have it set to 1e-5, but my guess is that might be too high. For the 0.91 degree test case in #432 could you provide the coordinates of the offending vertices that are removed? That might give us an idea if it should be e.g. 1e-5, 1e-6, or 1e-9. 

Lastly, it would be a good idea to add at least one test case of the former failing behavior in case it comes back.

gumyr:
I've got more testing to do but I think you're correct in that the problem is precision.  Mesher already does the elimination of redundant vertices so all I had to add was the following:
image

gumyr:
Thanks but the less complex change (cd834d1a16211e8b8613514deef7ecf8fbf791b7) has fixed the problem.

KarimJedda:
Fantastic, I'm very glad it worked out. I agree that your solution is much better, thank you both for the review!

PR #448, opened 2024-01-02
Update GroupBy to use the generic types

jermatic1:
The `Comparable` class isn't needed to fix the issue but fixed a pylance error


gumyr:
Thanks!

PR #449, opened 2024-01-02
Add support for lofting faces with inner wires

MatthiasJ1:
Automatically computes the optimal path between inner wires. Also refactors the original loft functionality so that the new version is about the same length.

```py
a = Rot(Z=20) * (Circle(5) - GridLocations(3,0,3,1)*Circle(1))
b = Pos(0,0,10) * Rot(Z=90) * a
c = Pos(0,0,20) * (Rectangle(10,10) - GridLocations(3,0,3,1)*Rectangle(2,8))
l1 = loft([a,b,c])
```
Screenshot


gumyr:
This is a nice bit of added functionality to loft, thank you!  I've got three requests before merging:
1) Minimally, add unit tests to cover all of the lines in this new loft implementation.  Ideally, there would be some tests of different configurations of holes to ensure the functionality is correct and stays that way.
2) Run pylint and ensure there are no findings for this code.  The pylint check was successful check in that it didn't drop below the pass threshold but doesn't indicate that things are getting worse.
3) As discussed on Discord, please add a small description of how the association between holes is done to the docstring.

gumyr:
Thank you!

PR #453, opened 2024-01-04
add `location_at` operator ^ for edge/wire, and tighten helix default tolerance

jdegenstein:
This PR enables e.g. `wire^0` to return Location which is a companion with `wire@0 -> position` and `wire%0 -> tangent`. In the process of creating the test for the above I discovered the default `make_helix` tolerance was `1e-5` which is too coarse, I tightened to `1e-9`.


jdegenstein:
`MaxSegment=2000` is not a cure-all but does give some more room before helixes start to break down, also the tightened tolerance may have other (hopefully beneficial) effects elsewhere. 

`Edge.make_helix(10,8001,50)` gives the following:
image

PR #462, opened 2024-01-06
Added FAQ about glob imports

barnabywalters:
Also clearfixed previous section to prevent weird layout issues

### DIFF
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -1,5 +1,5 @@
 #####################
-Tips & Best Practices
+Tips, Best Practices and FAQ
 #####################
 
 Although there are countless ways to create objects with build123d, experience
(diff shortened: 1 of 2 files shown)


gumyr:
Excellent, thank you.  If you don't mind, I have a minor request to go with this: please add a reference to your section here (https://build123d.readthedocs.io/en/latest/tutorial_lego.html#step-1-setup) where the current text reads: "There are over 100 python classes in build123d so we’ll just import them all with a from build123d import * but there are other options that we won’t explore here."

barnabywalters:
took me a while to figure out how to persuade sphinx to do that, but the link is in there now!

gumyr:
Done - thanks!

Issue #446, opened 2023-12-25
Algebra mode subtraction would not need to be limited to same dimension


voneiden:
Boolean subtraction has a built in limitation that checks that all `other` shapes have same dimension as `self`. 

https://github.com/gumyr/build123d/blob/cb5625cbe6bf1220ea82819632c21d6c9a394c58/src/build123d/topology.py#L1692-L1696

This is more limiting as necessary. OpenCascade documentation on [Boolean Operations Algorithm](https://dev.opencascade.org/doc/occt-7.4.0/overview/html/occt_user_guides__boolean_operations.html#occt_algorithms_9) states that 

> For Boolean operation Cut the minimal dimension of S2 should not be less than the maximal dimension of S1.

Further down the line it is elaborated a bit more:

> The result of the operation Cut is defined for arguments S1 and S2 that have values of dimensions Dim(S2) that should not be less than Dim(S1). The result can contain shapes of different dimensions, but the minimal dimension of the result will be equal to the minimal dimension of the objects Dim(S1). The result of the operation Cut12 is not defined for other cases. For example, it is impossible to cut an edge from a solid, because a solid without an edge is not defined.

Therefore, for example, subtracting a 3D shape from a 2D shape is from OpenCascade perspective. 

Should be a fairly straightforward fix. I'm open to making a PR if requested. 

Thoughts?

gumyr:
You are correct, the dimension check is overly limiting. The fix should be relatively easy; however, we'll need some unit tests to check limits like subtracting from a Vertex, etc.  A PR would be great, thanks for the offer.

PR #427, opened 2023-12-09
Generalize Shape.split to support arbitrary splitting surfaces

MatthiasJ1:

PR #465, opened 2024-01-06
Support multidimensional minus in algebra mode

voneiden:
Previously algebra minus had a check that verified all arguments were of same dimension. However OCCT supports cutting a shape with shapes that are of equal OR larger dimension. This commit modifies the check to support larger dimensions and adds tests for all dimension variations.

Ref #446


MatthiasJ1:
One note is you can in fact subtract one lower dimension, e.g 3d-2d or 2d-1d. The split operation (#427) is essentially subtracting 3d-2d with an extra step which separates the resulting parts based on their orientation relative to the splitting surfaces.

gumyr:
Thank you!

Issue #319, opened 2023-09-11
2D objects have a non-null volume


Jojain:
To reproduce : 
```py
>>> from build123d import *
>>> f = Face.make_rect(10,20)
>>> f.volume
199.99999999999997
```

Same with 2D shells

jdegenstein:
From the OCCT docs: "
<html>
<body>
<div class="memproto" style="font: bold 14px / 22px Roboto, sans-serif; border-top: 1px solid rgb(168, 184, 217); border-left: 1px solid rgb(168, 184, 217); border-right: 1px solid rgb(168, 184, 217); padding: 6px 0px; color: rgb(37, 53, 85); text-shadow: rgba(255, 255, 255, 0.9) 0px 1px 1px; background-color: rgb(223, 229, 241); box-shadow: rgba(0, 0, 0, 0.15) 5px 5px 5px; border-top-right-radius: 4px; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">

Standard_Real GProp_GProps::Mass | ( |   | ) | const
-- | -- | -- | -- | --


[...]

PR #488, opened 2024-01-19
Return 0.0 for 1D and 2D shapes

jdegenstein:
Per https://github.com/gumyr/build123d/issues/319

Last remaining question: should `Shape.volume` be eliminated and instead have `Mixin3D.volume` ? `Shape.compute_mass` would possibly be affected as well.

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -965,6 +965,14 @@ class Mixin1D:
 
         return return_value
 
+class Mixin2D:
+    """Methods to add to the Face and Shell classes"""
+
+    @property
+    def volume(self) -> float:
+        """volume - the volume of this Face or Shell, which is always zero"""
+        return 0.0
+
 
 class Mixin3D:
     """Additional methods to add to 3D Shape classes"""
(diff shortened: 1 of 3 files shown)


gumyr:
Nicely done. 

To address your question, `Shape.volume` probably should be removed as it's being replaced by custom volume calculations. How about this?
- `Vertex.volume` returns zero
- `Solid.volume` returns `Shape.compute_mass` 
- `Compound.volume` returns the sum of the volume of all `Solid` and `Shell` sub-shapes
- `Shell.volume` has some logic as shown below:

(i.e. Mixin3D doesn't have a `volume` property)

Currently:
```py
b = Box(1, 1, 1)
print(b.shell().is_manifold)
print(b.shell().volume)
print(b.shell().area)

s = Shell.make_shell(b.faces()[0:5])
print(s.is_manifold)
print(b.volume)
print(s.area)
```
```
True
6.0
6.0
False
0.9999999999999998
5.0
```
which is kinda messed up.  How about if a manifold `Shell` returns `Solid.make_solid(shell).volume` and a non-manifold shell returns zero?

PR #492, opened 2024-01-23
Small doc change: Area() to property area

asteppke:
Small changes to documentation 
- `Rectangle.Area()` does not (anymore?) exist but `Rectangle.area` is now a property.
- `spline` changed to `Spline` in examples


gumyr:
Thanks!

PR #510, opened 2024-01-29
Added Apple Silicon install instructions under Rosetta

shehabattia96:
A quick alternative for installing BD on Apple Silicon.


gumyr:
We have a solution to remove all the special instructions for Apple Silicon which should be available in a few days so hopefully this content won't be needed.

Issue #486, opened 2024-01-18
ValueError("polyline requires three or more pts")


MatthiasJ1:
Seems like an unnecessary restriction.

jdegenstein:
I agree. I think this should be changed.

gumyr:
`Polyline` with two points is just a `Line` which probably means that someone has made a mistake which is why the check is there. Why do you see it as "unnecessary"?

jdegenstein:
I see it as unnecessary because `Polyline` could be used in e.g. a programmatic setting (such as a python function) where the number of points is variable. Having to switch between `Line` and `Polyline` based on the number of points when the latter could generate a perfectly valid output seems like an arbitrary choice.

We haven't mentioned `FilletPolyline` which will fillet all "middle" points. Should `FilletPolyline` also change to a minimum of 2 points? I don't have a strong opinion either way.

MatthiasJ1:
@jdegenstein exactly. Also from a purely from a semantics point of view, a 2 point polyline is valid. It's the same for `FilletPolyline`.

jdegenstein:
Incorporated suggestions into the merged PR here https://github.com/gumyr/build123d/pull/516

PR #516, opened 2024-01-31
change Polyline and FilletPolyline to accept 2 pts as minimum

jdegenstein:
Per discussion at issue https://github.com/gumyr/build123d/issues/486


jdegenstein:
Incorporated suggestions into the merged PR here https://github.com/gumyr/build123d/pull/516

PR #552, opened 2024-02-16
Add Face.make_bezier_surface()

fanf2:
When making superellipses and superellipsoids with build123d, I have found I get the best results using Bézier curves and surfaces to approximate the shapes. Bézier curves are already supported, but Bézier surfaces were not - at least not without dropping down to OCP.

Here's a sketchy patch based on the code i'm using to construct Bézier surfaces. I have not tried installing build123d from source so I have not properly tested this myself, but I think it's a plausible start.

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -5718,8 +5718,7 @@ class Face(Shape):
             )
         if len(points) > 25 or len(points[0]) > 25:
             raise ValueError("The maximum number of control points is 25")
-        if weights && (len(points) != len(weights) or
-                       len(points[0]) != len(weights[0])):
+        if weights and (len(points) != len(weights) or len(points[0]) != len(weights[0])):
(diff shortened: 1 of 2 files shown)


src/build123d/topology.py
```diff
@@ -5689,6 +5690,54 @@ def make_surface_from_array_of_points(
...
+        Returns:
+            Face: a potentially non-planar face
+        """
+        if len(points) < 2 or len(points[0]) < 2:
+            raise ValueError(
+                "At least two control points must be provided (start, end)"
+            )
+        if len(points) > 25 or len(points[0]) > 25:
+            raise ValueError("The maximum number of control points is 25")
+        if weights && (len(points) != len(weights) or
```
jdegenstein: There is some invalid syntax here on 5721-, can you fix this?

Issue #544, opened 2024-02-08
geom_type() documentation lists incorrect return values


MatthiasJ1:

BogdanTheGeek:
@gumyr didn't #559 also fix this?

jdegenstein:
I believe this was fixed as observed by @BogdanTheGeek

Issue #466, opened 2024-01-07
Face.geom_type() returns string rather than GeomType


MatthiasJ1:
`Face.geom_type()` might also make more sense as `@property`.

```py
print(f"{Rectangle(1,1).face().geom_type() = }")
print(f"{GeomType.PLANE = }")
print(f"{Rectangle(1,1).face().geom_type()==GeomType.PLANE = }")
```
```
Rectangle(1,1).face().geom_type() = 'PLANE'
GeomType.PLANE = <GeomType.PLANE>
Rectangle(1,1).face().geom_type()==GeomType.PLANE = False
```

gumyr:
Duplicate of Issue #321

jdegenstein:
Fixed with PR #559
```py
from build123d import *
print(f"{Rectangle(1,1).face().geom_type = }")
print(f"{GeomType.PLANE = }")
print(f"{Rectangle(1,1).face().geom_type==GeomType.PLANE = }")
```
now returns
```py
Rectangle(1,1).face().geom_type = <GeomType.PLANE>
GeomType.PLANE = <GeomType.PLANE>
Rectangle(1,1).face().geom_type==GeomType.PLANE = True
```

PR #559, opened 2024-02-21
Return `GeomType` enum, from `Shape.geom_type()`

### LINKED ISSUE #321: Change Shape.geom_type() to use GeomType values
`GeomType` should be changed to a `StrEnum` and use the same values as generated by `Shape.geom_type()` so that `edge.geom_type() == GeomType.CIRCLE` works.

Should `Shape.geom_type` be a property instead of a method?

BogdanTheGeek:
Fixes: #321 and #466

Solution:
Return GeomType directly. Add type hints to LUTs

Compromises:
Cannot return arbitrary strings or Shapes from GeomType (Eg. Vertex, Component, Face etc.)
GeomType is now a str enum to maintain compatibility

Breaking changes:
 - geom_type is now a property by popular demand

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -3304,7 +3304,7 @@ class ShapeList(list[T]):
         elif isinstance(filter_by, GeomType):
 
             def predicate(obj):
-                return obj.geom_type == filter_by.name
+                return obj.geom_type == filter_by
 
         else:
             raise ValueError(f"Unsupported filter_by predicate: {filter_by}")
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -341,29 +341,29 @@ geom_LUT = {
 
 
 geom_LUT_FACE = {
-    ga.GeomAbs_Plane: "PLANE",
-    ga.GeomAbs_Cylinder: "CYLINDER",
-    ga.GeomAbs_Cone: "CONE",
-    ga.GeomAbs_Sphere: "SPHERE",
-    ga.GeomAbs_Torus: "TORUS",
-    ga.GeomAbs_BezierSurface: "BEZIER",
-    ga.GeomAbs_BSplineSurface: "BSPLINE",
-    ga.GeomAbs_SurfaceOfRevolution: "REVOLUTION",
-    ga.GeomAbs_SurfaceOfExtrusion: "EXTRUSION",
-    ga.GeomAbs_OffsetSurface: "OFFSET",
-    ga.GeomAbs_OtherSurface: "OTHER",
+    ga.GeomAbs_Plane: GeomType.PLANE,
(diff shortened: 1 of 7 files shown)


BogdanTheGeek:
Seeing how some of the LUTs are only used by geom_type(), should we make them "private" members of `Shape`?

src/build123d/topology.py
```diff
@@ -5125,8 +5082,10 @@ def project_to_shape(
 
     def to_axis(self) -> Axis:
         """Translate a linear Edge to an Axis"""
-        if self.geom_type() != "LINE":
-            raise ValueError("to_axis is only valid for linear Edges")
+        if self.geom_type() != GeomType.LINE:
+            raise ValueError(
+                "to_axis is only valid for linear Edges not " + self.geom_type()
```
Jojain: I think `f"to_axis is only valid for linear edges not edges of type {self.geom_type()} "` is better.

src/build123d/topology.py
```diff
@@ -3246,10 +3198,12 @@ class Comparable(metaclass=ABCMeta):
     """Abstract base class that requires comparison methods"""
 
     @abstractmethod
-    def __lt__(self, other: Any) -> bool: ...
+    def __lt__(self, other: Any) -> bool:
+        ...
```
Jojain: I don't know what formatting you ran but it might be good to run the same as the source repo so this kind of changes doesn't appear.

Jojain: Make `geom_type` a property would be nice aswell

BogdanTheGeek: I ran ruff, aka fast black. I didn't see anything about a formatter, only pylint and mypy

src/build123d/topology.py
BogdanTheGeek: wanted to see if people actually thought it was a good inclusion first, used it for debugging

src/build123d/topology.py
Jojain: I think Roger uses black. Maybe setting a standard if there isn't one could be great to avoid formatting changes !

gumyr: I use black.  Not sure how the `...` happened but don't worry about it.

gumyr: The property change is good.  There are a few other places where this needs to be done (e.g. `is_valid` IIRC).

BogdanTheGeek:
also fixes #544

gumyr (changes requested):
Functionally this looks great, thank you.  However, many of the changes in the topology.py file are formatting changes - I'm concerned that when I format with black there will be a whole bunch of changes to revert what was done here.  I assume this is a ruff difference, is there a way to avoid flipping back and forth?

PR #573, opened 2024-02-27
Add Color to allowable types for fill_color and line_color.

jrmobley:
Adds support for color opacity values to `ExportSVG` `fill_color` and `line_color`.


gumyr:
Thanks!

PR #589, opened 2024-03-16
Avoid a crash in Mesher if triangulation fails

openvmp:
Exactly the same problem exists in CadQuery's tessellate() (see [this issue](https://github.com/CadQuery/cadquery/issues/705)).

I don't know how to reproduce it on a small model.
It will be reproducible by the following PartCAD command after the refactoring for OBJ rendering is published (the refactoring is to use build123d instead of CadQuery):

```
pc render -t obj -a /pub/robotics/multimodal/openvmp/robots/don1:robot
```

### DIFF
--- a/src/build123d/mesher.py
+++ b/src/build123d/mesher.py
@@ -285,6 +285,10 @@ class Mesher:
         for facet in ocp_mesh.faces():
             # Triangulate the face
             poly_triangulation = BRep_Tool.Triangulation_s(facet.wrapped, loc)
+            if poly_triangulation is None:
+                warnings.warn("Face triangulation failed.")
+                # TODO(clairbee): Avoiding a crash. Need to determine a smarter way to handle this.
+                continue
             trsf = loc.Transformation()
             # Store the vertices in the triangulated face
             node_count = poly_triangulation.NbNodes()


gumyr:
If a face is skipped isn't the whole tessellation invalid?  How does this actually help or is it just intended to be a placeholder for some alternative solution?

bernhard-42:
I think I have the same in my tessellator:

https://github.com/bernhard-42/ocp-tessellate/blob/3940f04206f207bfdc4e5206590643f00ea984f8/ocp_tessellate/tessellator.py#L229-L234

OCCT returns None for faces without proper polygons

openvmp:
> How does this actually help or is it just intended to be a placeholder for some alternative solution?

It helps to avoid an exception being thrown a few lines further in case the object is None.
I don't know if there is anything else we can do with such faces.

gumyr:
@openvmp would you add some tests to cover this change?

openvmp:
I was only able to reproduce it while processing [that model](https://partcad.org/repository/assembly/robotics/multimodal/openvmp/robots/don1:robot).
It was producing a 1.2GB large OBJ file at that moment and it was very hard to troubleshoot.

I guess it should be relatively simple to reproduce, using a face with all vertices on one line or when two of them coincide.
But, unfortunately, I'm not proficient in build123d modeling to be able to reproduce that.
I'm serving those using build123d, but I don't know how to use it myself.
I only know how to use import/export functions.

gumyr:
Thanks for the proposed guard against failed triangulation. We’re going to close this PR.

Skipping an untriangulated face avoids the immediate exception, but produces an incomplete mesh. For 3MF export, that is not a safe recovery: the resulting mesh may be open or non-manifold while appearing to export successfully.

As there is no minimal reproducer or regression test, we do not have a reliable basis for changing this behavior. If this is revisited, the preferred approach would be a targeted reproducer and a clear error explaining that a face could not be triangulated, rather than silently omitting geometry.

PR #610, opened 2024-04-15
Created the PartCAD package for build123d examples. Added links.

openvmp:


gumyr:
What is the partcad.yml file for?  I understand what it does but do partcad users really want to build with any of the build123d examples?  What is the expectation for support of this going forward, do all new examples need to be included in this file?

openvmp:
It makes the examples "packaged".

Once they are packaged, they become available (in the "/examples" folder) to anyone browsing the public repo of PartCAD packages (on PartCAD.org or through the PartCAD VSCode extension) so that people can see how to use build123d to create reusable packages.

I don't anticipate people actually using them in their designs. Though I can imagine someone using the T-slot or the din rail. Also you can start using PartCAD to render examples for the documentation purposes (both render the parts into images and the package itself into the documentation page). However your own rendering of examples already works. Though it might make sense to push that functionality to PartCAD to make it reusable.

In terms of support, you get no obligation to maintain it operational as I will reference a specific commit in your repo. If and only if you make a new revision that is functioning well and you want to publish it, we'll have to make a pull request in the PartCAD public repo to publish the new commit id.

gumyr:
Merged.  I don't know if I or others will remember to update the yml file so it's good to know that keeping this up-to-date isn't a requirement.

Issue #428, opened 2023-12-09
RigidJoint and RevoluteJoint don't align correctly


bernhard-42:
https://discord.com/channels/964330484911972403/1074840524181217452/1183060124353433641

```python
b1 = Box(10, 10, 1)
b2 = Box(5, 5, 1)
j1 = RigidJoint("j1", b1, Pos(-4, -4, 0.5))
j2 = RevoluteJoint("j2", b2, Axis((2.5, 2.5, 0), (0, -1, 0)))

j1.connect_to(j2, angle=30)
show(b1, b2, render_joints=True)
```

Three issues I don't get:
1) Revolute axis ignored when adding an angle
2) Absolute positions not the same after connect_to
3) Wrong order of joints when I think about connecting like a carpenter does
Admittedly, the third one might be subjective, and is of minor importance. I can flip the order in my mind. But the first two are unclear to me

![image](https://github.com/gumyr/build123d/assets/954408/0a41293a-86e8-4f57-af13-2836b8eea5a6)

```python
print(j1.parent.location * j1.relative_location)
print(j2.parent.location * j2.relative_axis.location)
# Location: (position=(-4.00, -4.00, 0.50), orientation=(-0.00, 0.00, -0.00))
# Location: (position=(-4.00, -4.00, 0.50), orientation=(-90.00, -60.00, -0.00))
```

bernhard-42:
I added to `RevoluteJoint`, `LinearJoint`, `CylindricalJoint`
```python
    @property
    def absolute_location(self):
        """Absolute location of joint"""
        return self.parent.location * self.relative_axis.location
```

and to `RigidJoint`, `BallJoint`
```python
    @property
    def absolute_location(self):
        """Absolute location of joint"""
        return self.parent.location * self.relative_location
```

and replaced `_connect_to` in class `Joint` by
```python
    def _connect_to(self, other: Joint, **kwargs):  # pragma: no cover
        """Connect Joint self by repositioning other"""
        if kwargs.get("angle") is not None:
            transform = Rotation(0, 0, kwargs["angle"])
        else:
            transform = Location()

        self.parent.location = (
            other.absolute_location * transform * self.absolute_location.inverse()
        ) * self.parent.location
        self.connected_to = other

    @property
    @abstractmethod
    def absolute_location(self) -> Location:  # pragma: no cover
        """A CAD object positioned in global space to illustrate the joint"""
        raise NotImplementedError
```

[...]

PR #613, opened 2024-04-28
Add a test for revolute joints around an axis other than Z

zackyancey:
This is a test for the bug in #428 . Not sure if I should be targeting `dev` or `joints-v2`, since the test currently fails on `dev` and passes on the other (or if I should be marking this `xfail` or something).

### DIFF
--- a/tests/test_joints.py
+++ b/tests/test_joints.py
@@ -161,6 +161,28 @@ class TestRevoluteJoint(DirectApiTestCase):
             target_location.orientation, j2.location.orientation, 5
         )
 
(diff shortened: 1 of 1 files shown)


gumyr:
Thanks for adding this test.  I've fixed Issue #428 and tried this test out and it works so please rebase and submit on the dev branch.  Thanks again and sorry it took so long to fix the underlying problem.

zackyancey:
Alright, rebased and targeting dev, I ran the tests locally and they passed. Thanks!

gumyr:
Thanks!

PR #626, opened 2024-05-16
Fix solution ttt-ppp0101.py

ABoss:
Now generates correct model and calculates correct mass. (note that Trapezoid requires it's width input field to be calculated using angles because it's not given)

### DIFF
--- a/docs/assets/ttt/ttt-ppp0101.py
+++ b/docs/assets/ttt/ttt-ppp0101.py
@@ -4,6 +4,7 @@ Too Tall Toby Party Pack 01-01 Bearing Bracket
 
 from build123d import *
 from ocp_vscode import *
+import math
 
 densa = 7800 / 1e6  # carbon steel density g/mm^3
 densb = 2700 / 1e6  # aluminum alloy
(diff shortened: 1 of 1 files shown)


gumyr:
Thanks for pointing this out, there was a change in how Trapezoid works in that now the width is the overall width not the width of the "base" line which results in the problem.

Although your solution is correct, I try to avoid having the user do trig - how is this instead?
```py
        with BuildLine() as bl:
            l1 = Line((0, 0), (18 / 2, 0))
            l2 = PolarLine(l1 @ 1, 8, 60, length_mode=LengthMode.VERTICAL)
            l3 = Line(l2 @ 1, (0, 8))
            mirror(about=Plane.YZ)
        make_face()
```
i.e. creating a trapezoid from lines - note that no trig is required here.

gumyr:
Maybe there should be an option in the Trapezoid object to select either the maximum or minimum length (this could be fairly complex as the left and right angles could be different)?

ABoss:
Ah, much better solution! 

> Maybe there should be an option in the Trapezoid object to select either the maximum or minimum length (this could be fairly complex as the left and right angles could be different)?

Maybe, it would make working with Trapezoid easier (although maybe doing everything with lines is actually easier)

gumyr:
Thanks!

PR #632, opened 2024-05-21
Add `align_z` argument to pack() Method. 

roman-dvorak:
Hi, 

I've added a new argument `align_z` to the `pack()` method. When you enable this, it aligns all objects to the same Z axis.

#### Motivation and Context

I made this change to help with preparing print plates for 3D printing (set of multiple 3D bodies). The current `pack()` method only arranges objects in the X and Y coordinates. With the new `align_z` option, it becomes much easier to prepare objects for printing without needing to adjust them by a slicer.

#### How Has This Been Tested?

I did some manual testing to make sure the objects align correctly in the Z axis. See following images:

##### Without `z_align` (default)
```
parts = [box, lid]
p = pack(parts, padding=5, align_z=False)
show(p, reset_camera=Camera.KEEP)
```
obrazek
obrazek

##### With `z_align`
```
parts = [box, lid]
p = pack(parts, padding=5, align_z=True)
show(p, reset_camera=Camera.KEEP)
```
obrazek

Thank you!

### DIFF
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -172,6 +172,14 @@ padding (right):
   :align: right
 
 
+By default, the original Z value of all objects packed using the :meth:`pack.pack` function is preserved. 
+If you want to align all objects so that they are "placed" on the zero Z coordinate, the :meth:`pack` 
+function has an `align_z` argument. When set to `True`, this will align all objects. 
+
+This can be useful, for example, when preparing print setups for 3D printing, giving you full control 
+over this alignment so you don't have to leave it to the slicer.
+
+
 .. _are_glob_imports_bad_practice:
 
 ***********************************************
--- a/src/build123d/pack.py
+++ b/src/build123d/pack.py
@@ -120,7 +120,17 @@ def _pack2d(
 
 
 def pack(objects: Collection[Shape], padding: float, align_z: bool = False) -> Collection[Shape]:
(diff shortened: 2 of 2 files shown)


gumyr:
The original implementation didn't have the complete docstring - would you add the following (or with any improvements you can think of) to the `pack` function?
```py
    """Pack objects in a squarish area in Plane.XY.

    Args:
        objects (Collection[Shape]): objects to arrange
        padding (float): space between objects
        align_z (bool, optional): align shape bottoms to Plane.XY. Defaults to False.

    Returns:
        Collection[Shape]: rearranged objects
    """
```

Thanks for the submission!

roman-dvorak:
Thanks for the comment. I added it as you wrote it, I have no problem with it and it is correct. Thanks for the help.

I would also like to expand the documentation because the mention of `pack` function is quite hidden in between 'tips'. I would probably look for such information in the 'assemblies' section. So I would do that in a separate page and additional PR. Do you have any suggestions on where to place it in the documentation?

gumyr:
Knowing where folks look for stuff is on of the difficult parts of writing good documentation so if you think you'd look in Assemblies I'm okay with that.  Maybe a reference in the 3D exporters section would be helpful as well if the next step is to 3D print their objects?

I've merged the PR - thanks again!

PR #641, opened 2024-06-08
Update documentation for pack function

roman-dvorak:
Hi,
As I mentioned, I tried to expand the documentation by including documentation of the `pack` function.

### DIFF
--- a/docs/assemblies.rst
+++ b/docs/assemblies.rst
@@ -157,3 +157,96 @@ adds the following attributes to :class:`~topology.Shape`:
     Any iterator can be assigned to the ``children`` attribute but subsequently the children
     are stored as immutable ``tuple`` objects.  To add a child to an existing :class:`~topology.Compound`
     object, the ``children`` attribute will have to be reassigned.
+
+    .. _pack:
+
+******
+pack
+******
+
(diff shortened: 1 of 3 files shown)


gumyr:
One minor typo: "If you place the arranged objects into a ``Compose``, you can easily determine their bounding box and check whether the objects fit on your print bed." - it should be `Compound` not `Compose`.

What do you think about putting most of this documentation into the `pack` docstring instead of the documentation?  `autofunction` can extract all of the text from the docstring when building the docs and this extra info will be there for anyone who is reading the code directly or using an IDE to view the function.  The `ExportDXF` has example code in the docsting as follows:
```
    Example:

        .. code-block:: python

            exporter = ExportDXF(unit=Unit.MM, line_weight=0.5)
            exporter.add_layer("Layer 1", color=ColorIndex.RED, line_type=LineType.DASHED)
            exporter.add_shape(shape_object, layer="Layer 1")
            exporter.write("output.dxf")

```

Lastly, I'm not sure that the description of `_pack2d` is required in the docs as users don't need this info.  Maybe just as a comment in the code? 

Thanks for improving the docs!

gumyr:
Thanks!

jdegenstein:
@roman-dvorak It looks like the SVG assets here https://build123d.readthedocs.io/en/latest/assemblies.html#pack.pack were never added and as a result the links are broken. Could you open another PR and add them?

roman-dvorak:
Thank you for notification. I'm going to check it.

PR #643, opened 2024-06-19
Add major_radius bool to HexLocations and rename apothem parameter to radius

### LINKED ISSUE #633: Add major_radius bool to HexLocations
Currently `RegularPolygon` supports bool `major_radius` that defaults to `True`. In order to place hexagonal regular polygons with sides touching with the default option, one must use math to convert the major radius to minor radius (aka the apothem). I propose adding `HexLocations(... major_radius: bool = False)` which will preserve the existing default behavior, while allowing for better interoperability between `RegularPolygon` and `HexLocations`.

```py
with BuildSketch() as s:
    with HexLocations(1,3,3):
        RegularPolygon(1,6)
```
image

jdegenstein:
This PR fixes https://github.com/gumyr/build123d/issues/633 by adding a major_radius: bool to HexLocations that defaults to False (to preserve current behavior). This PR is technically breaking because the prior `apothem` parameter had to be named to a more generic `radius` which can be either a major radius or minor radius (aka apothem).

Here is an example of the new functionality:
```py
hex = RegularPolygon(1, 6)
with BuildSketch() as s:
    with HexLocations(1, 3, 3, False) as hloc:
        add(hex)
    with Locations((6,0)):
        with HexLocations(1, 3, 3, True) as hloc2:
            add(hex)
```

image

### DIFF
--- a/examples/heat_exchanger.py
+++ b/examples/heat_exchanger.py
@@ -48,7 +48,7 @@ with BuildPart() as heat_exchanger:
     tube_locations = [
         l
         for l in HexLocations(
-            apothem=(tube_diameter + tube_spacing) / 2,
+            radius=(tube_diameter + tube_spacing) / 2,
             x_count=exchanger_diameter // tube_diameter,
             y_count=exchanger_diameter // tube_diameter,
         )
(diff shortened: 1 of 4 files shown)


gumyr:
Looks good to me.  Your use case for this was something I didn't envision but is perfectly valid.  Two of the examples would be broken though.

jdegenstein:
Fixed the two examples and merged.

PR #647, opened 2024-06-24
Propose adding `Plane.isometric` to built-in planes

jdegenstein:
Currently there are a number of built-in planes like Plane.XY, YZ, front, top, etc. I propose adding a `Plane.isometric` built-in plane that will (1) allow aligning objects to the viewers isometric view direction (tested against OCP CAD Viewer and CQ-editor) (2) Provide access to a less trivial plane definition that could help improve robustness of unit tests (none of the `Plane.isometric` axes align with the global XYZ axes). There may also be some benefits to the 2D exporters, since providing an easier-to-access isometric view direction could be helpful for e.g. "more intelligible" 2D representations of 3D objects.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1895,6 +1895,12 @@ class PlaneMeta(type):
         """Bottom Plane"""
         return Plane((0, 0, 0), (1, 0, 0), (0, 0, -1))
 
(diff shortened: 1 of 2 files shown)


gumyr:
Looks good to me.  Thanks!

Issue #662, opened 2024-07-22
Adding a STEP-imported part fails with pickle error


NelisW:
Windows 11  
Python 3.12.3 in a conda environment set up for build123d   
`python                    3.12.3               h1d929f7_1`

Using the latest build123d github version as of about four weeks ago.    
`build123d                 0.5.1.dev62+g4561105          pypi_0    pypi`


```
import build123d as b3d
import ocp_vscode as ov
import numpy as np
ov.set_port(3939)
ov.set_defaults(reset_camera=ov.Camera.KEEP, ortho=True, black_edges=True,measure_tools=True)


snaps = b3d.import_step("Multiboard Tight Snap.step")
with b3d.BuildPart() as snap:
  b3d.add(snaps)
```

for the STEP file downloaded from here:  
https://thangs.com/designer/Keep%20Making/3d-model/Multiboard%20Tight%20Snap-974403

(Much!) earlier version of build123d worked for STEP files downloaded from the multiboard thangs pages.

As another test, I created a simple cube in FreeCAD and exported a STEP file using checked `Export invisible objects`, and AP214 standard, and got the same error.  So on at least two differently-sourced files the error remains.
[Unnamed-PartCube.zip](https://github.com/user-attachments/files/16328815/Unnamed-PartCube.zip)


Both resulted in this error message

```
$ python HAGreen.py                                                                                                                                  
Traceback (most recent call last):                                                                                                                   
  File "K:\WorkP\3dprint\3dprint\projects\kameelperd\homeassistantgreen\HAGreen.py", line 32, in <module>                                            
    b3d.add(snaps)                                                                                                                                   
  File "C:\Users\nwillers\AppData\Local\anaconda3\envs\b123d\Lib\site-packages\build123d\operations_generic.py", line 123, in add                    
    object_iter = [obj.moved(rotation) for obj in object_iter]                                                                                       
                   ^^^^^^^^^^^^^^^^^^^                                                                                                               
  File "C:\Users\nwillers\AppData\Local\anaconda3\envs\b123d\Lib\site-packages\build123d\topology.py", line 2448, in moved                           
    shape_copy: Shape = copy.deepcopy(self, None)                                                                                                    
                        ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                    
  File "C:\Users\nwillers\AppData\Local\anaconda3\envs\b123d\Lib\copy.py", line 143, in deepcopy                                                     
    y = copier(memo)                                                                                                                                 
[...]

PR #663, opened 2024-07-24
import_step: do not assign labels to Compound.for_construction

mbugert:
This is a proposed fix for #662.

After importing a STEP file, the `for_construction` attribute of `Compound`s contain an `OCP.TDF.TDF_Label`. This will make `Shape.__deepcopy__` fail because the label is not pickleable.
I removed the assignment of `TDF_Label`s to `for_construction` in the STEP import, and pass the label as a function argument to `build_assembly` instead.
I tried importing a few STEP files and it appears to work. WDYT?

### DIFF
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -32,7 +32,7 @@ license:
 import os
 from math import degrees
 from pathlib import Path
-from typing import TextIO, Union
+from typing import TextIO, Union, Optional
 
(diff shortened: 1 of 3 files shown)


gumyr:
Thanks!  I had originally expected to do more with the `TDF_Label` stored temporarily in the `for_construction` attribute but as you've shown it isn't really required.

PR #668, opened 2024-08-09
add black.yml format check github action, and reformat remaining files in repo that did not pass check

jdegenstein:
This will e.g. check new PR for black formatting and fail if it was not properly formatted. I chose to use `~= 24.0` as the version for now.


gumyr:
From what I understand projects put a requirement on using a white space enforcement tool to reduce the effort in accepting PRs as white space changes aren't flagged and therefore not reviewed.  However, so far I haven't found this to be a problem and I worry it may reduce PRs as there is yet another requirement to be met.  As an example of what might happen, there are a few long standing open PRs that require unit tests where the authors aren't willing to add them. 

I'm happy to have the files periodically cleaned up with black though.  Would this be an acceptable compromise?

jdegenstein:
Yes, I don't feel this is a critical issue to solve so I am OK with your compromise.

PR #677, opened 2024-08-29
Change Empty Builders RuntimeError to a UserWarning

jdegenstein:
Currently builder patterns like:
```py
with BuildPart() as p:
    with BuildSketch() as s:
        pass
    pass
```
result in `RuntimeError: sketch is None - BuildSketch didn't create anything`

I propose that this behavior is changed to a UserWarning instead of a RuntimeError, as this can be quite annoying to work around while developing a part that is only partially finished. I often resort to adding "dummy objects" like boxes and rectangles to ensure that the builders are not empty.

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -280,9 +280,7 @@ class Builder(ABC):
                 "Transferring object(s) to %s", type(self.builder_parent).__name__
             )
             if self._obj is None and not sys.exc_info()[1]:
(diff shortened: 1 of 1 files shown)


gumyr:
+1

jdegenstein:
Merged this, haven't seen it cause many issues in testing but am looking for any unintended consequences it may cause with OCP CAD Viewer and CQ-editor.

PR #680, opened 2024-08-30
fix location_at, position_at and tangent_at with PositionMode.LENGTH for Wire

### LINKED ISSUE #572: position_at doesn't get the correct position with Wires and PositionMode.LENGTH
```py
circle = Wire(
    [
        Edge.make_circle(1, start_angle=0, end_angle=180),
        Edge.make_circle(1, start_angle=180, end_angle=360),
    ]
)
p1 = circle.position_at(pi, position_mode=PositionMode.LENGTH)
```
image

danieledapo:
Apparently Wire use a different parametrization than `Edge` because `Edge` has its `LastParameter` being the length of the Edge while `Wire`has `LastParameter` being the numer of `Edges` with the fractional part being the parameter inside the edge.

The way the patch works is to always divide the distance provided by the user by the real length of the Wire/Edge so that we get a parameter in [0..1] which we can then pass straight to `param_at` to get the real parameter. 

Fixes https://github.com/gumyr/build123d/issues/572

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -444,7 +444,7 @@ class Mixin1D:
             if position_mode == PositionMode.PARAMETER:
                 parameter = self.param_at(position)
             else:
-                parameter = position
+                parameter = self.param_at(position / self.length)
         else:
             try:
                 pnt = Vector(position)
@@ -682,7 +682,7 @@ class Mixin1D:
         if position_mode == PositionMode.PARAMETER:
             param = self.param_at(distance)
         else:
-            param = distance
+            param = self.param_at(distance / self.length)
 
         return Vector(curve.Value(param))
 
@@ -733,7 +733,7 @@ class Mixin1D:
         if position_mode == PositionMode.PARAMETER:
             param = self.param_at(distance)
         else:
-            param = distance
+            param = self.param_at(distance / self.length)
 
         law: GeomFill_TrihedronLaw
         if frame_method == FrameMethod.FRENET:
--- a/tests/test_direct_api.py
+++ b/tests/test_direct_api.py
@@ -1989,6 +1989,33 @@ class TestMixin1D(DirectApiTestCase):
         )
         self.assertTrue(all([0.0 < v < 1.0 for v in point]))
 
+        wire = Wire([Edge.make_line((0, 0, 0), (10, 0, 0))])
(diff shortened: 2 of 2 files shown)


danieledapo:
Note: this makes the `Edge` version slower because we're doing calculation that it would not need, but I do not know if it's worth it to have something like the following in each method

```python
if isinstance(self, Wire):
    parameter = self.param_at(distance / self.length)
else: # edge
   parameter = distance
```

but let me know what you think.

EDIT: this doesn't work because `Edge.make_circle(2)` has the same parametrization as of `Edge.make_circle(1)` or any other circle of any radius (it likely uses the start and end angle as the first and last parameter) which means we still have to do the division.

gumyr:
Without any changes to the code, consider this test:
```py
circle_wire = Wire(
    [
        Edge.make_circle(1, start_angle=0, end_angle=180),
        Edge.make_circle(1, start_angle=180, end_angle=360),
    ]
)
p1 = circle_wire.position_at(pi, position_mode=PositionMode.LENGTH)
p2 = circle_wire.position_at(pi / circle_wire.length)
print(p1, p2)

circle_edge = Edge.make_circle(1)
p3 = circle_edge.position_at(pi, position_mode=PositionMode.LENGTH)
p4 = circle_edge.position_at(pi / circle_edge.length)
print(p3, p4)
```
```
Vector(-0.9026853619331, -0.4303012170001, 0) Vector(-1, 0, 0)
Vector(-1, 0, 0) Vector(-1, 0, 0)
```
From this I conclude that the Edge case is currently correct and the Wire case is incorrect for `PositionMode.LENGTH`. Do we agree on this?

If so, `position_at`, `tangent_at` and `location_at` need different implementations for the Edge and Wire classes; therefore, these methods shouldn't be in the `Mixin1D` class.

danieledapo:
> From this I conclude that the Edge case is currently correct and the Wire case is incorrect for `PositionMode.LENGTH`. Do we agree on this?
> 

Unfortunately, this test passes on my machine so I'm not quite sure how to debug it. I've committed and pushed it, let's see what the CI says.

image

gumyr:
I erroneously recalled that the `param_at`  was really expensive but when I measured it that isn't so - looks like a nice fix.  Thanks!

Issue #689, opened 2024-09-07
Pack demo assets not showing in docs


jdegenstein:
on this page some of the links are broken to the SVG files https://build123d.readthedocs.io/en/latest/assemblies.html#pack

jdegenstein:
@roman-dvorak can you add the missing SVG files for the pack function? https://github.com/gumyr/build123d/pull/641

roman-dvorak:
I apologize for the delay. Images are attached in the new PR. Thanks for pinging me again.

jdegenstein:
Closed by https://github.com/gumyr/build123d/commit/1c653f87b939e61ef6b72adef98d582f58dbcef3 thanks again!

PR #690, opened 2024-09-07
Add missing svg files in assembly page

roman-dvorak:
To solve #689 and repair #641


jdegenstein:
Thanks @roman-dvorak for your contribution, this is a valuable feature and the documentation is very helpful too!

PR #695, opened 2024-09-11
Fix thicken normal calculation when no override is provided

### LINKED ISSUE #623: Thicken normal incorrect
```py
l1 = JernArc((0,0), (-1,0), 1, 180).edge().reversed()
l2 = JernArc((0,0), (1,0), 2, -90)
s = RadiusArc((0,0), (0,-1), 1)
f = sweep((l1^0)*s, l1) + sweep((l2^0)*s, l2) # Workaround for #622
```

```py
p = thicken(f, 0.4)
```

danieledapo:
Do not calculate normals if the user did not provide any override, it makes the calculation coherent with what OCCT does.

Fixes https://github.com/gumyr/build123d/issues/623.

### DIFF
--- a/src/build123d/operations_part.py
+++ b/src/build123d/operations_part.py
@@ -592,12 +592,12 @@ def thicken(
     logger.info("%d face(s) to thicken", len(to_thicken_faces))
 
     for face in to_thicken_faces:
-        normal_override = (
+        face_normal = (
             normal_override if normal_override is not None else face.normal_at()
         )
         for direction in [1, -1] if both else [1]:
             new_solids.append(
-                face.thicken(depth=amount, normal_override=normal_override * direction)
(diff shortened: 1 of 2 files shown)


gumyr (changes requested):
The real problem here is that the normal_override parameter is being changed for subsequent faces.  Here is a smaller change that just renames the variable:
```py
    for face in to_thicken_faces:
        face_normal = (
            normal_override if normal_override is not None else face.normal_at()
        )
        for direction in [1, -1] if both else [1]:
            new_solids.append(
                face.thicken(depth=amount, normal_override=face_normal * direction)
            )
```
I'd like to go this this less complex change.

danieledapo:
good catch! should be ok now!

gumyr:
Thanks!

PR #713, opened 2024-09-25
New section "Iterating Over Compounds" to Assembly documentation.

slobberingant:
Iterating over compounds like this has been a solution to a few of my previous problems and I haven't seen it in the documentation. Hope it helps.

### DIFF
--- a/docs/assemblies.rst
+++ b/docs/assemblies.rst
@@ -164,7 +164,7 @@ adds the following attributes to :class:`~topology.Shape`:
 Iterating Over Compounds
 ************************
 
-As Compounds are containers for shapes, Build123 can iterate over these as required.
+As Compounds are containers for shapes, build123d can iterate over these as required.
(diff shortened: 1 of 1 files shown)


gumyr (changes requested):
One minor change if you don't mind.  The `sum` doesn't need the list comprehension, if its implemented as `sum(part.volume for part in assembly3.solids())` it is called a **generator expression** and is the recommended way to do sums like this.

docs/assemblies.rst
```diff
@@ -160,6 +160,28 @@ adds the following attributes to :class:`~topology.Shape`:
 
     .. _pack:
 
+************************
+Iterating Over Compounds
+************************
+
+As Compounds are containers for shapes, Build123 can iterate over these as required.
```
jdegenstein: please use "build123d" -- all lowercase as per convention in Python

slobberingant:
Feedback has been incorporated. Thank you.

gumyr:
Thanks!

PR #718, opened 2024-09-28
Add import_svg_as_forced_outline function 

BlueDrink9:
Make importing unclean outlines easier. Works really well on the KiCad exports of two unrelated PCBs that I tested it on.

### DIFF
--- a/docs/import_export.rst
+++ b/docs/import_export.rst
@@ -261,6 +261,7 @@ For example:
 
 .. autofunction:: import_svg
 .. autofunction:: import_svg_as_buildline_code
+.. autofunction:: import_svg_as_forced_outline
 
 3D Importers
 ============
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -29,8 +29,9 @@ license:
 # pylint has trouble with the OCP imports
 # pylint: disable=no-name-in-module, import-error
 
+import copy
 import os
-from math import degrees
+from math import degrees, sqrt
 from pathlib import Path
 from typing import TextIO, Union, Optional
 
@@ -66,10 +67,14 @@ from OCP.XCAFDoc import (
     XCAFDoc_DocumentTool,
 )
 from ocpsvg import ColorAndLabel, import_svg_document
-from svgpathtools import svg2paths
+import svgpathtools as svg
 import unicodedata
 
+from build123d.build_enums import CenterOf, Mode, AngularDirection
+from build123d.build_line import BuildLine
 from build123d.geometry import Color, Location
+from build123d.objects_curve import Line, EllipticalCenterArc
+from build123d.operations_generic import add, offset, mirror
 from build123d.topology import (
     Compound,
     Edge,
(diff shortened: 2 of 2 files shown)


BlueDrink9:
Wait, noticed a bug with the second outline

gumyr:
Here are a few high level things to consider:
- I wonder if this functionality wouldn't be better suited to the ocpsvg project (https://github.com/snoyer/ocpsvg) which provides almost all of the SVG import capabilities of build123d.
- The contribution rules are here: https://github.com/gumyr/build123d/blob/dev/CONTRIBUTING.md which covers some of the following.
- Unittests will need to be provided with coverage at (or very close to) 100% 
- pylint needs to be happy with the code
- The content following line 617 shouldn't be included here (looks like the start of some tests)
- There shouldn't be any print statements (ValueError takes a string)
- Tests aren't running this morning so I can't check, but this code needs to run on Windows, Linux, Mac (both Intel and Arm architectures)

BlueDrink9:
Sure, might not get around to fulfilling all of those requests in the near future, but I'll aim to following the guidelines. Apologies for missing those the first time.

Regarding ocpsvg, currently I'm using b123d features to connect the wires cleanly, so I think that would be more work than I'm willing and able to put in

snoyer:
I've not studied the whole thing but it looks like what this PR accomplishes is pretty close to, or at least overlaps with, existing functionality in `build123d` like `make_wire()` that makes wire from unorganized edges.
If my understanding is correct I don't think that's in `ocpsvg`'s scope and it would be more appropriate (and more convenient/easier) to implemented it _after_ the SVG import at the `build123d` level

BlueDrink9:
I'll have to read the source for make wire. All I know is that it didn't work for the faces I imported, and this makes some notable changes to the imported that I wouldn't want made by default for all shapes

gumyr:
> I've not studied the whole thing but it looks like what this PR accomplishes is pretty close to, or at least overlaps with, existing functionality in `build123d` like `make_wire()` that makes wire from unorganized edges. If my understanding is correct I don't think that's in `ocpsvg`'s scope and it would be more appropriate (and more convenient/easier) to implemented it _after_ the SVG import at the `build123d` level

IIRC the SVG import crashes so there is no opportunity to fix the broken wires.

gumyr:
Another way to approach this problem might be to enhance the SVG importer to enable a request for just Edges with no attempt to build Wires or Faces.  With just the Edges, the `Wire(random_edges, sequenced=True)` could used (or possible enhanced) to build the perimeter of a Face for these poorly formed SVG files.

snoyer:
> IIRC the SVG import crashes so there is no opportunity to fix the broken wires.

Is there any data to reproduce this crash?

jdegenstein:
@BlueDrink9 could you provide a sample SVG that reproduces the bad behavior you have observed?

gumyr:
Although the implementation needs some work this PR illustrates a need for the ability to make imports of SVGs and Fonts more robust as this represents some of the more common problems users have. A new Issue has been created for this work, Issue #1403, where a new design is proposed to handle both of these problematic sources. As such, this PR is closed but not forgotten.

PR #733, opened 2024-10-10
Remove mesh from `bounding_box` and implement `optimal=False` in a few more places

### LINKED ISSUE #705: Use optimal=False in bounding_box size calculations where applicable
In several places in the codebase `bounding_box().diagonal` is used to calculate the max object size. Worth investigating if these occurrences can be replaced by `bounding_box(optimal=False).diagonal` to improve performance as this can be computationally intensive otherwise.

jdegenstein:
Fixes https://github.com/gumyr/build123d/issues/705 in a few places that will improve performance and not impact stability

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -203,7 +203,7 @@ class Builder(ABC):
     @property
     def max_dimension(self) -> float:
         """Maximum size of object in all directions"""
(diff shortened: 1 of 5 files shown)


jdegenstein:
Thanks to @bernhard-42 for his contributions to this.

Issue #457, opened 2024-01-05
Add Align.None


MatthiasJ1:
When used, the created object would simply skip the step of aligning according to its bounding box. An example use would be when specifying coordinates with `Polygon`

gumyr:
This is already supported, albeit in a slightly different form: `Polygon((0, 0), (1, 0), (1, 1), align=None)` operates exactly as you describe and skips all the alignment code.  Unfortunately, this feature isn't well documented. The `Align` Enum could be expanded to include a `NONE` option as well - should there be redundant mechanisms for doing this?

MatthiasJ1:
Ah thanks, `align=None` fine, like you said it just needs to be documented. One way to do this could be:

```py
Align(Enum):
  CENTER=auto()
  ...
  NOALIGN=None
```

gumyr:
The type of `align` in the `BoundBox.to_align_offset` method is incorrect and should be fixed as part of this issue.

PR #734, opened 2024-10-13
Add an "Align.NONE" enum member.

smurfix:
This option means "leave me alone".
It is necessary for placing text on its baseline.


gumyr:
A few lines of unittests are required - patch coverage should be 100%.

gumyr:
Oh, how about changing `elif self.align[i] == Align.NONE:` to `elif self.align[i] == Align.NONE or self.align[i] is None:`?

smurfix:
I can do that. However … having the same set of tests in no less than five different locations is a red flag. Time to refactor that instead of adding more repetitive code (and tests for it).

Ideas for a good function name? I tend towards `align_offset(min: Tuple[float], max: Tuple[float], align: Tuple[Align]|Align|None) -> VectorLike`.

gumyr:
What are you proposing to change?

smurfix:
> What are you proposing to change?

This code
```
            align_offset=[]
            for i in however_many_axes:
                if align[i] == Align.MIN:
                    align_offset.append(min_coord()[i])
                elif align[i] == Align.CENTER:
                    align_offset.append((min_coord()[i]+max_coord()[i])/2)
                elif align[i] == Align.MAX:
                    align_offset.append(max_coord()[i])
                elif align[i] == Align.NONE:
                    align_offset.append(0)
                # note no check for any other value
```
appears five times in build123d's codebase. That's four times too many. Also I'd like to calculate the coordinates once each instead of up to three times, plus I don't like to rely on typechecking to prevent results like the one you get if you `align=(None,Align.MAX)`.

gumyr:
Sure, that code could be improved. There is an related issue https://github.com/gumyr/build123d/issues/457 - are you planning on fixing this too?

smurfix:
Thanks for the pointer. Yes.

gumyr:
This change was completed in PR #782

Issue #752, opened 2024-11-01
Dimensional check is incorrect in Compound.__sub__(), other possible latent problems


eecharlie:
```
from build123d import *
shape1 = Pos(0,0,0) * Box(30,30,30)
shape2 = Pos(30,0,0) * Box(50,5,5)
shape1 - shape2                                     # This is fine and produces a box with a hole
Compound(shape1) + Compound(shape2)                 # This is fine and produces a box on a stick
Compound([shape1,shape3]) + Compound(shape2)        # This works
sl = (ShapeList([shape1]) - ShapeList([shape2]))[0] # Executes, but just returns shape1, why?

Compound(children=[shape1]) - Compound(children=[shape2])   # This throws an error, why?
Compound(shape1) - Compound(shape2)                     # This throws an error, why?

# Error: TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
# when the dimension of self & other are checked
```

There are two problems quickly apparent in reviewing the source code for `__sub__()`:
```
        for _other in others:
            if type(_other)._dim < type(self)._dim:
                raise ValueError(
                    f"Only shapes with equal or greater dimension can be subtracted: "
                    ...
```
The error language is inconsistent with the actual conditional, in that the conditional should be `<=`, though this is not why the above example fails. The example fails because `type(Compound(Box(1,1,1)))._dim` is `None` when I think it should be `3`. 

The reason that this bug exists in `__sub__` and not in `__add__` is that add only checks for *equality* of dimension, and `None == None`. This is a significant separate latent issue. Here, fixing the conditional in `__sub__` in a careful way will allow it work, but at that point both `__add__` and `__sub__` and possibly other operations on `Compound` are silently failing to actually check dimensional compatibility of their arguments.

--> My plan is to fix the trivial bug in `__sub__()`'s dimension check such that it allows equality, knowing that this reproduces the latent failure in common with `__add__()`, but unblocks a use case for me. In a separate PR/issue, the fact that `Compound`'s dimension is `None` should be fixed.

eecharlie:
Opened #754

gumyr:
A Compound is a collection of any Shape; one can build a Compound of mixed 2D and 3D shapes so the `_dim` of Compound is explicitly set to None.  The `Part`, `Sketch` and `Curve` versions of `Compound` are intended to be used for collections of 3D, 2D and 1D Shapes respectively.

However, your other findings are worth investigation.

eecharlie:
[...]

PR #754, opened 2024-11-01
Compound sub dim check

eecharlie:
This addresses the immediate problem in #752 , and adds a narrow test of the behavior.

It does not address the fact that `Compound`s have `._dim = None`.

----
Update after better understanding the intention of the codebase: `Compound` is not meant to support add/subtract, it's an awkward part of the inheritance hierarchy. So this PR instead raises an informative error if add/subtract are attempted on Compound objects, to save others confusion. 

-----
Update after understanding longer-term roadmap of supporting these operations with `Assembly` objects: `Compound` must support add/subtract, and in case it contains many objects, for performant execution the operation should be consolidated into a single call to OpenCASCADE and not a naive nested for loop over every pair of shape primitives in `self` and `other`.

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -1697,37 +1697,76 @@ class Shape(NodeMixin):
         return result
 
     def __add__(self, other: Union[list[Shape], Shape]) -> Self:
-        """fuse shape to self operator +"""
-        # if type(self) == Compound:  # Only block generic Compounds
-        #     raise NotImplementedError(
-        #         "Addition not supported for generic Compounds. Use Part for 3D objects, "
-        #         "Sketch for 2D objects, or Curve for 1D objects instead."
-        #     )
+        """Generalized fuse shape to self operator +"""
 
         others = other if isinstance(other, (list, tuple)) else [other]
 
-        if not all([type(other)._dim == type(self)._dim for other in others]):
-            raise ValueError("Only shapes with the same dimension can be added")
+        # Only do dimension check for non-Compound shapes
+        # or dimensional Compounds (Part, Sketch, Curve)
+        if self._dim is not None:
+            for _other in others:
+                # Skip check if other is a generic Compound 
+                if _other._dim is not None and type(_other)._dim != type(self)._dim:
+                    raise ValueError(
+                        f"Only shapes with the same dimension can be added: "
+                        f"not {type(self).__name__} ({type(self)._dim}D) and "
+                        f"{type(_other).__name__} ({type(_other)._dim}D)"
+                    )
 
+        new_shape = None
         if self.wrapped is None:
             if len(others) == 1:
                 new_shape = others[0]
             else:
-                new_shape = others[0].fuse(*others[1:])
+                new_shape = others[0].fuse(*others[1:]) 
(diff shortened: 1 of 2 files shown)


gumyr:
We can do better then just returning an error for objects wrapped in `Compound`; if we replace `self` with `minuend` (the mathematical term which is used to avoid modifying `self`) and "unwrap" the `Compounds` like this:
```py
    def __sub__(self, other: Shape) -> Self:
        """cut shape from self operator -"""
        minuend = self.unwrap(fully=True) if isinstance(self, Compound) else self
        others = other if isinstance(other, (list, tuple)) else [other]
        others = [
            o.unwrap(fully=True) if isinstance(o, Compound) else o for o in others
        ]
        ...
```
more of the test cases you created in #752 pass.  `Compound` is used for assemblies as well - a new Assembly class will be introduced before release 1.0.0 to help clarify - which complicates this situation.

By the way, there are expectation for contributions listed here: https://github.com/gumyr/build123d/blob/dev/CONTRIBUTING.md.  Also, please don't delete unit tests.

eecharlie:
Ah, thanks. Apologies for _not_ reviewing the detailed contributing guidelines before pushing the PR.

I was making changes to topology.py in a rush and may have goofed up what I changed, I'm going to review that now; but where did I delete unit tests? My intent was only to _add_ two unit tests confirming an error is raised when add/sub are applied to a `Compound`. 

Regarding `minuend = self.unwrap...` I'll need to wrap my head around that, I don't yet understand wrapping & unwrapping in enough detail to understand what it would do here.

gumyr:
The changes in white space seem have to messed up the diff so some of the unittests where shown as deleted; however, I don't see that now - don't know why.

eecharlie:
Changed __sub__() logic to accommodate generalized Compound objects, and condensed to single call into OpenCASCADE, which appears to speed things up at least a little and sometimes ~10x. 
One failing test of STEP export.
Feedback welcome @gumyr before I make similar changes to __add__() or worry about why the export test broke

eecharlie:
Have run lint, black, mypy. A couple new complaints from mypy around return type hints, and linter complaints about the number of conditionals. Seems on par with existing code. Requesting workflow approval & review.

gumyr:
I'm sorry to say this now that you've put so much effort into it but this PR will not be accepted.  All the maintainers have discussed this problem as it's at the heart of build123d functionality and to fully solve the problem there are other changes required.

By the way, when creating unittests assertions must be made against the results - it's not sufficient to just have the code run to completion.

We all appreciate the effort you have put into improving build123d. Thank you.

eecharlie:
Thank you for letting me know. I will use my fork that provides a short-term resolution to this issue for my own application. 

If there is some way I can contribute, either to code or documentation, let me know. Is there a documented roadmap for the project?

gumyr:
Thanks for working on the issue - hopefully those operators do everything you need now.

We're trying to work towards a 1.0.0 release and have 17 open issues - select Milestones->Gating Release 1.0.0 to see them which is the closest we have to a roadmap at this point.  If you'd like to work on an Issue (gating the release or not) we'd be grateful, just let me know first.

Issue #756, opened 2024-11-04
Loft produces unexpected faces of GeomType.BSPLINE or filter_by(Plane) should work with GeomType.BSPLINE shapes


dalibor-frivaldsky:
The following code creates the shape in the picture below:
```
with BuildPart() as mount:
    with BuildSketch():
        Rectangle((1+16+4)*mm, 20*mm, align=(Align.MIN, Align.CENTER))
    with BuildSketch(Pos(1*mm, 0, 4*mm)):
        Rectangle(16*mm, 20*mm, align=(Align.MIN, Align.CENTER))
    loft()
```

<img width="860" alt="Screenshot 2024-11-04 at 14 49 17" src="https://github.com/user-attachments/assets/6c934b55-bc7b-43ff-88ff-96560a2c1a1e">

I would like to filter out the trapezoidal faces with `mount.faces().filter_by(Plane.XZ)`, however, this returns an empty list.

I have narrowed down the issue to the implemenation of the `Shape.filter_by` method, specifically the `plane_parallel_predicate`. The shape being filtered needs to be both an instance of `Face` and of a `GeomType.PLANE`. The `loft` creates the trapezoidal faces as `GeomType.BSPLINE`, which fails the second part of the check and results in the empty list after applying the filter.

I don't know whether the issue is with the `loft` creating the faces as b-splines (which maybe seems unnecessary?) or whether the `filter_by` function should be able to work with (plannar?) b-spline faces. What do you think?

gumyr:
Unfortunately, the OCCT CAD kernel returns these BSPLINE faces where a PLANAR face would be more appropriate. The check for planar needs to be change from just looking at the `geom_type` property to something like this:
```py
from build123d import *
from OCP.BRep import BRep_Tool
from OCP.GeomLib import GeomLib_IsPlanarSurface
from ocp_vscode import show_all


def is_face_planar(face: Face) -> bool:
    surface = BRep_Tool.Surface_s(face.wrapped)
    is_planar = GeomLib_IsPlanarSurface(surface, 1e-6)
    return is_planar.IsPlanar()


with BuildPart() as mount:
    with BuildSketch():
        Rectangle((1 + 16 + 4) * MM, 20 * MM, align=(Align.MIN, Align.CENTER))
    with BuildSketch(Pos(1 * MM, 0, 4 * MM)):
        Rectangle(16 * MM, 20 * MM, align=(Align.MIN, Align.CENTER))
    loft()

for f in mount.faces():
[...]

PR #764, opened 2024-11-05
Plane instantiation from planar Geom_BoundedSurface faces Issue #756

dalibor-frivaldsky:
This PR addresses additional problems with #756, specifically instantiating a Plane from planar faces that are not of type Geom_Plane.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -62,7 +62,6 @@ from OCP.BRepMesh import BRepMesh_IncrementalMesh
 from OCP.BRepTools import BRepTools
 from OCP.Geom import Geom_BoundedSurface, Geom_Line, Geom_Plane
 from OCP.GeomAPI import GeomAPI_ProjectPointOnSurf, GeomAPI_IntCS, GeomAPI_IntSS
-from OCP.GeomLib import GeomLib_IsPlanarSurface
 from OCP.gp import (
     gp_Ax1,
     gp_Ax2,
@@ -2093,8 +2092,7 @@ class Plane(metaclass=PlaneMeta):
         elif arg_face:
             # Determine if face is planar
             surface = BRep_Tool.Surface_s(arg_face.wrapped)
-            is_surface_planar = GeomLib_IsPlanarSurface(surface, TOLERANCE).IsPlanar()
-            if not is_surface_planar:
+            if not arg_face.is_planar:
                 raise ValueError("Planes can only be created from planar faces")
             properties = GProp_GProps()
             BRepGProp.SurfaceProperties_s(arg_face.wrapped, properties)
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -60,8 +60,9 @@ from OCP.BRepBuilderAPI import BRepBuilderAPI_MakeFace
 from OCP.BRepGProp import BRepGProp, BRepGProp_Face  # used for mass calculation
 from OCP.BRepMesh import BRepMesh_IncrementalMesh
 from OCP.BRepTools import BRepTools
-from OCP.Geom import Geom_Line, Geom_Plane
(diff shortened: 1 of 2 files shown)


gumyr:
There is no issue or description of what problem this change is attempting to fix.  Please describe the problem first.

dalibor-frivaldsky:
@gumyr apologies, I assumed the link to the issue in the commit message would be sufficient for provding the context.

This addresses additional issues with loft() creating planar faces of different type than Geom_Plane (or of GeomType.Plane), as described in #756 . I've provided description of these related issues after you've closed the issue. I thought it better to reopen the closed issue, but I can open a new one if this is the prefered way.

src/build123d/geometry.py
```diff
@@ -2092,18 +2093,20 @@ def optarg(kwargs, name, args, index, default):
...
             BRepGProp.SurfaceProperties_s(arg_face.wrapped, properties)
             self._origin = Vector(properties.CentreOfMass())
-            self.x_dir = (
-                Vector(arg_x_dir)
-                if arg_x_dir
-                else Vector(
-                    BRep_Tool.Surface_s(arg_face.wrapped).Position().XDirection()
-                )
-            )
+            if isinstance(surface, Geom_BoundedSurface):
```
gumyr, line 2100: Why is the special case of `Geom_BoundedSurface` handled here?

src/build123d/geometry.py
```diff
@@ -2092,18 +2093,20 @@ def optarg(kwargs, name, args, index, default):
...
+                point = gp_Pnt()
+                face_x_dir = gp_Vec()
+                tangent_v = gp_Vec()
+                surface.D1(0.5, 0.5, point, face_x_dir, tangent_v)
+            else:
+                face_x_dir = surface.Position().XDirection()
```
gumyr, line 2106: The important ability to force the x direction has been removed and needs to be returned.

tests/test_direct_api.py
```diff
@@ -2556,6 +2556,17 @@ def test_plane_init(self):
         with self.assertRaises(TypeError):
             Plane(Edge.make_line((0, 0), (0, 1)))
 
+        # can be instantiated from planar faces of different surface types
```
gumyr: Tests must make assertions to validate the results and not just run to completion.

gumyr (changes requested):

src/build123d/geometry.py
dalibor-frivaldsky: The added test to this PR fails otherwise when trying to instantiate a Plane from the trapezoid faces created by the loft(), since the underlying kernel creates them as Geom_BSplineSurface and not Geom_Plane. Initially I thought of making the special case for Geom_BSPlineSurface specifically, but there are other surface types, like Geom_BezierSurface. These can be planar as well and it should be possible to create a Plane object from those too. The Geom_BoundedSurface is a base class of these and makes the implementation more general, that's why I opted for it.

src/build123d/geometry.py
dalibor-frivaldsky: I don't think that is the case, the line 2109 retains the ability to force the x direction from the argument to the Plane constructor. All the tests pass so I suppose such a regression would have been caught by them, unless I overlooked something.

tests/test_direct_api.py
dalibor-frivaldsky: I've extended the test to also verify properties of the two trapezoid faces, which are created as Geom_BSplineSurface by loft() and are the focus of the introduced changes.

dalibor-frivaldsky:
I've updated the Plane instantiation code to use the new is_planar property of the Face class and added missing tests.

gumyr:
My mistake about forcing the x direction.  Thanks for the fix!

PR #767, opened 2024-11-07
Added Shell.loft

Jojain:
I also wanted to add the shell loft into the shell operations but that is rather involved to make it work in `BuildPart`
I think for it to work we would need to be able to have a list of available `Wire` in the `BuildPart` context, which currently arent stored.

It wouldn't be has difficult for the algebra api part.

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -415,9 +415,7 @@ class Mixin1D:
         curve = self._geom_adaptor()
 
         length = GCPnts_AbscissaPoint.Length_s(curve)
-        return GCPnts_AbscissaPoint(
-            curve, length * distance, curve.FirstParameter()
-        ).Parameter()
+        return GCPnts_AbscissaPoint(curve, length * distance, curve.FirstParameter()).Parameter()
 
     def tangent_at(
         self,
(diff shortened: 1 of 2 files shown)


Jojain:
When running black latest it changes existing formatting is the version to use pinned somewhere? Or do I need to do something special to get the correct formatting?

Jojain:
Ok so I've found what was the issue, my black was setup to line per char = 100, the default is 88.
I would still like that you provide somewhere a config to run black with so I don't have to modify my settings to contribute.

jdegenstein:
@Jojain the defaults to use black ARE setup in pyproject.toml, including the 88 char setting you mentioned

We haven't settled on a stable black version to use, so for now the newest is fine.

Jojain:
> @Jojain the defaults to use black ARE setup in pyproject.toml, including the 88 char setting you mentioned
> 
> We haven't settled on a stable black version to use, so for now the newest is fine.

You are right, I think I got confused, because I had a pyproject toml without it. But I think it was before I update everything. 
Anyway idk, sorry for the request then !

gumyr:
Thank you - loft is better than it ever was!

PR #770, opened 2024-11-08
Shell thicken & Shape.split with Shell

Jojain:
Added a shell thicken method and accept a shell as a tool to split a shape


gumyr:
Unfortunately black has reformatted everything and the diffs are huge. It looks like the line length is set too small. If you can't get black to work just don't use it.

Jojain:
I'm on it, trying to solve my way around black and the conflicts. Long time I haven't pushed anything I'm rusty lol.

PR #771, opened 2024-11-08
Added thicken and new split tools

Jojain:
I redid the PR on another branch, was simpler than to solve the other one

### DIFF
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -6989,10 +6989,6 @@ class Shell(Shape):
 
         Args:
             depth (float): Amount to thicken face(s), can be positive or negative.
-            normal_override (Vector, optional): The normal_override vector can be used to
-                indicate which way is 'up', potentially flipping the face normal direction
-                such that many faces with different normals all go in the same direction
-                (direction need only be +/- 90 degrees from the face normal). Defaults to None.
 
         Raises:
             RuntimeError: Opencascade internal failures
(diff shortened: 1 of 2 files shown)


src/build123d/topology.py
```diff
@@ -366,6 +366,7 @@
 
 Shapes = Literal["Vertex", "Edge", "Wire", "Face", "Shell", "Solid", "Compound"]
 
+TrimmingTool = Union[Plane,"Shell", "Face"]
```
jdegenstein: I think there is a space missing between `Plane, "Shell"` (just don't want this to randomly show up in a future unrelated PR).

jdegenstein:
Looks like there are a few formatting inconsistencies, I left one comment but found more in the added tests. Can you run try running black again on that file too? Hope it doesn't cause more issues. Also one test should be "planar" instead of plarnar.

src/build123d/topology.py
```diff
@@ -7000,6 +6982,26 @@ def make_loft(
...
         return cls(_make_loft(objs, False, ruled))
 
+    def thicken(self, depth: float) -> Solid:
+        """Thicken Shell
+
+        Create a solid from a shell by thickening along the normals.
+
+        Args:
+            depth (float): Amount to thicken face(s), can be positive or negative.
+            normal_override (Vector, optional): The normal_override vector can be used to
```
gumyr: The `Shell` version of `thicken` doesn't have the `normal_override` parameter so it needs to be removed from the docstring.

gumyr:
Thanks!

PR #782, opened 2024-11-13
Alignment rework

erooke:
While using `build123d` I noticed that the type signature for `BoundBox.to_align_offset` was wrong. While fixing it I noticed that there was a lot of replicated aligning logic. This is an attempt to rectify that.

- add a generic `to_align_offset` function to the `geometry` module to allow for aligning
- add an `Align.NONE` enum member
- change `to_align_offset` to return a `Vector` as every call site immediately constructed a vector anyway

Not sure if there is interest in merging such a pr. If so I can go ahead and make sure the documentation lines up with the changes and make any changes y'all deem appropriate.

Would close:
- #457

Supersedes:
- #734

### DIFF
--- a/src/build123d/build_enums.py
+++ b/src/build123d/build_enums.py
@@ -36,6 +36,7 @@ class Align(Enum):
     MIN = auto()
     CENTER = auto()
     MAX = auto()
+    NONE = None
 
     def __repr__(self):
         return f"<{self.__class__.__name__}.{self.name}>"
--- a/src/build123d/build_enums.py
+++ b/src/build123d/build_enums.py
@@ -27,8 +27,11 @@ license:
 """
 
 from __future__ import annotations
+
 from enum import Enum, auto
-from typing import Union, TypeAlias
+from typing import Union
+
+from typing_extensions import TypeAlias
 
 
 class Align(Enum):
--- a/src/build123d/build_enums.py
+++ b/src/build123d/build_enums.py
@@ -28,6 +28,7 @@ license:
 
 from __future__ import annotations
 from enum import Enum, auto
+from typing import Union, TypeAlias
 
 
 class Align(Enum):
@@ -40,6 +41,17 @@ class Align(Enum):
 
     def __repr__(self):
         return f"<{self.__class__.__name__}.{self.name}>"
+
+
+Align2DType: TypeAlias = Union[
+    Union[Align, None],
+    tuple[Union[Align, None], Union[Align, None]],
+]
+
+Align3DType: TypeAlias = Union[
+    Union[Align, None],
+    tuple[Union[Align, None], Union[Align, None], Union[Align, None]],
+]
 
 
 class ApproxOption(Enum):
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -2530,15 +2530,27 @@ def to_align_offset(
     min_point: Sequence[float],
     max_point: Sequence[float],
     align: Sequence[Align],
+    center: Optional[Sequence[float]] = None,
 ) -> Vector:
     """Amount to move object to achieve the desired alignment"""
     align_offset = []
 
-    for alignment, min_coord, max_coord in zip(align, min_point, max_point):
+    if center is None:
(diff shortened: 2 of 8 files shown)


src/build123d/geometry.py
```diff
@@ -2536,3 +2526,35 @@ def intersect(self, *args, **kwargs):
 
         elif shape is not None:
             return shape.intersect(self)
+
+
+def to_align_offset(
+    min_point: Sequence[float],
```
gumyr: I think this function would make more sense if the inputs were `VectorLike` then all of the math could be done with Vectors instead of looping over floats.

src/build123d/build_enums.py
```diff
@@ -36,6 +36,7 @@ class Align(Enum):
     MIN = auto()
     CENTER = auto()
     MAX = auto()
+    NONE = None
```
gumyr, line 43: I'd like to introduce the following two TypeVars here that can be used on all of the objects (not necessarily in the PR):
```py
Align2DType = TypeVar(
    "Align2DType", Union[Align, None], Tuple[Union[Align, None], Union[Align, None]]
)
Align3DType = TypeVar(
    "Align3DType",
    Union[Align, None],
    Tuple[Union[Align, None], Union[Align, None], Union[Align, None]],
)
```
This could be done later though.

src/build123d/geometry.py
```diff
@@ -2536,3 +2526,35 @@ def intersect(self, *args, **kwargs):
...
+    max_point: Sequence[float],
+    align: Sequence[Align],
```
gumyr: align could also be None so I think should be typed as `align: Sequence[Union[Align,None]]`.

gumyr (changes requested):
There should be some tests of `Align.NONE` (even though they might not increase coverage).

src/build123d/build_enums.py
erooke: I was under the impression that typevars were meant to signal [generics](https://docs.python.org/3/library/typing.html#generics). I have added these using [type aliases](https://docs.python.org/3/library/typing.html#type-aliases) as that feels more correct to me.

src/build123d/geometry.py
erooke: I'm not sure I follow how making them vectors helps us avoid the looping behavior in the general case. I have changed to `VectorLike` as the signature and the only place I could see to remove the looping behavior was in the trivial case where we have a single alignment variable.

src/build123d/geometry.py
erooke: That makes sense to me. I guess if we add the `Align2DType` and `Align3DType` it could be typed `Union[Align2DType, Align3DType]`

erooke:
Made the requested changes, waiting to make sure the design aspects are ironed out before adding tests.

src/build123d/build_enums.py
gumyr: Thanks for pointing this out, I think there are other examples of where `TypeAlias` should be used instead of `TypeVar`.  This might help with the documentation as well as it seems as though ReadtheDocs will show the `TypeAlias` and not the full definition which will makes the docs more readable.

erooke: Whoops, responded to a comment mid it being edited :sweat_smile:.

gumyr: I'm learning about `TypeAlias` :)

gumyr:
> Made the requested changes, waiting to make sure the design aspects are ironed out before adding tests.

I'm surprised there is no coverage for the existing Align tests - I would have guessed that there are tests that would have covered them already. Add a few tests for the Align options shouldn't be difficult though.

erooke:
Added some, potentially overkill, tests. Wasn't quite sure where to stick them so I just threw them in their own file for now.

gumyr:
This looks good to me.  Are you ready to merge?

erooke:
> This looks good to me. Are you ready to merge?

Yup, forgot it was still a draft.

gumyr:
Thank you!

PR #783, opened 2024-11-13
Housekeeping

erooke:
While getting to grips with the codebase I noticed some small fixes which could
be made. Instead of opening a PR for each I thought I would just open one PR
addressing all of them. If any changes aren't desired I can remove them from
the request or open a bunch of smaller PRs if that's preferable.

## Changes

- remove bare exceptions
- explicitly state `stacklevel=2` in warnings
- remove function calls from default arguments
- remove unused imports
- make consistent pattern for `not in` and `is not None`
- remove unreachable code

## Rational

Replaced all bare exceptions with `except Exception` most exceptions
derive from this[^1], the ones which don't are not intended to be caught.
These are things Like `SystemExit` and `KeyboardInterrupt`. Catching these
makes it hard to kill the python program.

The default behavior for `warnings.warn` is to only print out the location
where the warning was issued. This is usually not the most helpful behavior as
it does not tell the consumer of `build123d` what they did to trigger the
warning. By setting `stacklevel=2` the call site will be marked with the warning
which is almost surely what is desired[^2].

Having a function call as a default argument causes the function to be called
once (upon declaration) which leads to confusing behavior. Instead we pass in
`None` and call the default function in the function body. This also helps
prevent accidentally mutating the default argument.

[^1]: https://docs.python.org/3/library/exceptions.html#exception-hierarchy
[^2]: https://docs.python.org/3/library/warnings.html#available-functions

### DIFF
--- a/src/build123d/joints.py
+++ b/src/build123d/joints.py
@@ -29,7 +29,7 @@ license:
 from __future__ import annotations
 
 from math import inf
-from typing import Union, overload
+from typing import Optional, Union, overload
 
 from build123d.build_common import validate_inputs
 from build123d.build_enums import Align
@@ -75,8 +75,8 @@ class RigidJoint(Joint):
     def __init__(
         self,
         label: str,
-        to_part: Union[Solid, Compound] = None,
-        joint_location: Location = Location(),
+        to_part: Optional[Union[Solid, Compound]] = None,
+        joint_location: Union[Location, None] = None,
     ):
         context: BuildPart = BuildPart._get_context(self)
         validate_inputs(context, self)
@@ -85,6 +85,9 @@ class RigidJoint(Joint):
                 to_part = context
             else:
                 raise ValueError("Either specify to_part or place in BuildPart scope")
+
+        if joint_location is None:
+            joint_location = Location()
 
         self.relative_location = to_part.location.inverse() * joint_location
         to_part.joints[label] = self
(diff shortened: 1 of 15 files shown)


jdegenstein:
Thanks for this PR. At first glance many of these changes look good, although there are some missing lines for test coverage (PRs should typically have 100% test coverage).

> - remove bare exceptions

These look generally good to me
> - explicitly state stacklevel=2 in warnings

I also think this change is great
> - remove function calls from default arguments

This makes sense too
> - remove unused imports

I will wait for @gumyr to comment on this, as there are a few cases where we might not want to eliminate the imports.
> - make consistent pattern for not in and is not None

This looks good to me.
> - remove unreachable code

Did not review this one specifically yet. Will wait for @gumyr on this one too.

erooke:
> PRs should typically have 100% test coverage

Sounds good, I'll work on fixing the test coverage.

src/build123d/joints.py
```diff
@@ -677,8 +680,8 @@ def symbol(self) -> Compound:
     def __init__(
         self,
         label: str,
-        to_part: Union[Solid, Compound] = None,
-        joint_location: Location = Location(),
+        to_part: Union[Solid, Compound, None] = None,
```
gumyr: Although I'm not a big fan of Optional as it seems redundant, I believe this is the correct way to do this:
```py
    to_part: Optional[Union[Solid, Compound]] = None,
    joint_location: Optional[Location] = None,
```

src/build123d/joints.py
```diff
@@ -75,8 +75,8 @@ def symbol(self) -> Compound:
...
```
gumyr: Optional?

gumyr (changes requested):
Thank you! There are lots of nice little improvements here.

erooke:
Addressed the code review and fixed the missing coverage on the drawing default date behavior. The rest of the missing coverage has to do with warnings which are raised due to OCP behavior (invalid meshes and the like). These paths had no coverage pre-pr and its going to take me a little longer to figure out how to test those particular branches.

gumyr:
Thanks for improving the code. These changes make things just a little better in many ways!

Issue #774, opened 2024-11-11
Support Path objects for import / export functions


lypwig:
Python's builtin [pathlib](https://docs.python.org/3/library/pathlib.html) normalizes path functions with an object-oriented Path object.

I could be very practical if all import and export functions could support Path objects for each argument involving a path.

I can make a PR for this if you agree.

KilowattSynthesis:
Duplicate of #685. Seems that there was no objection when it was brought up earlier; a PR would be fantastic, I'm sure :)

jdegenstein:
@roipoussiere I support this idea, PR would be very welcome

gumyr:
+1

PR #786, opened 2024-11-14
use pathlib to manage path names in import/export functions

lypwig:
see #774


gumyr:
I don't think this has ever happened before but two PR covering the same change arrived within 4 hours.  PR #787 is more comprehensive so I think it makes sense to continue with that one.

PR #787, opened 2024-11-14
Path support

### LINKED ISSUE #685: Feature Request: Add support for exports to `pathlib.Path`: `.export_stl(path_object)`
In the theme of being Pythonic, I want to be able to do this:

```python
from pathlib import Path

with BuildPart() as box_builder:
    Box(1, 1, 1)

box_builder.part.export_stl(Path(__file__).with_name("output_box.stl"))
```

Currently, this example code simply does not support calling `.export_stl()` with a `Path`-type arg.

### LINKED ISSUE #774: Support Path objects for import / export functions

erooke:
For all import / export methods make the path a `Union[Pathlike, str, bytes]` argument to follow python conventions as outlined in [pep 519](https://peps.python.org/pep-0519/).

closes: https://github.com/gumyr/build123d/issues/685
closes: https://github.com/gumyr/build123d/issues/774

~~I got carried away big time and this PR is probably unreasonably large. To test the new functionality I changed all the tests to feed in `pathlib.Paths`. In addition any tests which wrote to disk were rewritten to use the pytest `tmp_path` fixture. Which then caused me to rewrite the tests in the pytest style...~~

~~I can revert the test changes and be less invasive with my changes if desired.~~

Edit: I decided to roll those back. I think they're good changes in the long run but they get in the way of reviewing the pr at hand.

Open questions:
- should the deprecated export methods in `topology.py` also be moved over to this scheme? Technically that is what #685 is requesting
- ~~should the tests attempt to verify that all three types: string, paths, and bytes work?~~ I added said tests 
- did I miss any exporting/importing methods

### DIFF
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -649,7 +649,7 @@ class ExportDXF(Export2D):
         # https://github.com/gumyr/build123d/issues/382 tracks
         # exposing viewport control to the user.
         zoom.extents(self._modelspace)
-        self._document.saveas(file_name)
+        self._document.saveas(fsdecode(file_name))
 
     # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -32,9 +32,11 @@ license:
 
 import math
 import xml.etree.ElementTree as ET
-from enum import Enum, auto
-from typing import Callable, Iterable, Optional, Union, List, Tuple
 from copy import copy
+from enum import Enum, auto
+from os import PathLike, fsdecode, fspath
+from pathlib import Path
+from typing import Callable, Iterable, List, Optional, Tuple, Union
 
 import ezdxf
 import svgpathtools as PT
@@ -633,13 +635,13 @@ class ExportDXF(Export2D):
 
     # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
-    def write(self, file_name: str):
+    def write(self, file_name: Union[PathLike, str, bytes]):
         """write
 
         Writes the DXF data to the specified file name.
 
         Args:
-            file_name (str): The file name (including path) where the DXF data will
+            file_name (Union[PathLike, str, bytes]): The file name (including path) where the DXF data will
                 be written.
         """
         # Reset the main CAD viewport of the model space to the
@@ -647,7 +649,6 @@ class ExportDXF(Export2D):
         # https://github.com/gumyr/build123d/issues/382 tracks
         # exposing viewport control to the user.
         zoom.extents(self._modelspace)
-
         self._document.saveas(file_name)
 
     # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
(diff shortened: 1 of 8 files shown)


src/build123d/exporters.py
```diff
@@ -1428,13 +1428,13 @@ def _color_attribs(c: Color) -> Tuple[str, str]:
...
 
-    def write(self, path: str):
+    def write(self, path: Union[PathLike, str, bytes]):
         """write
 
         Writes the SVG data to the specified file path.
 
         Args:
-            path (str): The file path where the SVG data will be written.
+            path (Union[str, Path]): The file path where the SVG data will be written.
```
gumyr: `bytes` is missing here

src/build123d/exporters3d.py
```diff
@@ -153,25 +153,27 @@ def _create_xde(to_export: Shape, unit: Unit = Unit.MM) -> TDocStd_Document:
     return doc
 
 
-def export_brep(to_export: Shape, file_path: Union[str, BytesIO]) -> bool:
+def export_brep(
```
gumyr, line 157: `BytesIO` provides a mechanism for in memory data exchange which seems to be dropped here, or at least not visible.

src/build123d/exporters3d.py
```diff
@@ -349,9 +351,8 @@ def export_stl(
 
     writer = StlAPI_Writer()
 
-    if ascii_format:
-        writer.ASCIIMode = True
-    else:
-        writer.ASCIIMode = False
+    writer.ASCIIMode = ascii_format
```
gumyr, line 358: Yikes .. thanks for fixing this!

gumyr (changes requested):
For my education, what does the input type `bytes` actually do here? Why would a user pass `bytes` - is it for unicode file names?

gumyr:
There is no need to refactor the deprecated export methods.

src/build123d/exporters.py
erooke: good catch

src/build123d/exporters3d.py
erooke: Pushed a fix and added tests to make sure the next guy doesn't accidentally clobber that behavior too :sweat_smile:

erooke:
> For my education, what does the input type `bytes` actually do here? Why would a user pass `bytes` - is it for unicode file names?

I am not 100% sure. It seems to be a leftover from the fact that the c standard library used raw byte strings to represent paths. There is some discussion in pep 519 [here](https://peps.python.org/pep-0519/#have-fspath-only-return-strings). I included it in the signature because all of the underlying `Pathlike` functions support it and they claim:

> In the end, it was decided that using bytes to represent paths is simply not going to go away and thus they should be supported to some degree. The hope is that people will gravitate towards path objects like pathlib and that will move people away from operating directly with bytes.

gumyr:
Thank you!  Another nice improvement.

DeflateAwning:
Thank you! I've been looking forward to this very small quality-of-life improvement :)

PR #790, opened 2024-11-17
Slots error if width <= height

### LINKED ISSUE #73: Slot should raise an exception if width <= height
BuildSketch slot classes can cause errors that are difficult to debug when creating an invalid slots

### LINKED ISSUE #415: SlotOverall invalid when height > width
```py
with BuildSketch() as s:
    SlotOverall(width=0, height=2)
```

gives

Screenshot 2023-12-02 at 8 24 52 PM

erooke:
It is invalid to try and construct a slot whose width <= height. This PR adds checks for this in `SlotOverall`, `SlotCenterToCenter`, and `SlotCenterPoint`. `SlotArc` is omitted for now as I believe the only error condition would be an empty arc. 

Closes: #73 
Closes: #415

### DIFF
--- a/tests/test_pack.py
+++ b/tests/test_pack.py
@@ -8,10 +8,8 @@ date: November 9th 2023
 desc: Unit tests for the build123d pack module
 """
 
-import operator
 import random
 import unittest
-from functools import reduce
 
 from build123d import *
 
@@ -53,15 +51,14 @@ class TestPack(unittest.TestCase):
         random.seed(123456)
         # 50 is an arbitrary number that is large enough to exercise
         # different aspects of the packer while still completing quickly.
-        inputs = [
-            SlotOverall(random.randint(1, 20), random.randint(1, 20)) for _ in range(50)
-        ]
+        widths = [random.randint(2, 20) for _ in range(50)]
+        heights = [random.randint(1, width - 1) for width in widths]
+        inputs = [SlotOverall(width, height) for width, height in zip(widths, heights)]
         # Not raising in this call shows successfull non-overlap.
         packed = pack(inputs, 1)
-        self.assertEqual(
-            "bbox: 0.0 <= x <= 124.0, 0.0 <= y <= 105.0, 0.0 <= z <= 0.0",
-            str((Sketch() + packed).bounding_box()),
-        )
+        bb = (Sketch() + packed).bounding_box()
+        self.assertEqual(bb.min, Vector(0, 0, 0))
+        self.assertEqual(bb.max, Vector(70, 63, 0))
 
 
 if __name__ == "__main__":
--- a/src/build123d/objects_sketch.py
+++ b/src/build123d/objects_sketch.py
@@ -425,6 +425,13 @@ class SlotCenterPoint(BaseSketchObject):
         self.slot_height = height
 
         half_line = point_v - center_v
+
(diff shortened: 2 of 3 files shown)


erooke:
Curious if there are any thoughts on the `test_random_slots` test case. Right now it seems like the logic is:

1. Make sure no overlap occurs
2. Make sure the packing is the same as it was previously

As the slots are constructed randomly I'm not sure the packing which we are enforcing is meaningful. That is if there was any change to the packing algorithm we use this test would break despite us not breaking the contract.

I would think reducing the test to just make sure there is no overlap would suffice. So something like:

```python
def test_random_slots():
    """Test pack for 2D objects."""
    random.seed(123456)
    # 50 is an arbitrary number that is large enough to exercise
    # different aspects of the packer while still completing quickly.
    widths = [random.randint(2, 20) for _ in range(50)]
    heights = [random.randint(1, width - 1) for width in widths]
    inputs = [SlotOverall(width, height) for width, height in zip(widths, heights)]
    # Not raising in this call shows successfull non-overlap.
    packed = pack(inputs, 1)
```

tests/test_pack.py
```diff
@@ -48,20 +46,20 @@ def test_random_boxes(self):
...
+    random.seed(123456)
+    # 50 is an arbitrary number that is large enough to exercise
+    # different aspects of the packer while still completing quickly.
+    widths = [random.randint(2, 20) for _ in range(50)]
+    heights = [random.randint(1, width - 1) for width in widths]
+    inputs = [SlotOverall(width, height) for width, height in zip(widths, heights)]
+    # Not raising in this call shows successfull non-overlap.
+    packed = pack(inputs, 1)
+    bb = (Sketch() + packed).bounding_box()
+    assert bb.min == Vector(0, 0, 0)
```
gumyr: These aren't valid unittest assertions. However, testing the min/max of the bounding box is better than a string compare.

gumyr (changes requested):

tests/test_pack.py
erooke: At least in ci y'all are using pytest to run your tests. Pytest will do the [right thing](https://docs.pytest.org/en/stable/how-to/assert.html#assert) with these assert statements.

gumyr: However, the test file itself only imports unittest and if run stand alone does:
```py
if __name__ == "__main__":
    unittest.main()
```
Let's make this file consistent at least with a unittest assertion.

erooke: > Let's make this file consistent at least with a unittest assertion.

Sure, I have changed that test back to the `unittest` style. 

In general are we aiming to be compatible with `unittest`? There are places in the test suite where we use these bare asserts and other places which we rely on `pytest` explicitly. Given this I was migrating tests over to `pytest` conventions as `pytest` opens up some useful functionality down the road (I also find it more readable but that may just be a personal bias).

erooke:
Not sure what's causing that new test failure... looks like an issue with floating point math on a random test.

tests/test_pack.py
gumyr: I started off exclusively using unittest but over time others have used pytest. When one receives generous donations in the form of tested code, best not to get into wars about what is "better".

gumyr:
Thanks again.

PR #794, opened 2024-11-19
Drop support for python 3.9

jdegenstein:
This PR drops support for python 3.9 in all places I am aware of (CI / pyproject.toml). Please merge when ready.

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -8,7 +8,8 @@ jobs:
       matrix:
         python-version: [
           "3.10",
-          "3.11"
+          # "3.11",
+          "3.12"
           ]
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


gumyr:
Should we be testing against 3.10 and 3.12 instead of 3.10 and 3.11?

jdegenstein:
> Should we be testing against 3.10 and 3.12 instead of 3.10 and 3.11?

Sure, switched to 3.10 and 3.12 in the latest commits. Everything looks good on tests.

gumyr:
No one has raised any concerns about dropping support for 3.9 so I vote for merging this PR whenever you are ready.  Thanks for doing this!  I'm looking forward to using 3.10 as there are some nice improvements from what I recall.

jdegenstein:
Merged.

PR #812, opened 2024-12-09
Refactored topology.py ready to split into multiple modules

gumyr:
Here is the refactor of the `topology.py` module that I've been working on.  The change will occur in two stages:
1) All of these changes are applied against the existing file structure and should pass all of the acceptance criteria as is.
2) The `refactor_topo.py` tool (look in the `build123d/tools/` directory) converts the file `topology_old.py` (just `topology.py` renamed) into the new structure including generation of the` __init__.py` file. Reviewers, it will be useful to run the tool to see what the resulting files look like.  Actually introducing the split files needs to be done carefully to avoid loosing history for the files.

The new structure is (lines file order main_classes):
```
     92 __init__.py
   2831 shape_core.py  Shape, ShapeList
    470 utils.py
    319 zero_d.py      Vertex
   2921 one_d.py       Mixin1D, Edge, Wire
   1265 two_d.py       Mixin2D, Face, Shell
   1337 three_d.py     Mixin3D, Solid
    730 composite.py   Compound (and soon to be Assembly)
  10173 total
```
The core idea is that there is a topology order of these objects - no object shall depend on a higher order module: `shape_core` being the base, `composite` being the highest. As `Edge` and `Wire` are in the same module they can depend on each other.

 Not being able to use a `Compound` in the lower order object results changes to the API - here are all of the API changes:
- boolean operations may generate a list: `Part() + ...` will generate a `Compound` if necessary, while `Solid() + ...` may generate a `ShapeList` of `Solids`.
- `surface.thicken(amount)` -> `Solid.thicken(surface, amount)`: this was an exception to the other class methods and was problematic because a `Face` was generating an object of a higher topological order which isn't possible anymore.  This is a key thing to look for in the future.
- deprecated methods were removed and the vtk ones commented out to see if anyone notices
- `shape.split(Keep.BOTH)` generates a `tuple` of `Shape` instead of a `Compound`
- `Shape.extrude()` is replaced by specific class versions like `Edge.extrude()`, `Face.extrude()`, etc.

Quite a bit of code changed - 14 files changed, 1865 insertions(+), 1777 deletions(-) - as many of the methods needed to rewritten to directly work at the OCCT level and not depend on resources provided by higher order classes.

This change will help with long term support of the project.  The `topology.py` file is almost 10,000 lines long which is kinda crazy and difficult to work with.

Any feedback you can provide on this change is greatly appreciated.

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -429,23 +429,30 @@ class Builder(ABC):
                 if mode == Mode.ADD:
                     if self._obj is None:
                         if len(typed[self._shape]) == 1:
-                            self._obj = typed[self._shape][0]
+                            combined = typed[self._shape][0]
                         else:
-                            self._obj = (
+                            combined = (
                                 typed[self._shape].pop().fuse(*typed[self._shape])
                             )
                     else:
-                        self._obj = self._obj.fuse(*typed[self._shape])
+                        combined = self._obj.fuse(*typed[self._shape])
                 elif mode == Mode.SUBTRACT:
                     if self._obj is None:
                         raise RuntimeError("Nothing to subtract from")
-                    self._obj = self._obj.cut(*typed[self._shape])
+                    combined = self._obj.cut(*typed[self._shape])
                 elif mode == Mode.INTERSECT:
                     if self._obj is None:
                         raise RuntimeError("Nothing to intersect with")
-                    self._obj = self._obj.intersect(*typed[self._shape])
+                    combined = self._obj.intersect(*typed[self._shape])
                 elif mode == Mode.REPLACE:
-                    self._obj = Compound(list(typed[self._shape]))
+                    combined = self._sub_class(list(typed[self._shape]))
+
+                # If the boolean operation created a list, convert back
+                self._obj = (
+                    self._sub_class(combined)
+                    if isinstance(combined, list)
+                    else combined
+                )
 
                 if self._obj is not None and clean:
                     self._obj = self._obj.clean()
--- a/src/build123d/drafting.py
+++ b/src/build123d/drafting.py
@@ -439,23 +439,25 @@ class DimensionLine(BaseSketchObject):
         overage = shaft_length + draft.pad_around_text + label_length / 2
         label_u_values = [0.5, -overage / path_length, 1 + overage / path_length]
 
-        # d_lines = Sketch(children=arrows[0])
         d_lines = {}
-        # for arrow_pair in arrow_shapes:
         for u_value in label_u_values:
-            d_line = Sketch()
-            for add_arrow, arrow_shape in zip(arrows, arrow_shapes):
-                if add_arrow:
-                    d_line += arrow_shape
+            select_arrow_shapes = [
+                arrow_shape
+                for add_arrow, arrow_shape in zip(arrows, arrow_shapes)
+                if add_arrow
+            ]
+            d_line = Sketch(select_arrow_shapes)
             flip_label = path_obj.tangent_at(u_value).get_angle(Vector(1, 0, 0)) >= 180
             loc = Draft._sketch_location(path_obj, u_value, flip_label)
             placed_label = label_shape.located(loc)
-            self_intersection = d_line.intersect(placed_label).area
+            self_intersection = Sketch.intersect(d_line, placed_label).area
             d_line += placed_label
             bbox_size = d_line.bounding_box().size
 
             # Minimize size while avoiding intersections
-            common_area = 0.0 if sketch is None else d_line.intersect(sketch).area
+            common_area = (
+                0.0 if sketch is None else Sketch.intersect(d_line, sketch).area
+            )
             common_area += self_intersection
             score = (d_line.area - 10 * common_area) / bbox_size.X
             d_lines[d_line] = score
--- a/src/build123d/drafting.py
+++ b/src/build123d/drafting.py
@@ -50,7 +50,7 @@ from build123d.objects_curve import Line, TangentArc
 from build123d.objects_sketch import BaseSketchObject, Polygon, Text
 from build123d.operations_generic import fillet, mirror, sweep
 from build123d.operations_sketch import make_face, trace
-from build123d.topology import Compound, Edge, Sketch, Vertex, Wire
+from build123d.topology import Compound, Curve, Edge, Sketch, Vertex, Wire
 
 
 class ArrowHead(BaseSketchObject):
@@ -704,7 +704,10 @@ class TechnicalDrawing(BaseSketchObject):
         )
         bf_pnt3 = box_frame_curve.edges().sort_by(Axis.X)[0] @ (1 / 3)
         bf_pnt4 = box_frame_curve.edges().sort_by(Axis.X)[0] @ (2 / 3)
-        box_frame_curve += Edge.make_line(bf_pnt3, (bf_pnt2.X, bf_pnt3.Y))
+        box_frame_curve = Curve() + [
+            box_frame_curve,
+            Edge.make_line(bf_pnt3, (bf_pnt2.X, bf_pnt3.Y)),
+        ]
         box_frame_curve += Edge.make_line(bf_pnt4, (bf_pnt2.X, bf_pnt4.Y))
         bf_pnt5 = box_frame_curve.edges().sort_by(Axis.Y)[-1] @ (1 / 3)
         bf_pnt6 = box_frame_curve.edges().sort_by(Axis.Y)[-1] @ (2 / 3)
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -450,7 +450,7 @@ class Vector:
 
     def __hash__(self) -> int:
         """Hash of Vector"""
-        return hash(self.X) + hash(self.Y) + hash(self.Z)
+        return hash(round(self.X, 6)) + hash(round(self.Y, 6)) + hash(round(self.Z, 6))
 
     def __copy__(self) -> Vector:
         """Return copy of self"""
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -861,8 +861,12 @@ class BoundBox:
     """A BoundingBox for a Shape"""
 
     def __init__(self, bounding_box: Bnd_Box) -> None:
-        self.wrapped: Bnd_Box = bounding_box
-        x_min, y_min, z_min, x_max, y_max, z_max = bounding_box.Get()
+        if bounding_box.IsVoid():
+            self.wrapped = None
+            x_min, y_min, z_min, x_max, y_max, z_max = (0,) * 6
+        else:
+            self.wrapped: Bnd_Box = bounding_box
+            x_min, y_min, z_min, x_max, y_max, z_max = bounding_box.Get()
         self.min = Vector(x_min, y_min, z_min)  #: location of minimum corner
         self.max = Vector(x_max, y_max, z_max)  #: location of maximum corner
         self.size = Vector(x_max - x_min, y_max - y_min, z_max - z_min)  #: overall size
@@ -870,6 +874,8 @@ class BoundBox:
     @property
     def diagonal(self) -> float:
         """body diagonal length (i.e. object maximum size)"""
+        if self.wrapped is None:
+            return 0.0
         return self.wrapped.SquareExtent() ** 0.5
 
     def __repr__(self):
@@ -914,13 +920,14 @@ class BoundBox:
 
         tmp = Bnd_Box()
         tmp.SetGap(tol)
-        tmp.Add(self.wrapped)
+        if self.wrapped is not None:
+            tmp.Add(self.wrapped)
 
         if isinstance(obj, tuple):
             tmp.Update(*obj)
         elif isinstance(obj, Vector):
             tmp.Update(*obj.to_tuple())
-        elif isinstance(obj, BoundBox):
+        elif isinstance(obj, BoundBox) and obj.wrapped is not None:
             tmp.Add(obj.wrapped)
 
         return BoundBox(tmp)
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -963,7 +963,7 @@ class BoundBox:
         return result
 
     @classmethod
-    def _from_topo_ds(
+    def from_topo_ds(
         cls,
         shape: TopoDS_Shape,
         tolerance: float = None,
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -450,7 +450,7 @@ class Vector:
 
     def __hash__(self) -> int:
         """Hash of Vector"""
-        return hash(round(self.X, 6)) + hash(round(self.Y, 6)) + hash(round(self.Z, 6))
+        return hash((round(self.X, 6), round(self.Y, 6), round(self.Z, 6)))
 
     def __copy__(self) -> Vector:
         """Return copy of self"""
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -542,7 +542,7 @@ class Vector:
 
 #:TypeVar("VectorLike"): Tuple of float or Vector defining a position in space
 VectorLike = Union[
-    Vector, tuple[float, float], tuple[float, float, float], Iterable[float]
+    Vector, tuple[float, float], tuple[float, float, float], Sequence[float]
 ]
 
 
--- a/src/build123d/operations_generic.py
+++ b/src/build123d/operations_generic.py
@@ -904,7 +904,7 @@ SplitType = Union[Edge, Wire, Face, Solid]
 
 def split(
     objects: Union[SplitType, Iterable[SplitType]] = None,
-    bisect_by: Union[Plane, Face] = Plane.XZ,
+    bisect_by: Union[Plane, Face, Shell] = Plane.XZ,
     keep: Keep = Keep.TOP,
     mode: Mode = Mode.REPLACE,
 ):
@@ -937,7 +937,16 @@ def split(
 
     new_objects = []
     for obj in object_list:
-        new_objects.append(obj.split(bisect_by, keep))
+        bottom = None
+        if keep == Keep.BOTH:
+            top, bottom = obj.split(bisect_by, keep)
+        else:
+            top = obj.split(bisect_by, keep)
+        for subpart in [top, bottom]:
+            if isinstance(subpart, Iterable):
+                new_objects.extend(subpart)
+            elif subpart is not None:
+                new_objects.append(subpart)
 
     if context is not None:
         context._add_to_context(*new_objects, mode=mode)
--- a/src/build123d/operations_sketch.py
+++ b/src/build123d/operations_sketch.py
@@ -305,4 +305,9 @@ def trace(
         context.pending_edges = ShapeList()
 
     combined_faces = Face.fuse(*new_faces) if len(new_faces) > 1 else new_faces[0]
-    return Sketch(combined_faces.wrapped)
+    result = (
+        Sketch(combined_faces)
+        if isinstance(combined_faces, list)
+        else Sketch(combined_faces.wrapped)
+    )
+    return result
--- a/src/build123d/operations_sketch.py
+++ b/src/build123d/operations_sketch.py
@@ -40,9 +40,8 @@ from build123d.topology import (
     Sketch,
     topo_explore_connected_edges,
     topo_explore_common_vertex,
-    TOLERANCE,
 )
-from build123d.geometry import Vector
+from build123d.geometry import Vector, TOLERANCE
 from build123d.build_common import flatten_sequence, validate_inputs
 from build123d.build_sketch import BuildSketch
 from scipy.spatial import Voronoi
--- a/src/build123d/topology.py
+++ b/src/build123d/topology.py
@@ -36,10 +36,12 @@ from __future__ import annotations
 #   too-many-arguments, too-many-locals, too-many-public-methods,
 #   too-many-statements, too-many-instance-attributes, too-many-branches
 import copy
+import inspect
 import itertools
 import os
 import platform
 import sys
+import types
 import warnings
 from abc import ABC, ABCMeta, abstractmethod
 from io import BytesIO
@@ -54,20 +56,22 @@ from typing import (
     Iterator,
     Optional,
     Protocol,
+    Sequence,
     Tuple,
     Type,
     TypeVar,
     Union,
     overload,
+    TYPE_CHECKING,
 )
 from typing import cast as tcast
 from typing_extensions import Self, Literal, deprecated
 
 from anytree import NodeMixin, PreOrderIter, RenderTree
-from IPython.lib.pretty import pretty
+from IPython.lib.pretty import pretty, PrettyPrinter
 from numpy import ndarray
 from scipy.optimize import minimize
-from scipy.spatial import ConvexHull
+from scipy.spatial import ConvexHull  # pylint:disable=no-name-in-module
 from vtkmodules.vtkCommonDataModel import vtkPolyData
 from vtkmodules.vtkFiltersCore import vtkPolyDataNormals, vtkTriangleFilter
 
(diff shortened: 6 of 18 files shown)


gumyr:
Would someone with a non-Intel Mac be able to help me find the problem with the tests failing (I don't have one)? 

```py
e1 = Edge.make_line((0, -1), (2, 1))
e2 = Edge.make_line((0, 1), (2, -1))
e3 = Edge.make_line((0, 0), (2, 0))

i = e1.intersect(e2, e3)
```
produces a `Vertex` on most systems but there is a problem on a Mac.  The problem is almost certainly in `topology.Edge.find_intersection_points` - possibly the `distance_to(pnt) <= TOLERANCE` checks are failing?

src/build123d/geometry.py
```diff
@@ -450,7 +450,7 @@ def __eq__(self, other: object) -> bool:
 
     def __hash__(self) -> int:
         """Hash of Vector"""
-        return hash(self.X) + hash(self.Y) + hash(self.Z)
+        return hash(round(self.X, 6)) + hash(round(self.Y, 6)) + hash(round(self.Z, 6))
```
```suggestion
        return hash((round(self.X, 6), round(self.Y, 6), round(self.Z, 6)))
```
~~any reason not to use the built-in tuple hash instead of adding individual number hashes?~~
actually summing the 3 hashes doesn't work correctly in the first place (because `hash(123) == 123` )

gumyr: I've made this change.  Thanks for pointing out the algorithm was flawed.

src/build123d/build_common.py
```diff
@@ -429,23 +429,30 @@ def _add_to_context(
                 if mode == Mode.ADD:
                     if self._obj is None:
                         if len(typed[self._shape]) == 1:
-                            self._obj = typed[self._shape][0]
+                            combined = typed[self._shape][0]
                         else:
-                            self._obj = (
+                            combined = (
                                 typed[self._shape].pop().fuse(*typed[self._shape])
                             )
```
snoyer, line 436:
```py
                        first, *nexts = typed[self._shape]
                        combined = first.fuse(*nexts) if nexts else first
```

src/build123d/drafting.py
```diff
@@ -439,23 +439,25 @@ def __init__(
...
-            d_line = Sketch()
-            for add_arrow, arrow_shape in zip(arrows, arrow_shapes):
-                if add_arrow:
-                    d_line += arrow_shape
+            select_arrow_shapes = [
+                arrow_shape
+                for add_arrow, arrow_shape in zip(arrows, arrow_shapes)
+                if add_arrow
+            ]
+            d_line = Sketch(select_arrow_shapes)
```
snoyer, line 449:
```suggestion
            d_line = Sketch(
                arrow_shape
                for add_arrow, arrow_shape in zip(arrows, arrow_shapes)
                if add_arrow
            )
```

src/build123d/geometry.py
```diff
@@ -450,7 +450,7 @@ def __eq__(self, other: object) -> bool:
 
     def __hash__(self) -> int:
         """Hash of Vector"""
-        return hash(self.X) + hash(self.Y) + hash(self.Z)
+        return hash((round(self.X, 6), round(self.Y, 6), round(self.Z, 6)))
```
snoyer, line 453: ideally `__hash__` should match `__eq__` I believe

src/build123d/operations_generic.py
```diff
@@ -937,7 +937,16 @@ def split(
 
     new_objects = []
     for obj in object_list:
-        new_objects.append(obj.split(bisect_by, keep))
+        bottom = None
+        if keep == Keep.BOTH:
+            top, bottom = obj.split(bisect_by, keep)
+        else:
+            top = obj.split(bisect_by, keep)
```
snoyer, line 944:
```suggestion
        if keep == Keep.BOTH:
            top, bottom = obj.split(bisect_by, keep)
        else:
            top, bottom = obj.split(bisect_by, keep), None
```
assign only once

src/build123d/operations_sketch.py
```diff
@@ -306,4 +305,9 @@ def trace(
         context.pending_edges = ShapeList()
 
     combined_faces = Face.fuse(*new_faces) if len(new_faces) > 1 else new_faces[0]
```
snoyer, line 307:
```suggestion
    combined_faces = Face.fuse(*new_faces)
```
the `if len(to_fuse) < 2: return ShapeList(to_fuse)` special case test should go inside the `fuse()` method

src/build123d/topology.py
```diff
@@ -379,26 +361,70 @@ def tuplify(obj: Any, dim: int) -> tuple:
...
     """Methods to add to the Edge and Wire classes"""
 
     def __add__(self, other: Union[list[Shape], Shape]) -> Self:
         """fuse shape to wire/edge operator +"""
 
-        # Convert `other` to list of base objects and filter out None values
+        # Convert `other` to list of base topods objects and filter out None values
         summands = [
             shape
             for o in (other if isinstance(other, (list, tuple)) else [other])
```
snoyer:
```suggestion
            for o in ([other] if isinstance(other, (Shape)) else other)
```
reverse test to allow other to be `Iterable[Shape]`

src/build123d/topology.py
```diff
@@ -379,26 +361,70 @@ def tuplify(obj: Any, dim: int) -> tuple:
...
         if not summands:
             return self
 
-        if not all(summand._dim == 1 for summand in summands):
+        if not all(summand.dim == 1 for summand in summands):
             raise ValueError("Only shapes with the same dimension can be added")
 
+        # Convert back to Edge/Wire objects now that it's safe to do so
+        summands = [Mixin1D.cast(s) for s in summands]
         summand_edges = [e for summand in summands for e in summand.edges()]
```
snoyer:
```suggestion
        summand_edges = chain.from_iterable(s.edges() for s in summands)
```

src/build123d/topology.py
```diff
@@ -408,26 +434,180 @@ def __add__(self, other: Union[list[Shape], Shape]) -> Self:
                     sum_shape = Wire(summand_edges)
                 except Exception:
                     sum_shape = summands[0].fuse(*summands[1:])
+                    if type(self).order == 4:
+                        sum_shape = type(self)(sum_shape)
         else:
             try:
                 sum_shape = Wire(self.edges() + summand_edges)
```
snoyer:
```suggestion
                sum_shape = Wire(chain(self.edges() , summand_edges))
```

src/build123d/topology.py
```diff
@@ -1561,45 +2027,53 @@ def is_manifold(self) -> bool:
...
-            # shape is not manifold.
-            if shape_map.FindFromIndex(i + 1).Extent() != 2:
-                result = False
-                break
+                # Check if the current edge has exactly two faces associated with it. If not,
+                # it means the edge is not shared by exactly two faces, indicating that the
+                # shape is not manifold.
+                if shape_map.FindFromIndex(i + 1).Extent() != 2:
+                    result = False
+                    break
```
snoyer:
```suggestion
                    return False
```

src/build123d/topology.py
```diff
@@ -1561,45 +2027,53 @@ def is_manifold(self) -> bool:
...
+            results.append(result)
 
-        return result
+        return all(results)
```
snoyer: building a whole list only to call `all()` on it isn't necessary, either do an early return instead of `.append(False)` or change the list to a generator

src/build123d/topology.py
```diff
@@ -4266,32 +4392,38 @@ def _pre_attach(self, parent: Compound):
...
             logger.debug("Adding children %s to %s", kids, self.label)
-            self.wrapped = Compound._make_compound([c.wrapped for c in self.children])
+            self.wrapped = _make_topods_compound_from_shapes(
+                [c.wrapped for c in self.children]
+            )
         # else:
         #     logger.debug("Adding no children to %s", self.label)
 
-    def __add__(self, other: Union[list[Shape], Shape]) -> Shape:
+    def __add__(self, other: Shape | Sequence[Shape]) -> Compound:
```
snoyer:
```suggestion
    def __add__(self, other: Shape | Iterable[Shape]) -> Compound:
```

src/build123d/topology.py
```diff
@@ -4301,36 +4433,77 @@ def __add__(self, other: Union[list[Shape], Shape]) -> Shape:
             curve = Curve() if self.wrapped is None else Curve(self.wrapped)
             self.copy_attributes_to(curve, ["wrapped", "_NodeMixin__children"])
             return curve + other
+
+        summands = [
+            shape
+            for o in (other if isinstance(other, (list, tuple)) else [other])
```
snoyer:
```suggestion
            for o in ([other] if isinstance(other, Shape) else other)
```

src/build123d/topology.py
```diff
@@ -4301,36 +4433,77 @@ def __add__(self, other: Union[list[Shape], Shape]) -> Shape:
...
-            else:
-                fuse_op = BRepAlgoAPI_Fuse()
-                fuse_op.SetFuzzyValue(TOLERANCE)
-                self.copy_attributes_to(
-                    summands[0], ["wrapped", "_NodeMixin__children"]
-                )
-                result = self._bool_op(summands[:1], summands[1:], fuse_op)
+        return result
+
+    def __sub__(self, other: Shape | Sequence[Shape]) -> Compound:
```
snoyer: could accept `Iterable[Shape]`

src/build123d/topology.py
```diff
@@ -4301,36 +4433,77 @@ def __add__(self, other: Union[list[Shape], Shape]) -> Shape:
...
+        """Cut other to self `-` operator"""
+        difference = Shape.__sub__(self, other)
+        difference = Compound(
+            difference if isinstance(difference, list) else [difference]
+        )
+        self.copy_attributes_to(difference, ["wrapped", "_NodeMixin__children"])
+
+        return difference
+
+    def __and__(self, other: Shape | Sequence[Shape]) -> Compound:
```
snoyer: could accept `Iterable[Shape]`

src/build123d/topology.py
```diff
@@ -4876,49 +5018,39 @@ def _intersect_with_edge(self, edge: Edge) -> Shape:
...
-        # Find Edge/Edge overlaps
-        intersect_op = BRepAlgoAPI_Common()
-        edge_intersections = self._bool_op((self,), (Edge(axis),), intersect_op).edges()
-
-        return Compound(vertex_intersections + edge_intersections)
+        return vertex_intersections, edge_intersections
 
     def find_intersection_points(
-        self, edge: Union[Axis, Edge] = None, tolerance: float = TOLERANCE
+        self, other: Axis | Edge = None, tolerance: float = TOLERANCE
```
snoyer:
```suggestion
        self, other: Optional[Axis | Edge] = None, tolerance: float = TOLERANCE
```

src/build123d/topology.py
```diff
@@ -7037,16 +6815,15 @@ class Solid(Mixin3D, Shape):
...
 
     @property
     def _dim(self) -> int:
         return 3
 
-    @overload
     def __init__(
         self,
-        obj: TopoDS_Shape,
+        obj: TopoDS_Shape | Shell = None,
```
snoyer:
```suggestion
        obj: Optional[TopoDS_Shape | Shell] = None,
```

src/build123d/topology.py
```diff
@@ -8592,7 +8379,58 @@ def chamfer_2d(
...
+    def order_chamfer_edges(reference_edge, edges) -> tuple[Edge, Edge]:
+        """Order the edges of a chamfer relative to a reference Edge"""
+        if reference_edge:
+            if reference_edge not in edges:
+                raise ValueError("One or more vertices are not part of edge")
+            edge1 = reference_edge
+            edge2 = [x for x in edges if x != reference_edge][0]
+        else:
+            edge1, edge2 = edges
+        return edge1, edge2
```
snoyer: assuming `edges` is always a pair not an arbitrary list
```suggestion
    def order_chamfer_edges(reference_edge: Optional[Edge], edges: tuple[Edge, Edge]) -> tuple[Edge, Edge]:
        """Order the edges of a chamfer relative to a reference Edge"""
        if reference_edge:
            edge1, edge2 = edges
            if edge1 == reference_edge:
                return edge1, edge2
            elif edge2 == reference_edge:
                return edge2, edge1
            else:
                raise ValueError("reference edge not in edges")
        else:
            return edges
```

src/build123d/topology.py
```diff
@@ -8592,7 +8379,58 @@ def chamfer_2d(
...
+        for v in vertices:
+            edges = vertex_edge_map.FindFromKey(v.wrapped)
+
+            # Index or iterator access to OCP.TopTools.TopTools_ListOfShape is slow on M1 macs
+            # Using First() and Last() to omit
+            edges = [edges.First(), edges.Last()]
+
+            # Need to wrap in b3d objects for comparison to work
+            # ref.wrapped != edge.wrapped but ref == edge
+            edges = [Edge(e) for e in edges]
```
snoyer:
```suggestion
            # Index or iterator access to OCP.TopTools.TopTools_ListOfShape is slow on M1 macs
            # Using First() and Last() to omit
            # Need to wrap in b3d objects for comparison to work
            # ref.wrapped != edge.wrapped but ref == edge
            edges = Edge(edges.First()), Edge(edges.Last())
```
there's the same bit of code in `chamfer_2d` too

src/build123d/topology.py
```diff
@@ -9248,6 +9068,242 @@ def unwrap_topods_compound(
     return compound
 
 
+def get_top_level_topods_shapes(topods_shape: TopoDS_Shape) -> ShapeList[TopoDS_Shape]:
```
snoyer: the `ShapeList[T]` parameter `T` is bound to `Shape | Vector` so that `TopoDS_Shape` looks sketchy

src/build123d/topology.py
```diff
@@ -9248,6 +9068,242 @@ def unwrap_topods_compound(
...
+                    # Traverse further into the compound
+                    stack.append(child_shape)
+                else:
+                    # Add non-compound shape
+                    first_level_shapes.append(child_shape)
+                iterator.Next()
+        else:
+            first_level_shapes.append(current_shape)
+
+    return ShapeList(first_level_shapes)
```
snoyer: building the raw `list` only to convert to `ShapeList` looks unecessary. Either use a generator and do `return ShapeList(first_level_shapes())` or do `first_level_shapes = ShapeList[...]()` and just `return first_level_shapes`

snoyer (commented):
Did a pass that ended up being mostly syntax/language nitpicks.
Not sure I'm really competent for an overall architecture review.

src/build123d/topology.py
```diff
@@ -300,68 +304,46 @@
 )
 
 
+@property
+def _topods_compound_dim(self) -> int | None:
```
Jojain: These are utilities but there are not in the utils module. Either put the utils locally in the module they belong or put them all in utils but it shouldn't be mixed up.

src/build123d/topology.py
```diff
@@ -300,68 +304,46 @@
...
+    u_val, v_val = projector.LowerDistanceParameters()
+
+    gp_pnt = gp_Pnt()
+    normal = gp_Vec()
+    BRepGProp_Face(self).Normal(u_val, v_val, gp_pnt, normal)
+
+    return Vector(normal).normalized()
+
+
+def apply_ocp_monkey_patches() -> None:
```
Jojain: I think it would be better to use utility function rather than monkey patching OCP. What does it offers than a plain function do not?

src/build123d/topology.py
```diff
@@ -379,26 +361,70 @@ def tuplify(obj: Any, dim: int) -> tuple:
     return result
 
 
+class _ClassMethodProxy:
```
Jojain: I propose to remove this, its currently only use for `extrude` with the `extrude` func being a general func but tied to the `Mixin1d`. Why not make extrude an abstract method on `Shape` and have every topo class implementing it ? (It can be made a OCP util func and then each subclass would add only some typing info over it, keeping the code reuse while simplifying how it works)

src/build123d/topology.py
```diff
@@ -408,26 +434,180 @@ def __add__(self, other: Union[list[Shape], Shape]) -> Self:
...
+        # Extend the lookup table with additional entries
+        constructor_lut = {
+            ta.TopAbs_VERTEX: Vertex,
+            ta.TopAbs_EDGE: Edge,
+            ta.TopAbs_WIRE: Wire,
+        }
+
+        shape_type = shapetype(obj)
+        # NB downcast is needed to handle TopoDS_Shape types
+        return constructor_lut[shape_type](downcast(obj))
```
Jojain: Maybe you would want to do a `if shape_type not in constructor_lut : raise SomeException` rather than getting a KeyError ?

src/build123d/topology.py
```diff
@@ -997,24 +1173,240 @@ def project(
...
+        """
+        return cls.cast(_extrude_topods_shape(obj.wrapped, direction))
+
+
+class Mixin2D:
+    """Additional methods to add to Face and Shell class"""
+
+    project_to_viewport = Mixin1D.project_to_viewport
+    extrude = _ClassMethodProxy(Mixin1D.extrude)
+    split = Mixin1D.split
```
Jojain: Instead of doing this kind of binding redefine the methods to you can add proper type hints. During my testing this lost me all the typing. Furthermore it is not consistent, extrude has a proxy split doesn't. 
This is too complicated for saving some characters, its not worth it IMO.

src/build123d/topology.py
```diff
@@ -1517,6 +1974,15 @@ def color(self) -> Union[None, Color]:
         self._color = node_color  # Set the node's color for next time
         return node_color
 
+    @property
```
Jojain: This method shouldn't belong here, why one would want to call that on a shape that is not a face ?
There should be a proper cast to Face first before wanting to check if its planar or not.

src/build123d/topology.py
```diff
@@ -2218,127 +2583,90 @@ def get_top_level_shapes(self) -> ShapeList[Shape]:
...
-                first_level_shapes.append(current_shape)
-
-        return ShapeList(first_level_shapes)
+        return ShapeList(
+            self.__class__.cast(s) for s in get_top_level_topods_shapes(self.wrapped)
+        )
 
     @staticmethod
-    def _get_shape_list(shape: Shape, entity_type: str) -> ShapeList:
+    def get_shape_list(shape: Shape, entity_type: str) -> ShapeList:
```
Jojain: Several comments : 
1. Why is this a staticmethod ? It would make sense to be a method.
2. Instead of making it rely on type literal it could be a virtual method, that way each subclass could pass the real type instead of a string.

With some generics we could also type the return type of the ShapeList (and some constraints in the input entity_type)

src/build123d/topology.py
```diff
@@ -2218,127 +2583,90 @@ def get_top_level_shapes(self) -> ShapeList[Shape]:
...
     def vertices(self) -> ShapeList[Vertex]:
-        """vertices - all the vertices in this Shape"""
-        return Shape._get_shape_list(self, "Vertex")
+        """vertices - all the vertices in this Shape - subclasses may override"""
+        return ShapeList()
 
     def vertex(self) -> Vertex:
         """Return the Vertex"""
-        return Shape._get_single_shape(self, "Vertex")
+        return None
```
Jojain: Make all of these func abstract and don't return anything return None breaks the type hints

src/build123d/topology.py
```diff
@@ -2640,32 +2979,54 @@ def _bool_op(
...
+            Self | ShapeList[Self]: Resulting object may be of a different class than self
+                or a ShapeList if multiple non-Compound object created
         """
 
         cut_op = BRepAlgoAPI_Cut()
 
         return self._bool_op((self,), to_cut, cut_op)
 
-    def fuse(self, *to_fuse: Shape, glue: bool = False, tol: float = None) -> Self:
+    def fuse(
```
Jojain: Are we sure we want the shape list as output ? 
Raising an error avoids delegating to the user the burden of dealing with 2 return types.
But returning both type is more flexible ... tradeoffs

src/build123d/topology.py
```diff
@@ -2689,36 +3052,54 @@ def fuse(self, *to_fuse: Shape, glue: bool = False, tol: float = None) -> Self:
...
             to_intersect (sequence of Union[Shape, Axis, Plane]): Shape(s) to
                 intersect with
 
         Returns:
-            Shape: Resulting object may be of a different class than self
+            Self | ShapeList[Self]: Resulting object may be of a different class than self
+                or a ShapeList if multiple non-Compound object created
         """
 
+        def _to_vertex(vec: Vector) -> Vertex:
```
Jojain: This is wrong Shape doesn't know its subclasses, this method needs to be abstract / virtual and each cast must be handled in the associated subclass.

src/build123d/topology.py
```diff
@@ -3161,12 +3505,12 @@ def to_arcs(self, tolerance: float = 1e-3) -> Face:
         """
         return self.__class__(BRepAlgo.ConvertFace_s(self.wrapped, tolerance))
 
-    def _repr_javascript_(self):
-        """Jupyter 3D representation support"""
+    # def _repr_javascript_(self):
+    #     """Jupyter 3D representation support"""
```
Jojain: Don't forget to delete this if you are sure you want to get rid of it.

src/build123d/topology.py
```diff
@@ -3275,203 +3615,46 @@ def project_faces(
...
+            projected_faces.append(
+                projection_face.project_to_shape(self, surface_normal * -1)[0]
+            )
+
+        logger.debug("finished projecting '%d' faces", len(faces))
+
+        return ShapeList(projected_faces)
 
 
 class Comparable(metaclass=ABCMeta):
```
Jojain: Use `Comparable(ABC)` instead of the metaclass

src/build123d/topology.py
```diff
@@ -3541,10 +3728,25 @@ def filter_by(
         # could be moved out maybe?
         def axis_parallel_predicate(axis: Axis, tolerance: float):
```
Jojain: In the future it could be interesting to have a module `predicates` with the existing ones and tools to build some new more easily. (For exemple to make composition of predicates).
This could for now go outside as a standalone function.

src/build123d/topology.py
```diff
@@ -3541,10 +3728,25 @@ def filter_by(
...
             def pred(shape: Shape):
-                if isinstance(shape, Face) and shape.is_planar:
-                    shape_axis = Axis(shape.center(), shape.normal_at(None))
-                elif isinstance(shape, Edge) and shape.geom_type == GeomType.LINE:
-                    shape_axis = Axis(shape.position_at(0), shape.tangent_at(0))
+                if shape.is_planar_face:
```
Jojain: Here we see why you've put `is_planar_face` on Shape, you need to work on a the OCP object not on `Shape` and make a `is_planar_topdsface` instead

src/build123d/topology.py
```diff
@@ -8592,7 +8379,58 @@ def chamfer_2d(
         Returns:
             Wire: chamfered wire
         """
-        return Face(self).chamfer_2d(distance, distance2, vertices, edge).outer_wire()
+        reference_edge = edge
+        del edge
```
Jojain: Why do you need to del the reference towards the edge?

Jojain (commented):
I have mainly done a pass about the architecture, I didn't looked too much into depth the behaviour of the code, I assume it has been mostly moving stuff around so that there should not be intended change in behaviour.

General feedback:
1. Splitting things by topological order and relying on OCP datastructure when we need to work on a lower level is a great way to solve the coupling issue. I think if it is ever needed it would be possible to split things further apart using the same techniques (so that's good)
2. There is still some design flaws, the whole casting / downcasting with string literal used as types seems still a bit brittle. It is better than is used to be but could be further improved.
3. While I see the benefit of code reuse of mixins one could question if it's a good idea or not. Using standalone functions and calling them in each related subclass is a bit more code to write but makes things more straightforward.
4. The way some 1d mixin func are reused in 2d and same goes for 2d -> 3d seems to break the type hinting. Furthermore with some dedicated functions per class the very general type hinting could probably be refined. (the extrude in mixin1D has any shape as return type)
5. The type hints are a mess, take the `wrapped` attribute for instance, it can be `None` but sometimes it is called assuming it is not `None` sometimes its checked against. A solution is to introduce a `_wrapped` as the current one and a `wrapped` property that would raise an error if its `None` that way you can call the right one when you assume it can or cannot be `None` . Sames goes for the optionals there are sometimes provided sometimes not. 
6. Btw since Python 3.9 isn't supported anymore and since you started using the pipe operator for unions it would be nice to update the code base (there might be tools already existing) to use the pipe operator everywhere. Special note on the `Optional`, it should be either `Optional[SomeType]` or `SomeType |  None` everywhere not a mix of both. It seems the preferred way is `SomeType |  None`

Congratulations for pulling this off, it was a huge task !
If this pass all the tests a lot of what I've said above could be done after merging this. It's already very long, it may be better to just continue refactor as another PR so changes could be handled more easily

src/build123d/topology.py
snoyer: Alternatively, EAFP-style:
```py
try:
    return constructor_lut[shape_type](downcast(obj))
except KeyError:
    raise SomeException()
```
Although the key is an enum so it's arguably fine to assume that the mapping is exhaustive and there will never be a `KeyError` ?

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: For now I've reestablished this - @jdegenstein 's PR will handle this stuff

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Eliminated

src/build123d/topology.py
gumyr: Removed the ShapeList, it isn't needed.

src/build123d/topology.py
```diff
@@ -2072,11 +2438,12 @@ def bounding_box(self, tolerance: float = None, optimal: bool = True) -> BoundBo
         Returns:
             BoundBox: A box sized to contain this Shape
         """
-        return BoundBox._from_topo_ds(
-            self.wrapped, tolerance=tolerance, optimal=optimal
-        )
+        if self.wrapped is None:
+            return BoundBox(Bnd_Box())
```
snoyer: the `BoundBox` class is not ready to handle open/empty `Bnd_Box` instances

src/build123d/topology.py
```diff
@@ -2125,7 +2494,7 @@ def combined_center(
                 sum_wc = sum_wc.add(weighted_center)
             middle = Vector(sum_wc.multiply(1.0 / total_mass))
         elif center_of == CenterOf.BOUNDING_BOX:
-            total_mass = len(objects)
+            total_mass = len(list(objects))
 
             weighted_centers = []
             for obj in objects:
```
snoyer: if the iterable happens to be lazy `len(list(objects))` would consume it and `for obj in objects:` would have nothing to loop on
```suggestion
            total_mass = 0
            weighted_centers = []
            for obj in objects:
                total_mass += 1
```

src/build123d/topology.py
```diff
@@ -2625,47 +2999,77 @@ def _bool_op(
         Returns:
 
         """
+        # Find the highest order class from all the inputs Solid > Vertex
+        order_dict = {type(s): type(s).order for s in [self] + list(args) + list(tools)}
```
snoyer: You're going over `args` and `tools` again below so you don't want to just consume them here in case they're lazy
```suggestion
        args, args2 = tee(args)
        tools, tools2 = tee(tools)
        order_dict = {type(s): type(s).order for s in (self, *args2, *tools2)}
```

src/build123d/topology.py
```diff
@@ -3873,111 +4185,99 @@ def face(self) -> Face:
...
+    def __ne__(self, other: ShapeList) -> bool:  # type: ignore
         """ShapeLists inequality operator !="""
         return (
             set(self) != set(other) if isinstance(other, ShapeList) else NotImplemented
         )
 
-    def __add__(self, other: ShapeList):
+    def __add__(self, other: ShapeList) -> ShapeList[T]:  # type: ignore
         """Combine two ShapeLists together operator +"""
         return ShapeList(list(self) + list(other))
```
snoyer: unnecessary to convert to list + concat only to convert to `ShapeList`
```suggestion
        return ShapeList(chain(self, other))
```

src/build123d/topology.py
```diff
@@ -3873,111 +4185,99 @@ def face(self) -> Face:
...
         return ShapeList(set(self) - set(other))
 
-    def __and__(self, other: ShapeList):
+    def __and__(self, other: ShapeList) -> ShapeList[T]:
         """Intersect two ShapeLists operator &"""
         return ShapeList(set(self) & set(other))
 
     @overload
-    def __getitem__(self, key: int) -> T: ...
+    def __getitem__(self, key: SupportsIndex) -> T: ...
```
snoyer: Isn't `SupportsIndex` just a mixin with the `__index__` method? how does it work here?

src/build123d/topology.py
```diff
@@ -9248,6 +9307,243 @@ def unwrap_topods_compound(
...
+    """Sew faces into a shell if possible"""
+    shell_builder = BRepBuilderAPI_Sewing()
+    for face in faces:
+        shell_builder.Add(face)
+    shell_builder.Perform()
+    return downcast(shell_builder.SewedShape())
+
+
+def _make_topods_compound_from_shapes(
+    occt_shapes: Sequence[TopoDS_Shape | None],
```
snoyer:
```suggestion
    occt_shapes: Iterable[TopoDS_Shape | None],
```
and change all occurrences of
```diff
- _make_topods_compound_from_shapes([...])
+ _make_topods_compound_from_shapes(...)
```

snoyer (changes requested):
When changing `Sequence[X]` parameters to `Iterable[X]` you need to review the method's implementation because `Iterable`s can be lazy in which case you can't loop over them twice (you _can_ but they'll be empty the second time around).
If the parameters are only looped over once you're good but otherwise you need to adjust by either refactoring the algorithm or `tee`ing the iterable (or checking the type and converting to a list if necessary) 

The existing tests won't detect that unless you change them from `method_that_now_takes_an_iterable([...])` to `method_that_now_takes_an_iterable(iter([...])`

src/build123d/topology.py
gumyr: Added this to geometry.py - `Bnd_Box().IsVoid()` is checked.

src/build123d/topology.py
gumyr: I've checked them all and in a few cases converted the inputs to lists explicitly. Thanks for the pointer.

src/build123d/topology.py
gumyr: I just used lists here as the number of objects are going to be small.

src/build123d/topology.py
gumyr: Done.

src/build123d/topology.py
gumyr: SupportsIndex was required to make mypy happy; otherwise I was redefining `list`.  I had to use a few `# type : ignore` statements in `ShapeList` because mypy doesn't like redefining operators.

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: I don't see how there could be a key error here.

src/build123d/topology.py
gumyr: `_ClassMethodProxy` is gone but I've left the other two for now as I don't see any typing problems.  Could be changed if it's still a problem though.

src/build123d/topology.py
snoyer: I got mixed up between [`__index__()`](https://docs.python.org/3/reference/datamodel.html#object.__index__) and `list`'s `.index()`, my bad!

src/build123d/topology.py
```diff
@@ -9248,6 +9395,244 @@ def unwrap_topods_compound(
     return compound
 
 
+def get_top_level_topods_shapes(topods_shape: TopoDS_Shape) -> list[TopoDS_Shape]:
```
snoyer:
```suggestion
def get_top_level_topods_shapes(topods_shape: TopoDS_Shape | None) -> list[TopoDS_Shape]:
```

src/build123d/topology.py
```diff
@@ -382,24 +356,31 @@ def tuplify(obj: Any, dim: int) -> tuple:
 class Mixin1D:
```
snoyer:
```suggestion
class Mixin1D(Shape):
```
would avoid all the `assert isinstance(self, Shape)` in member methods

gumyr: I was trying to avoid doing this as the Mixin classes are defined before the Shape class in the topology.py file; however, I was forced to do this for other typing reasons and have removed all of the Shape assertions.  So ... much ... change.  There will be nothing left of the original file when I'm done :(

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: ... and undone.  This change caused strange typing problems where mypy complained about a Sequence when  generator was used even though the type was changed from Sequence to Iterable.  Worth looking at this again in the future.

src/build123d/topology.py
gumyr: ... and undone.  MacOS-13 (Python 3.12 IIRC) fails with this change; however, it works everywhere else.  More investigation is required.

src/build123d/topology.py
gumyr: Actually, a ShapeList wasn't needed here anyway - changed to just a list.

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Nice - used your version.

src/build123d/topology.py
gumyr: These types of errors have all be found by mypy and fixed.

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done - although some None handling needed to be added too.

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

src/build123d/topology.py
gumyr: Done

PR #819, opened 2024-12-16

slobberingant:
The third party SVG exporter has issues creating very small arcs. 
These arcs can be created in detailed Compound objects especially when scaling.
I've added a check to print a warning and ignore any circular elements that are very small.
The test previously failed and now passes.

Looking forward to feedback.

### DIFF
--- a/src/build123d/drafting.py
+++ b/src/build123d/drafting.py
@@ -29,7 +29,7 @@ license:
 from dataclasses import dataclass
 from datetime import date
 from math import copysign, floor, gcd, log2, pi
-from typing import ClassVar, Optional, Union
+from typing import cast, ClassVar, TypeAlias
 
 from collections.abc import Iterable
 
@@ -102,7 +102,7 @@ class Arrow(BaseSketchObject):
 
     Args:
         arrow_size (float): arrow head tip to tail length
-        shaft_path (Union[Edge, Wire]): line describing the shaft shape
+        shaft_path (Edge |  Wire): line describing the shaft shape
         shaft_width (float): line width of shaft
         head_at_start (bool, optional): Defaults to True.
         head_type (HeadType, optional): arrow head shape. Defaults to HeadType.CURVED.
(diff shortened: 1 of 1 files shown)


src/build123d/exporters.py
```diff
@@ -1187,6 +1187,9 @@ def _line_element(self, edge: Edge) -> ET.Element:
 
     def _circle_segments(self, edge: Edge, reverse: bool) -> list[PathSegment]:
         # pylint: disable=too-many-locals
+        if edge.length < 1e-6:
+            print("Skipping arc that is too small to export safely (length < 1e-6)")
```
gumyr: Shouldn't this be a warning? Something like:
```py
            warnings.warn("Skipping arc that is too small to export safely (length < 1e-6)", stacklevel=2)
```
I'm not sure what the `stacklevel` should be here - I guess it should point at the `export_svg` call.

slobberingant: I had it as a warning before but the warning came up when I ran the test suit which I didn't think was the desired outcome. Do I need to change the test somehow so that the warning is not shown when the whole test suit is run?

gumyr: The test can be written like this:
```py
        with self.assertWarns(UserWarning):
            Wire.make_rect(1, 1).vertex()
```

slobberingant:
See revised commit.
Diving further I needed to do the edge length check on the ellipse_segment also. Associated test has been added. Now that this check is done twice, would you prefer a separate method or leave as is?

Not sure regarding stacklevel. Some experimentation resulted in:
- stacklevel=2 points to _circle_segments or _ellipse_segments.
- stacklevel=6 points to _add_single_shape which is part of the ExportSVG class.
- stacklevel=7 points to add_shape which is the calling method.

gumyr:
> Not sure regarding stacklevel. Some experimentation resulted in:
>     * stacklevel=7 points to add_shape which is the calling method.

This seems weird but will make most sense to users so let's go with that.

slobberingant:
Ok. Leave it with me. I will write another test based on the codecov result and rebase the commits.

slobberingant:
Updated tests to check for empty list return as well as warning to satisfy codecov.
Stacklevel has been updated to 7 as discussed.
I had a go at squashing the commits but in the end I didn't feel comfortable messing with git history.

gumyr:
There has been a lot of churn in the code and now there are conflicts with this PR.  Once those are fixed we'll get this closed.

slobberingant:
New pull request with conflicts resolved with clean commit has been created.

PR #820, opened 2024-12-17
pyproject.toml add optional dependencies, jupyter_tools is optional

jdegenstein:
Opening as a draft PR for now for further consideration and also not intended to merge until PR https://github.com/gumyr/build123d/pull/812 is complete. Open to naming suggestions for the optional dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,8 +66,10 @@ ocp_vscode = [
 
 # development dependencies
 development = [
+    "wheel",
     "pytest",
-    "coverage",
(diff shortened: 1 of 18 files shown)


victorpoughon:
Hi! I just saw that PR after I was done working on #829. Good changes :) I think maybe you might wanna call it `jupyter_display` instead of `jupyter_tools` as that's a bit more clear to what it does. But that's minor.

I see you also added a `tests/test_jupyter_tools.py`, so I'll probably have to update that in my PR.

PR #821, opened 2024-12-17
Workflow cleanup

jdegenstein:
`coverage.yml` was not set up the same way as other workflows, this PR fixes that and deletes the now unnecessary `requirements.txt` file by moving the `pytest-cov` installation elsewhere.


jdegenstein:
In the process of fixing one issue, I discovered that `.github/actions/setup/action.yml` was not setup correctly to accept the input python version. I also fixed that issue and now everything is working as expected.

PR #824, opened 2024-12-18
Fix function doc for bounding_box() in operations_generic.py

ZzetT:
fixed inconsistency between method signature and documentation


ZzetT:
closed because there are more places where this inconsistency occours

PR #838, opened 2024-12-30
simplify `import_svg` API

snoyer:
Importing an SVG file from Inkscape using the labels set from the "Layers and Objects" panel currently requires passing both `label_by="label"` and `is_inkscape_label=True` to `import_svg()`.

This PR proposes to remove the `is_inkscape_label` parameter and allow using `label_by="inkscape:label"` (matching the attribute as it appears in the XML: `<path d="..." inkscape:label="foo" />`)

The implementation is basically handling the `inkscape:` prefix in the `label_by` values as a shortcut for the actual `{http://www.inkscape.org/namespaces/inkscape}` namespace.
This should be future-proof enough to handle more shortcuts for other software in addition to Inkscape if the need arises (as opposed to adding more `use_xxx_label: bool` parameters as per the current approach)


jdegenstein (approved):
Looks good to me

gumyr:
Thank you!

PR #849, opened 2025-01-07
Add a benchmark workflow to build123d

jdegenstein:
This adds a new `tests/test_benchmarks.py` file which adds 12 new complex tests based on this  page of the docs https://build123d.readthedocs.io/en/latest/tttt.html (some of which had to be updated and fixed, which I will do in a separate PR)

If the `pytest-benchmark` package is installed, this will run these 12 new tests in a "benchmarking mode". If this package is NOT installed these new tests will EDIT: not run as normal tests.

In order to have this automatically run on new PRs this PR also adds a "benchmarks" github actions workflow that ONLY runs the benchmarking tests in benchmarking mode.

The motivation for this PR is to ensure that it is possible to examine the performance impacts of any new PRs across all 4 major platforms. In the past we have observed significant performance regressions that e.g. only happened on macos-arm64. I chose to limit to Python 3.12 for now.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -14,5 +14,5 @@ runs:
     - name: Install Requirements
       shell: bash
       run: |
-        pip install wheel mypy pytest pytest-benchmark pytest-cov pylint 
+        pip install wheel mypy pytest pytest-cov pylint 
         pip install .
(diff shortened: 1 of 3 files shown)


jdegenstein:
@gumyr this is ready for review

gumyr:
Looks good to me at this point.  I don't think the extra time is going to be a problem with just git push but I'll let you know if it's a problem.

It's good to track performance over time so this is an important addition to build123d.

PR #852, opened 2025-01-08
Upgrade to Python 310+ syntax

jdegenstein:
Most of the changes are related to eliminating the use of Union, but there are also some changes to other things like string literal types and some changed imports as well. The changes were generated by the pyupgrade package.


gumyr:
These all look good to me.  Would you merge them as I'm making changes to geometry.py and will have to merge my changes to what pyupgrade has done here?  My refactor script could have used pyupgrade had I known about it.

PR #853, opened 2025-01-08
objects_sketch.py -> mypy typing improvements 

jdegenstein:
Solves most issues with the module, but I am not sure about this one:
```py
class BaseSketchObject(Sketch):
...
    def __init__(
        self,
        obj: Compound | Face | None, # added or None, which is not handled well upstream
```


gumyr:
If there are places where `None` isn't handled well you could just leave the typing problem there for now and we can circle back and look into it in depth.  The `geometry.py` module contains many tricky problems do to the separation of geometry and topology - I might do a first pass with a few problems remaining.

jdegenstein:
@gumyr Sounds good, I went ahead and reverted that change. As far as I know this is ready to merge.

PR #861, opened 2025-01-13

slobberingant:
Small arcs and ellipses did not export properly to SVG.
Code has been added to ignore arcs and ellipses smaller than can safely be exported.
Tests have also been added.


slobberingant:
Sorry I have no idea why the checks are all failing now. The code added is as before. Conflict has been resolved to the best of my knowledge. I will need some hand holding on this one.

PR #862, opened 2025-01-13

slobberingant:
Small arcs and ellipses did not export properly to SVG.
Code has been added to ignore arcs and ellipses smaller than can safely be exported.
Tests have also been added.


gumyr:
Thank you!  Timing made this more difficult than it should have been, thanks for your patience.

PR #872, opened 2025-01-15
readthedocs fixes to topology inheritance diagram, add mixin classes …

jdegenstein:
…to `topology/__init__.py`, delete docs/requirements.txt, streamline workflows using new optional dependencies


gumyr:
This looks good to me.  @jdegenstein thanks for fixing this and all that you do!

PR #875, opened 2025-01-17
use `BRep_Builder` for single face shell

snoyer:
Tried making a shell from the single curved face of a cone, got segfault from `BRepBuilderAPI_MakeShell`. `BRep_Builder` seems to work fine in that case.
I'm not sure this is a 100% reliable solution but according to the tests it's not worse.

### DIFF
--- a/src/build123d/topology/two_d.py
+++ b/src/build123d/topology/two_d.py
@@ -62,7 +62,7 @@ from typing import Any, Tuple, Union, overload, TYPE_CHECKING
 from collections.abc import Iterable, Sequence
 
 import OCP.TopAbs as ta
-from OCP.BRep import BRep_Tool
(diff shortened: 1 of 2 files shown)


gumyr:
Oh, sorry I split the test_direct_api.py file into 32 separate files which breaks this PR.  It should be easy to fix though as the content of the tests hasn't changed so `test_shells.py` should be the new home of this test.

Otherwise, thanks for this change.  This lower level implementation looks better in handling the case of a non-planar face which doesn't seem like it should be too difficult.

snoyer:
I can't fix it right now but you should have write access to the PR branch if you want to resolve

gumyr:
Thanks for the change.  It was going to be a pain to fix the PR so I just applied your changes directly. Hope you don't mind.

PR #877, opened 2025-01-20
pyproject.toml -> add optional `cadquery-ocp-stubs` to [development] optional extras

jdegenstein:
Please review version pin too. Tests currently fail because this package does not yet exist on PyPI, therefore opening as a draft PR for now.

Also simplified version pin for `cadquery-ocp`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ classifiers = [
 ]
 
 dependencies = [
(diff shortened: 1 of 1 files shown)


gumyr (approved):

gumyr:
It will be nice to get all of this stuff installed with a single command.

PR #879, opened 2025-01-25
allow to filter and group by property

snoyer:
This would allow 
```py
shape.edges().filter_by(Edge.is_interior)
```
as an alternative to 
```py
shape.edges().filter_by(lambda e: e.is_interior)
```

### DIFF
--- a/src/build123d/topology/shape_core.py
+++ b/src/build123d/topology/shape_core.py
@@ -2447,7 +2447,10 @@ class ShapeList(list[T]):
         if callable(filter_by):
             predicate = filter_by
         elif isinstance(filter_by, property):
-            predicate = filter_by.__get__
+
+            def predicate(obj):
+                return filter_by.__get__(obj)
+
         elif isinstance(filter_by, Axis):
             predicate = axis_parallel_predicate(filter_by, tolerance=tolerance)
         elif isinstance(filter_by, Plane):
--- a/src/build123d/topology/shape_core.py
+++ b/src/build123d/topology/shape_core.py
@@ -2351,7 +2351,7 @@ class ShapeList(list[T]):
 
     def filter_by(
         self,
-        filter_by: ShapePredicate | Axis | Plane | GeomType,
+        filter_by: ShapePredicate | Axis | Plane | GeomType | property,
         reverse: bool = False,
         tolerance: float = 1e-5,
     ) -> ShapeList[T]:
@@ -2446,6 +2446,8 @@ class ShapeList(list[T]):
         # convert input to callable predicate
         if callable(filter_by):
             predicate = filter_by
+        elif isinstance(filter_by, property):
+            predicate = filter_by.__get__
         elif isinstance(filter_by, Axis):
             predicate = axis_parallel_predicate(filter_by, tolerance=tolerance)
         elif isinstance(filter_by, Plane):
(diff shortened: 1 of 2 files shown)


gumyr:
@snoyer thank you for sharing your mastery of Python.  Was there a reason to not include `sort_by` in this PR?

Using shape properties directly as `sort_by`/`group_by`/`filter_by` parameters is very elegant and makes me wonder about the whole `SortBy` enum - should it be replaced entirely?  Currently one might `group_by(SortBy.AREA)` which looks kinda weird.  `group_by(Face.area)` makes more sense to me and is future proof as need properties would be automatically supported.  Should the use of `SortBy` be deprecated?

The `GeomType`/`geom_type` enum/property should probably stay though as I don't see how one could use this property without using a lambda.

snoyer:
> Was there a reason to not include `sort_by` in this PR?

It didn't occur to me at all! I wanted to do `filter_by` and remembered `group_by` worked basically the same but I just didn't think of sorting. But you are right it should work there too.

Regarding `SortBy` I'm not sure, it looks like you've got at least `SortBy.DISTANCE` that would be a bit involved to replace, but even the other ones do a bit more than just fetching the property because of the rounding in `group_by` (although I guess you could check the type of the property value and do the rounding if it's a number)

Either way, let me know if I should add `sort_by` to this PR (property support, that is) or if you want to do it later once you've decided what to do with the enum.

jdegenstein:
>Should the use of SortBy be deprecated?

If this is possible as described then absolutely yes it should be deprecated. The naming is already confusing since `group_by(SortBy.XYZ)` has been possible for a long time.

> Regarding SortBy I'm not sure, it looks like you've go at least SortBy.DISTANCE that would be a bit involved to replace

Good point.

gumyr:
Please add `sort_by`. For now let's leave the `SortBy` changes as I'd like to look into changing `SortBy.DISTANCE` so it has the flexibility to work on any point not just the origin.

snoyer:
> Please add `sort_by`.

done

> I'd like to look into changing `SortBy.DISTANCE` so it has the flexibility to work on any point not just the origin.

Turns out you can already do `.sort_by(other_shape.distance)` so maybe it can just be dropped?

```py
from itertools import product

from build123d import Box, ShapeList, Solid, Vertex
from ocp_vscode import ColorMap, show

boxes = ShapeList(
    Box(1, 1, 1).scale(0.75 if (i, j) == (1, 2) else 0.25).translate((i, j, 0))
    for i, j in product(range(-5, 6), repeat=2)
)

# sort by distance to origin
boxes = boxes.sort_by(Vertex().distance)
show(*boxes, colors=ColorMap.listed(len(boxes)))
```
image

```py
# sort by distance to largest box
boxes = boxes.sort_by(boxes.sort_by(Solid.volume).last.distance)
show(*boxes, colors=ColorMap.listed(len(boxes)))
```
image

gumyr:
Thank you.  This is a really nice enhancement. I need to get my head around `boxes = boxes.sort_by(boxes.sort_by(Solid.volume).last.distance)` - cool that it works though.

snoyer:
@gumyr 
> Thank you. This is a really nice enhancement. I need to get my head around `boxes = boxes.sort_by(boxes.sort_by(Solid.volume).last.distance)` - cool that it works though.

You're welcome :)

This last example was a bit dense on purpose but if we break it down we have
```py
largest_box = boxes.sort_by(Solid.volume).last  # sorting by property as per this PR
boxes = boxes.sort_by(largest_box.distance)     # sorting by callable as was already possible
```
the key here is that the `Shape` class has a `def distance(self, other: Shape) -> float` method so:
- `Shape.distance` is a `(Shape, Shape) -> float` function
- `a_shape.distance` is `(Shape) -> float` function, and it can be used it to sort/filter/group

gumyr:
Got it, thanks.  There will need to be some good documentation around these capabilities as it probably will be a little more difficult for new users to understand as there isn't a simple list of things that can be done now (which is kinda the whole point but still...).

PR #905, opened 2025-02-15
make `Axis.position` and `Axis.direction` properties

snoyer:
The `.position` and `.direction` members of the `Axis` class are currently copies made at initialization which means that they are modifiable but not kept in sync with the underlying `.wrapped` object.

```py
axes = [Axis.Y for _ in range(10)]
for i, a in enumerate(axes):
    a.position += (i, 0, 0)  # translating each axis
show_all()  # all axes show at (0,0,0)
```


gumyr:
Thank you!

PR #908, opened 2025-02-21
add `align` parameter to `import_svg`

snoyer:
`ocpsvg >= 0.5` now properly interprets the viewbox when importing SVG documents and PR takes advantage of this change to introduce an `align` parameter to the `import_svg` function.

The default is set to `align=Align.MIN` to keep the current import behavior of positioning the bottom left corner of the document at `(0,0)`. `align=None` is allowed and can be used to preserve the viewbox origin.


gumyr:
Nice feature, thanks!

PR #911, opened 2025-02-22
Run tests in parallel by default, and update CONTRIBUTING.md

fischman:
Also:
- replace explicit development dependencies list with `.[development]`; and
- replace reference to docs/requirements.txt (deleted in d78ca933fcda7ff2d54a8664a560418f6f247a60) with `.[docs]`.
- `test_mesher.py` was reusing the same filename across tests which meant that when running in parallel tests would stomp on each other. Robustified by having each test use a distinct file name.


gumyr:
Nice - thank you!

Issue #909, opened 2025-02-22
Examples should be kept working by being run in a test


fischman:
@bernhard-42 said on [Discord](https://discord.com/channels/964330484911972403/1074840524181217452/1341107076251779233):
> I typically run through all the build123d examples when testing a new release.
This time I see some errors (haven't tried all of the examples though). Looks like the examples have not been updated for the breaking changes of the topology refactoring.
Nothing important or urgent, but maybe it could be fixed at some point of time?

@gumyr replied:
> Thanks, I know that some things are broken. I need to create a test for the examples so the problems get flagged immediately.

Filing this issue to track one way for how such a test could work.
My first cut at this points out at least one problem: while most of the examples are small and run quickly, others take a lot longer. Total for all examples right now is 3m51s, with the slowest example taking 17.6s.

fischman:
FTR @gumyr fixed the examples so they work again in https://github.com/gumyr/build123d/commit/aeb6b32b6594b51209a70ee1922ff63a00080dc1 (thanks!).

fischman:
Execution time is somewhat mitigated by running tests in parallel: #911.

PR #912, opened 2025-02-23
TooTallToby tutorials: unbreak and test.

### LINKED ISSUE #848: Fix TTT tutorials: 23-T-24 Curved Support, 01-02-post-cap, 01-10-light-cap
https://build123d.readthedocs.io/en/latest/tttt.html#t-24-curved-support is not currently working correctly on build123d==0.8.0 (and higher most likely)

fischman:
- Unbreak the three broken tutorials (fixes #848)
  - This involved a rewrite of PPP-01-10 because I already had my own solution to that one and I couldn't easily tell what was going wrong with the previous solution.
- Add assertions to all the tutorials so that non-raising means success
- Add the TTT examples to `test_examples.py` added recently for #909
- Also `dos2unix`'d the TTT .py files to match the rest of the codebase.
- Drop the copies of TTT examples from test_benchmarks.py and read them directly from the docs directory now.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -22,5 +22,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: benchmark
         run: |
-          pip install pytest-benchmark
           python -m pytest --benchmark-only
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,4 +23,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: test
         run: |
-          python -m pytest -n auto
+          python -m pytest -n auto --ignore=tests/test_benchmarks.py 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,4 +23,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: test
         run: |
-          python -m pytest -n auto --ignore=tests/test_benchmarks.py 
+          python -m pytest -n auto --benchmark-disable
--- a/docs/assets/ttt/ttt-23-02-02-sm_hanger.py
+++ b/docs/assets/ttt/ttt-23-02-02-sm_hanger.py
@@ -92,6 +92,13 @@ with BuildPart() as sm_hanger:
     mirror(about=Plane.YZ)
     mirror(about=Plane.XZ)
 
-print(f"Mass: {sm_hanger.part.volume*7800*1e-6:0.1f} g")
+got_mass = sm_hanger.part.volume*7800*1e-6
+want_mass = 1028
+tolerance = 10
+delta = abs(got_mass - want_mass)
+print(f"Mass: {got_mass:0.1f} g")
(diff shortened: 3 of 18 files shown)


fischman:
FYI @jdegenstein. LMK if you'd rather I not replace your PPP0110 solution (and if you have a fix in mind already).

fischman:
@gumyr WDYT, worth testing TTT examples as part of automated testing?

jdegenstein:
@fischman FYI TTT examples are (sort of) tested through the benchmarking workflow (in a separate python script)

fischman:
> @fischman FYI TTT examples are (sort of) tested through the benchmarking workflow (in a separate python script)

Thanks for the heads-up, I was unaware of [that](https://github.com/gumyr/build123d/blob/1d8980441798fd35984eccc44e78f0520e04ea07/tests/test_benchmarks.py#L9-L679) before. 
Seems suboptimal that the examples' code is duplicated in the tests/ subdirectory (because of the likelihood of drift as one or the other place is edited, and the other is forgotten, and also because edits must be duplicated). Also ttt-23-t-24-curved_support is missing from the test.

Do you prefer to keep the structure at HEAD, or are you suggesting I keep this PR moving forward and delete the test_benchmarks copy of the examples, or something else?
(maybe test_benchmarks should run test_examples?)

jdegenstein:
@fischman yes I agree that having two versions is suboptimal, I was just trying to get something up and running for benchmarking purposes with some changes that were happening on the OCCT/OCP and build123d side. Yes, I would like `TTT 23-T-24` to be part of the benchmarks.

Can we simply import the models from docs/ into test_benchmarks.py ? That would probably be the "cleanest" solution requiring less work to maintain long-term.

RE: PPP0110 -- I am fine with your solution, I am not married to what I created. Hopefully OCCT 7.9 will make creating this model easier.

fischman:
@jdegenstein dropped the copies in the test_benchmarks.py file and now pulling from the docs copy. PTAL.

jdegenstein:
@fischman looking at the way that [benchmarks] category was eliminated resulted in benchmarks trying to run then failing because of multiple threads in the tests workflow.
```
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/pytest_benchmark/logger.py:39: PytestBenchmarkWarning: Benchmarks are automatically disabled because xdist plugin is active.Benchmarks cannot be performed reliably in a parallelized environment.
```
If this is the way it is setup, I would prefer that an appropriate flag is passed to pytest to explicitly disable benchmarking for the tests workflow. The current behavior seems fragile to me.

fischman:
@jdegenstein good catch! Fixed to exclude benchmarks from the tests workflow.

fischman:
@gumyr @jdegenstein any idea about the CI [failure](https://github.com/gumyr/build123d/actions/runs/13665527519/job/38205976954?pr=912)? Seems unrelated to this PR.

jdegenstein:
@fischman I am going to take one more quick look, I don't believe the CI failure is related to anything you did -- just a random problem with the runner I suspect.

jdegenstein:
OK I am merging despite what appears to be a different random CI failure.

EDIT: thanks for your contribution @fischman

PR #914, opened 2025-02-24
Avoid deepcopy'ing Shape.topo_parent.

fischman:
Speeds up benchy example from 27s to 5.5s.

Motivated by #909 adding examples that take a while to run.
Discussion in Discord: [thread](https://discord.com/channels/964330484911972403/1343338604180803646).

### DIFF
--- a/src/build123d/topology/shape_core.py
+++ b/src/build123d/topology/shape_core.py
@@ -866,7 +866,10 @@ class Shape(NodeMixin, Generic[TOPODS]):
         if self.wrapped is not None:
             memo[id(self.wrapped)] = downcast(BRepBuilderAPI_Copy(self.wrapped).Shape())
         for key, value in self.__dict__.items():
-            setattr(result, key, copy.deepcopy(value, memo))
+            if key == 'topo_parent':
+                result.topo_parent = value
+            else:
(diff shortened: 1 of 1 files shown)


gumyr:
I'm going to try this out...

fischman:
> I'm going to try this out...

Any findings?
(no rush, just curious)

gumyr:
My current project uses selectors extensively so it should exercise this code thoroughly.  Mainly I'm selecting faces which aren't as numerous as the vertices of the benchy example so I'm seeing only modest improvements in performance - say 308 seconds to 297 seconds for a long test. Functionality seems to be fine but I've seen some things I can't explain so more investigation is required.

gumyr:
After more testing this PR look good to go for me. Are you ready?

fischman:
@gumyr I don't have any reason not to merge this PR, but also my understanding of the intended semantics of the build123d implementation is near-non-existent, so I defer to you entirely w.r.t. the correctness of this change.
If it seems right to you then I suggest merging it and keeping eyes open in case a revert is necessary.

gumyr:
Merged.  The PR easily searchable if we suspect something unexpected is happening in the future.  Thanks for working this out!

PR #917, opened 2025-02-27
feat: switch to lib3mf official pypi release from py_lib3mf

### LINKED ISSUE #799: Switch to lib3mf from py_lib3mf
3MF Consortium has released an official lib3mf package, look at switching to their package from py_lib3mf.

jdegenstein:
Previously there was no official release of lib3mf on pypi, so I instead created and published py_lib3mf to pypi. The lib3mf release on pypi was recently improved to the point that it now surpasses the py_lib3mf that I bundled -- namely to explicitly separate the library files for each platform (linux, macos, windows) and have 3 separate wheels (macos has a universal2 wheel).

The only "downside" of switching is that there is not (and never has been) support for other CPU architectures such as linux-aarch64. It was previously possible to install build123d on e.g. a raspberry pi (with some workarounds) and py_lib3mf/lib3mf would appear to install successfully (but there was not actually an appropriate *.so file for this platform/CPU type). This ambiguity has been corrected in the recent release of lib3mf, which means it is now worth switching in my opinion.

Some other important bugs have also been corrected in lib3mf 2.4.x as well such as https://github.com/3MFConsortium/lib3mf/issues/355

This PR fixes https://github.com/gumyr/build123d/issues/799

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -41,5 +41,5 @@ ignore_missing_imports = True
 [mypy-setuptools_scm.*]
 ignore_missing_imports = True
 
-[mypy-py_lib3mf.*]
+[mypy-lib3mf.*]
 ignore_missing_imports = True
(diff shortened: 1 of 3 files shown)


gumyr:
Thanks!  @jdegenstein you put a lot effort into the whole lib3mf thing by creating a custom repo then replacing it. Seems like things are going in the right direction with 3MF supporting their own code.

PR #923, opened 2025-03-01
Update type signature for component getters

mingmingrr:
Pyright complains when calling `faces()` / `edges()` / etc due to the missing default argument in the type signature. Using Protocol allows the default argument to be picked up by the type checker.


gumyr:
Thank you!

mingmingrr:
Looks like this got reverted in 10a466f. Was this failing in mypy or what was the reason?

PR #930, opened 2025-03-06
Revise docstrings for build objects

jwagenet:
Update object_*.py docstrings for more consistency. Clarify how parameters work and objects are constructed.

General improvements:
- follow a similar opening description: `Add/subtract a {object} defined by...` with more explicit behavior
- update union type declarations in string
- remove plurality references (sketch, part). Methods to create multiple parts/sketches occur outside the objects
- remove angles plurality from 1d and 2d objects
- remove hanging end stops (.) from lists
- try to make arg descriptions more consistent and descriptive
- try to specify viable enums if mentioned and CAPITALIZE
- other editing for wordiness

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -31,7 +31,6 @@ from __future__ import annotations
 import copy as copy_module
 from math import copysign, cos, radians, sin, sqrt
 from scipy.optimize import minimize
-from typing import Union
 
 from collections.abc import Iterable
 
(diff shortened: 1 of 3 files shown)


gumyr:
Thank you for doing this.  I only have one point to discuss:

All of the object description now start with "Add" which, given we're trying to tidy up the wording, might be a little confusing.  Consider:
```py
my_thing -= Circle(...)
```
where the docstring states that it will be added but the code is going to subtract it.  Same with:
```py
with BuildPart() as my_thing:
    ...
   Cylinder(..., mode=Mode.SUBTRACT)
```
How about we use a more neutral word like "Create" or something similar?

jwagenet:
Yes, I agree, using _Create_ is better because the object is created regardless of what happens in the context it modifies. I was going to use _Create_, but switched back to _Add/Subtract_ to align with the default mode because of Hole/CounterSink/CounterBore. I can update.

gumyr:
Thank you!

PR #931, opened 2025-03-06
PolarLine() make localized direction a unit vector to only scale by length

jwagenet:
Update `PolarLine()` with potentially breaking change to `direction` kwarg
- Add direction to docstring 
- Make localized direction unit vector with `normalize()`
- Update tests for angle/direction parity

Currently, when `PolarLine()` is defined by `direction` the result is a line made by `direction` using the vector magnitude and then scaling by `length`. This is unexpected behavior based on how `length` + `angle` works; **only `length` should scale the resulting line**.

This change normalizes the localized direction so e.g. `PolarLine((0,0), 3, angle=45)`  and `PolarLine((0,0), 3, direction=(1,1))` produce the same result.

### Justification for acceptable breakage
Most people are probably giving `direction` a unit vector already through `%`

### Uses within the codebase
Affected:
- `/docs/objects_1d.py` Shortens JernArc svg dashed tangent line by ~.06. Cosmetic, not fixing or rebuilding docs

Unaffected unit vectors:
- `/docs/objects_1d.py` 2 other places
- `/examples/pegboard_j_hook_algebra.py` 3 places
- `/examples/pegboard_j_hook.py` 3 places
- `/docs/assets/ttt/ttt-23-02-02-sm_hanger.py` 2 places

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -711,10 +711,11 @@ class PolarLine(BaseEdgeObject):
     Args:
         start (VectorLike): start point
         length (float): line length
-        angle (float): angle from the local "X" axis.
+        angle (float, optional): angle from the local "X" axis
+        direction (VectorLike, optional): point to determine angle
         length_mode (LengthMode, optional): length value specifies a diagonal, horizontal
             or vertical value. Defaults to LengthMode.DIAGONAL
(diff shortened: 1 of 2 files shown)


gumyr:
Not normalizing the direction vector was definitely an oversight.  Thank you for your comprehensive investigation and fix!

Issue #833, opened 2024-12-24
add `project` to cheat sheet 


jdegenstein:
add [operations_generic.project](https://build123d.readthedocs.io/en/latest/operations.html#operations_generic.project) to cheat sheet https://build123d.readthedocs.io/en/latest/cheat_sheet.html

Issue #900, opened 2025-02-09
Plane rotation over Z-axis not working as expected


Niketin:
I stumbled upon this line in the documentation while playing with planes and rotating them: 
https://github.com/gumyr/build123d/blob/f22f54af5f4ce5e648d1932d6872cbe572b3d5ad/src/build123d/geometry.py#L2611-L2612

I believe rotation of a plane does not work as described. Rotating plane over Z-axis does not rotate it over its own Z-axis, but over the world's Z-axis. I will demonstrate my observation below.

```python
p1 = Plane.XZ
p2 = p1.rotated((0, 0, 90))
p3 = p2.rotated((0, 0, 90))
p4 = p3.rotated((0, 0, 90))

p2 = p2.move(Location((10, 0, 0)))
p3 = p3.move(Location((20, 0, 0)))
p4 = p4.move(Location((30, 0, 0)))

show(
    location_symbol(p1.location, l=3),
    location_symbol(p2.location, l=3),
    location_symbol(p3.location, l=3),
    location_symbol(p4.location, l=3),
)
```
In the example, I chose Plane.XZ as the base plane. Then I created copies of it, each rotated by 90 degrees from the previous plane, around the Z-axis. Then, the planes are moved so that they are not on top of each other, for easier visualization.
![Image](https://github.com/user-attachments/assets/d285170d-e1e0-49f3-9beb-c10a68294ac2)
As can be seen from the picture, The plane is rotated around its Y-axis, which is parallel to the world's Z-axis.

Is it a bug? In case it is not a bug and it works as intended, I have a follow-up question: how can I rotate a plane around its own Z-axis?

gumyr:
The rotations are global.  You want to set/change the Plane's x_dir.  
```py
pln0 = Plane.XY
pln1 = Plane(origin=pln0.origin, x_dir=(0, 1, 0), z_dir=pln0.z_dir)
```

Niketin:
@gumyr I see. Then the part of the documentation referred above doesn't really make sense to me. The statement seems false in the context of rotating a plane using this method.

Issue #899, opened 2025-02-09
The documentation is missing the definition for plane_symbol helper function


Niketin:
In the documentation, there is a definition for a helper function `location_symbol` https://github.com/gumyr/build123d/blob/f22f54af5f4ce5e648d1932d6872cbe572b3d5ad/docs/location_arithmetic.rst?plain=1#L14-L15 and it is used in the following example code. https://github.com/gumyr/build123d/blob/f22f54af5f4ce5e648d1932d6872cbe572b3d5ad/docs/location_arithmetic.rst?plain=1#L27

There is a similar function for planes, `plane_symbol`, and it is also used in the same way as `location_symbol`. https://github.com/gumyr/build123d/blob/f22f54af5f4ce5e648d1932d6872cbe572b3d5ad/docs/location_arithmetic.rst?plain=1#L40
The problem is that the function's definition is not documented anywhere on the page (or anywhere else as far as I know).

Issue #809, opened 2024-12-06
Cannot Downlad STEP file for Hinge Tutorial


JiffyRob:
I was working through the hinge tutorial and found that there wasn't any builtin way to download the STEP file for the countersunk screw (https://github.com/gumyr/build123d/blob/dev/docs/M6-1x12-countersunk-screw.step) off of the docs website.  I managed to pull it from the link above, but it ought to be able to be downloaded from the docs.

PR #933, opened 2025-03-10
Quick fix documentation issues

jwagenet:
Quick fixes for some gh issues and minor things I found or saw mentioned in discord.

Resolves:
- #761 : add note for how `cone_angle` affects radius
- #809 : add links in docs to available example imports
- #899 : add missing `plane_symbol` def. The circle radius looks close enough
- #900 : remove note about how `Plane.rotated()` works about z

I want to specifically discuss change 96d9875a7b270fbc45b0b785b09e044367290c07, to make sure I haven't misunderstood the behavior: chaining `rotated()` doesn't appear to actually change the final rotation orientation because the rotation order is still `x > y' > z''` even if x and y are 0. If the desired rotation is `z > y' > x''` then `ordering=Intrinsic.ZYX` seems to be correct.

```python
# these produce same result
Plane.XY.rotated((0,0,5)).rotated((0,10,0)).rotated((15,0,0))
Plane.XY.rotated((15,10,5))

# whereas I believe this is the desired result the docstring incorrectly described
Plane.XY.rotated((5,10,15), Intrinsic.ZYX)
```

I have not added `Intrinsic` and `Extrinsic` to Cheat Sheet: Enums, but I could do that.


jwagenet:
I considered targeting some more issues, but I think they can fit in another scope. @jdegenstein clarified `Plane().rotated()` behavior changed when `Intrinsic/Extrinsic` was added but the docstring may have been neglected, so I'm moving this forward.

jdegenstein:
First off thanks for these changes, I have reviewed and so far they look good to me.

> I have not added Intrinsic and Extrinsic to Cheat Sheet: Enums, but I could do that.

Yes, please do that. Can you also cross-check other enums? I believe they are all supposed to be on cheat sheet.

Can you also address #833 ?

jwagenet:
Done.

I left `Align.NONE` off but all the other enums should be there.

Now Resolves #833

gumyr:
Thank you!

Issue #924, opened 2025-03-01
Separate out dependencies that are not necessary for 3D object creation


BlueDrink9:
To simplify the installation of build123D for use as a build library, rather than as a design process tool, it would be good to explicitly separate out those dependencies that are only for interactive use. This would then be specified as a special argument to the pip installation, e.g. `pip install "build123d[interactive]"`. This would have the downside of slightly complicating installation for users that need that feature (but this should be easy to remedy with an include guard that prints out the required installation steps for the pretty printing action).

The crux of this issue really is ipython, which is used for pretty printing in build123d but breaks Google Collab when updated (explicitly - they only support one version). I'd like to use Collab to allow people to easily run and customize objects with [snakeskin](https://GitHub.com/BlueDrink9/keeb_snakeskin) without installing it on their own machine.

As an alternative, would you welcome a PR to remove the ipython dependency in favour of some simpler dependency?

gumyr:
There is a larger effort to separate all of the display functionality out of build123d so in general there is alignment to this request.  `jupyter_tools.py` would still need to be handled.

With respect to the `IPython.pretty` functionality, how about replacing it with the follows?
```py
from build123d import *
from build123d import GroupBy
from typing_extensions import Iterable

v = Box(1, 1, 1).vertices().group_by(Axis.Z)
print(v)


def GroupBy__repr__(self):
    return self._custom_pretty()


def GroupBy_custom_pretty(self, level=0):
    """
    Recursively build a bracketed, multi-line representation.
    """
    indent = "  " * level
    lines = []
    lines.append(indent + "[")  # opening bracket

    for idx, item in enumerate(self):
        if idx > 0:
            lines[-1] += ","  # append comma to previous line
        if not isinstance(item, (Vector, Vertex)) and isinstance(item, Iterable):
            lines.append(GroupBy_custom_pretty(item, level + 1))
        else:
            lines.append(indent + "  " + str(item))

    lines.append(indent + "]")  # closing bracket
    return "\n".join(lines)


GroupBy.__repr__ = GroupBy__repr__
GroupBy.__str__ = GroupBy__repr__
GroupBy._custom_pretty = GroupBy_custom_pretty
print(v)
```
```
[[Vertex(-0.5, -0.5, -0.5),
  Vertex(-0.5, 0.5, -0.5),
  Vertex(0.5, -0.5, -0.5),
  Vertex(0.5, 0.5, -0.5)],
 [Vertex(-0.5, -0.5, 0.5),
[...]

PR #942, opened 2025-03-18
Use lazy imports in topology to speed up library import.

jsmnbom:
Loads IPython and scipy only when actually needed. Made build123d load about 1 second (on average) faster on my machine. Python caches modules so all in all this should only be a net benefit.

This is nice when needing to restart the jupyter kernel and having to not wait as long for the library to import.
It would be even better if we could perhaps split the exporters and importers into seperate submodules, and require the user to be explicit about importing them. This would allow ezdxf and svgpathtools to be lazily imported, allowing for about half a second speed up (theoretically).

### DIFF
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -71,7 +71,7 @@ from build123d.topology import (
 from build123d.build_common import UNITS_PER_METER
 
 if TYPE_CHECKING:
-    import svgpathtools as PT
+    import svgpathtools as PT  # pylint: disable=C0415
     PathSegment: TypeAlias = PT.Line | PT.Arc | PT.QuadraticBezier | PT.CubicBezier
     """A type alias for the various path segment types in the svgpathtools library."""
 
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -30,6 +30,8 @@ license:
 # pylint: disable=no-name-in-module, import-error
 # pylint: disable=too-many-lines
 
+from __future__ import annotations
+from typing import TYPE_CHECKING
 import math
 import xml.etree.ElementTree as ET
 from copy import copy
@@ -41,7 +43,6 @@ from warnings import warn
 from collections.abc import Callable, Iterable
 
 import ezdxf
-import svgpathtools as PT
 from ezdxf import zoom
 from ezdxf.colors import RGB, aci2rgb
 from ezdxf.math import Vec2
@@ -69,8 +70,10 @@ from build123d.topology import (
 )
 from build123d.build_common import UNITS_PER_METER
 
-PathSegment: TypeAlias = PT.Line | PT.Arc | PT.QuadraticBezier | PT.CubicBezier
-"""A type alias for the various path segment types in the svgpathtools library."""
+if TYPE_CHECKING:
+    import svgpathtools as PT
+    PathSegment: TypeAlias = PT.Line | PT.Arc | PT.QuadraticBezier | PT.CubicBezier
+    """A type alias for the various path segment types in the svgpathtools library."""
 
 # ---------------------------------------------------------------------------
 # ---------------------------------------------------------------------------
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -67,8 +67,6 @@ from OCP.XCAFDoc import (
     XCAFDoc_ColorSurf,
     XCAFDoc_DocumentTool,
 )
-from ocpsvg import ColorAndLabel, import_svg_document
-from svgpathtools import svg2paths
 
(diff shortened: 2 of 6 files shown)


gumyr:
I'm not convinced about this one.  Users can skip reloading all of the build123d (and ocp_vscode) content by following the best practices shown here: https://github.com/bernhard-42/vscode-ocp-cad-viewer?tab=readme-ov-file#best-practices (yes, this should be in the build123d documentation too).


_Use the Jupyter extension for a more interactive experience. This allows to have one cell (separated by # %%) at the beginning to import all libraries_
```py
# %%
from build123d import *
from ocp_vscode import *

# %%
b = Box(1,2,3)
show(b)
# %%
```

In addition, there is already a proposal to optimize the IPython content see: https://github.com/gumyr/build123d/issues/924

jsmnbom:
My use case is to have helper code in a separate python package. This helper code needs to import build123d objects.

I thought the only way to refresh the help code in jupyter was to restart the kernel, since [IPython extension autoload](https://ipython.org/ipython-doc/3/config/extensions/autoreload.html) didn't seem to work - after adddional testing, autoreload does seem to work, but it just behaves weird in quite a few circumstances. All in all I still believe that it makes sense to try to make the library load faster and lazy load heavier dependencies when they are required, but if you do not agree, feel free to close the PR :)

gumyr:
Getting faster loading times is clearly a good thing. Let me think about it a bit...

jdegenstein:
I haven't had time to fully digest, but PEP about lazy imports is here https://peps.python.org/pep-0690/

I personally prefer that lazy imports are not adopted at this time, but I would like to see some more performance profiling on different platforms to understand the impact this could have. Perhaps the profiling could be done on CI for the same platforms we run tests against?

jsmnbom:
Do note that that PEP does not actually apply to what this PR does, and was eventually rejected. It only specifies a mechanism for handling how libraries can implement lazy submodules. Whereas all this code change does, is delay the import of a module until it is actually needed.
In terms of how much this would speed up imports I did a quick test:
Before removing scipy as dependency (incl svgpathtools which i missed in this PR)
```sh
> uv run python -m timeit -n 1 -r 1 "import build123d" 
1 loop, best of 1: 1.14 sec per loop
```
after:
```sh
>uv run python -m timeit -n 1 -r 1 "import build123d" 
1 loop, best of 1: 832 msec per loop
```
---
After reading that PEP I do think build123d should adopt some of what it describes. I see quite a lot of benefit in splitting export and import of different filetypes into submodules. This could even allow users to maybe install the library without most of the dependencies it currently has. Much in a similar way to how VTK is currently somewhat optional (if you can figure out how to compile OCP without).

Supporting lazy imports would also help with #924 and even #946, cutting down the amount of mandatory dependencies to `anytree, cadquery-ocp, numpy, trianglesolver, typing-extensions` I think.

Other libraries seem to support this using https://peps.python.org/pep-0562/ .

I propose that the exporters and importers should get migrated to a folder structure like this:
image
Where `__init__.py` uses __getattr__ and __dir__ to only import the other files when necessary.
At the same time I think it would make a lot of sense to refactor some of the exporting and importing code, so that it is more universal across different file types. (this is what the base_*.py files do), and then inside `__init__.py` can still keep simple user facing and always availible in __all__ and therefore using `from build123d import *` functions for each format:
```python
def export_step(...):
     from build123d.exporters import ExportStep
    exporter = ExportStep(unit, write_pcurves, precision_mode)
    exporter.add_shape(to_export)
    exporter.write(file_path)
```

Hope this all makes sense - I got a bit rambly :/

jsmnbom:
I'm gonna go ahead and close this PR, as it seems svgpathtools also loads scipy, and that svgpathtools is used a ton inside the exporter for svg. Meaning that it's unfortunately not quite this simple to avoid loading scipy. Instead I will submit a draft PR of my suggestion above soonish :)

Edit: I'm gonna hold off on submitting the draft PR. Started writing it and it is overall looking pretty good in my opinon, but it's also a lot of work, and I don't really see a need to do it, unless y'all think it's a good idea.

jsmnbom:
Found out how to properly profile this, so I'm gonna open this back up. Sorry for all the back and forth.
Using `python -X importime -c "import build123d"` and https://kmichel.github.io/python-importtime-graph/ here are two graphs showing 2173ms import time vs. 1010ms import time.

import_time_current
import_time_lazy

I definitely still think there is value in doing this change.

gumyr:
Thank you for all the work done here profiling build123d import times but we're not going to go ahead with this lazy import solution. In the future it might be worthwhile to do a larger refactor that separates some of the expensive imports though.

PR #947, opened 2025-03-23
Add Tangent objects for Point and Arc

jwagenet:
Adding a batch of tangent objects to create tangent arcs and lines between arcs and arcs and points. Please scrutinize my style and the tests I added (which are fairly extensive).

### New Objects
`PointArcTangentLine`
`PointArcTangentArc`
`ArcArcTangnetLine`
`ArcArcTangentArc`

image

### Shortcomings
* The arc objects should be able to find tangents when the start point/arc is overlapping or inside, but I neglected to consider these cases until recently, so they are NotImplemented for now
* I suspect py solvespace could probably replace the internals of these objects, but that is out of scope for now.

## Open questions
* Is there a preference for whether a tangent object should be found anywhere on a complete arc or only if the tangent is coincident with the supplied, trimmed arc(s)?
* Although they are 1d objects, I wonder if these objects should be defined/tested/documented separately for organization since they work differently than other 1d objects. This would include DoubleTangentArc and other tangent objects Id like to make based on lines.
* Any changes to naming

### Side vs Keep
* As incompletely discussed on discord, I am using `Side.LEFT/RIGHT` to pick starting side for all objects, and using `Keep.INSIDE/OUTSIDE` for the ArcArc objects. 
* Using `Side` is a departure from `Keep.TOP/BOTTOM' in DoubleTangentArc, so I am open to discussing alternatives.
* Extending `ArcArcTangentArc` for overlapping arcs would increase the possible tangents to 8 so some solution will be necessary

### Sympy
* I am using sympy to find intersections due to OCP intersection methods not working well as implemented.
* Only `ArcArcTangentArc` uses sympy as committed. Finds circle intersections.
* I ditched sympy and increased the minimize max_size on `PointArcTangentArc` (based on `DoubleTangentArc`). For some reason sympy is 4 - 10 times slower than using the minimizer, even if the minimizer fails at a larger size. Would use circle and line intersections
* Something I hadn't considered until writing this up is to precalculate intersections symbolically and substitute at runtime. Based on light testing this would be about 10x faster, but to be determined how to store symbolic solutions.
* I do want to be cognizant about adding a new dependency in light of #942.


jwagenet:
Merging with dev + mypy/pyline/black still needs to be done.

gumyr:
At first glance these look great but it will be a few days before I have the time to review them in detail.

jwagenet:
No problem (I think) this is a complicated addition and I'd want to make sure it fits with the project goals.

gumyr:
This looks like some excellent work and welcome additions to build123d. I don't really have any substantial input to what you've done here other than to fix up the type checking and code coverage failures.

As for naming and design choices like Side vs. Keep how about we get some feedback from the community on this after it has been merged?  I don't think we have enough content to do a release for a little while so we have some time to explore what is best - even if that is changing `DoubleTangentArc` to align with the new direction.

Thanks again!

jwagenet:
I've marked this for review, but there are codecov errors I don't understand (at least the ones highlighting comments).

gumyr:
The code coverage errors don't make sense; there is at least one comment that doesn't have coverage.  How about we merge this code and see what's what?

jwagenet:
Sure, that's fine with me.

Issue #886, opened 2025-01-29
Add a section to the docs on `sort_by`/`filter_by`/`group_by` shape property


gumyr:
As a replacement for the `SortBy` functionality of the `ShapeList`, one can use class properties, like:
```py
shape.edges().filter_by(Edge.is_interior)
```
or
```py
from itertools import product

from build123d import Box, ShapeList, Solid, Vertex
from ocp_vscode import ColorMap, show

boxes = ShapeList(
    Box(1, 1, 1).scale(0.75 if (i, j) == (1, 2) else 0.25).translate((i, j, 0))
    for i, j in product(range(-5, 6), repeat=2)
)

# sort by distance to origin
boxes = boxes.sort_by(Vertex().distance)
show(*boxes, colors=ColorMap.listed(len(boxes)))
```
and
```py
# sort by distance to largest box
boxes = boxes.sort_by(boxes.sort_by(Solid.volume).last.distance)
show(*boxes, colors=ColorMap.listed(len(boxes)))
```

![Image](https://github.com/user-attachments/assets/1232dc57-8368-4940-ab16-523b8857cb26)

jwagenet:
Do you have more essential examples of the shapelist operators in action? Especially callables since `sort_by(Vertex().distance)` is replaced by sort_by_distance(Vertex()).

gumyr:
How about these?
```py
.filter_by(lambda f: len(f.inner_wires()) >= 3) # 3 or more holes
.filter_by(lambda w: all(e.geom_type == GeomType.CIRCLE for e in w.edges())) # All the edges in the wire are circular
.filter_by(lambda w: len(set(e.radius for e in w.edges())) == 1) # wires with a radius of 1
.filter_by(lambda e: cyl_length * 0.4 < e.length < cyl_length * 0.9) # edges of a given length range
.filter_by(lambda f: MotorClass.spline_c1_continuity(f)) # custom methods
.filter_by(lambda f: any(e.geom_type == GeomType.LINE for e in f.edges())) # contains a straight line
.filter_by(lambda f: cylinder_dir.dot(f.bounding_box().size) > 2 * f.radius) # cylinder is longer than diameter
.filter_by(lambda f: len(f.outer_wire().edges().filter_by(GeomType.LINE)) >= 8) # nested filter_by
```

Are these the types of things you're looking for?

jwagenet:
[...]

Issue #711, opened 2024-09-24
Example for using a lambda inside `ShapeList.filter_by`


CarpeNecopinum:
Question that came up in the Discord server was how to filter faces by being parallel to a given axis, as there is no built-in filter for it:

```py
part.faces().filter_by(lambda f: abs(f.normal_at(None).dot(axis)) < 1e-6)
```

Jern noted that there are no examples showing this feature (lambda as filter), so we should probably add one / a few.

jdegenstein:
Thank you! Would be good to add this to page https://build123d.readthedocs.io/en/latest/key_concepts.html#custom-sorting-and-filtering

gumyr:
```py
shape = Box(1, 1, 1)
f_parallel = shape.faces().filter_by(lambda f: abs(f.normal_at().dot(Axis.X.direction)) < 1e-6)
f_perpendicular = shape.faces().filter_by(Axis.X)
```
Might want to clarify that the `dot` method expects a Vector and not an Axis.

jwagenet:
This can be closed with the implementation of #959

Issue #675, opened 2024-08-24
Improved documentation on Select ALL, LAST, NEW


gumyr:
The docs need to be improved here: https://build123d.readthedocs.io/en/latest/advantages.html#last-operation-objects

jwagenet:
This can be closed with the implementation of #959.

Issue #455, opened 2024-01-04
edges(Select.NEW) fails for chamfer


MatthiasJ1:
```py
with BuildPart() as p:
    Box(10,10,10)
    chamfer(edges(), 1)
    e = edges(Select.NEW)
    assert e
```
```
AssertionError
```

`topology.new_edges` "sequence of shapes" needs to account for `Mode`.

gumyr:
Maybe this is more of a documentation/expectations problem.  `Select.NEW` is designed to find new edges that were not in either of the Shapes being combined as follows:
```py
with BuildPart() as new_ex:
    Box(10, 10, 10)
    Cylinder(2, 20)
    n = edges(Select.NEW)
```
![image](https://github.com/gumyr/build123d/assets/85638518/9ba4b5d4-ca0e-4349-9253-29342b83f475)
so that these new edges could be easily worked with (for example, filleted).  `Select.LAST` will select all of the edges changed by the last operation, as shown here for the above example:
![image](https://github.com/gumyr/build123d/assets/85638518/8c3bc593-42be-4c31-b1e8-ee14da26e4ce)

Does this address the issue?

MatthiasJ1:
I might have missed it but I could't find documentation for `Select.NEW`. For the example below:

```py
with BuildPart() as part:
    Box(10, 10, 10)
    chamfer(edges()>>Axis.Z<<Axis.Y, 5)
    new = edges(select=Select.NEW)
    last = edges(select=Select.LAST)
```
I get the following. The edges under "new expected" did not exist in the object prior to the chamfer operation (although you could argue one of the |X edges was the original edge moved).
<img width="579" alt="Screenshot" src="https://github.com/gumyr/build123d/assets/149267773/3234f159-9536-4ddd-a703-a2171824864b">

jwagenet:
This can be closed with the implementation of #959 as a documentation solution.

However, I agree `Select.NEW` is confusing with fillet and chamfer (possibly other operations) and an exception for those ops to return "new expected" as in @MatthiasJ1's example would not be unwelcome.

Those edges can be found with `new_edges` (which is what `Select.NEW` uses under the hood):
````python
with BuildPart() as part:
    box = Box(10, 10, 10)
    chamfer(edges() >> Axis.Z << Axis.Y, 5)
    new = new_edges(box, combined=part)
````

PR #959, opened 2025-04-07
Add new docs section "Selectors and Operators"

jwagenet:
### Overview
- [Demo](https://jwagenet.github.io/build123d/html/selectors_operators.html)
- Expand ideas and concepts from key concepts section
- Provide minimal usage examples in front matter
- Add structure for examples of low to medium complexity for different criteria types and filtering methods

Intended to resolve:
- #886
- #711
- #675
- #455

### Notes
- I've generally used a conversational, tutorial tone, but could switch to something more succinct
- In general there are multiple ways to arrive at a selection. As such, I acknowledge some of the examples may not offer the best method to accomplish a selection, but the intent is to show an option. However, please let me know if any of the examples are bad or misleading
- Some of the examples certainly lack imagination, but hopefully they lay a foundation for improvements
- @gumyr offered some great examples in #886, and I adapted a few I had decent ideas for
- This does not (yet) edit or replace `selectors.rst` from "Key Concepts"


gumyr:
Very nice - thank you! Here are some things to think about:

- I'm not even sure where the term "Selectors" come from but I've used it many times in the documentation.  What do you think of changing the name of this section to "Topological Selection and Exploration"? It would provide a natural link to the https://build123d.readthedocs.io/en/latest/key_concepts.html#topology section (either way there should be a link to the topology section of the docs near the top of this section). 
- The `Selector` section doesn't mention that the `Select` stuff is a parameter of the `edges`, `faces` etc. methods (although that is shown later).
- A short statement in the `Selector` section that describes these methods as finding / or not OCCT "reused" shapes might be helpful. Maybe something like: In general, OCCT will attempt to "reuse" shapes which we can take advantage of to find newly created shapes.
- There is a `new_edges` function which can be used in Algebra mode to emulate the functionality of the `Builder`/`Select`. build123d functions are not documented well at all.
- I wonder if we could build some type of "cheat sheet" for selectors? If we could dynamically extract Properties of the base classes we could have a nice reference either in the existing "cheat sheet" section of here (or both).
- In the `Group` section it might be helpful to state that `GroupBy` is just a list of ShapeList which can be printed out if that is helpful.  Sometimes it's helpful just to see how many things are in each sub-group to identify problems.

Would you like me to merge this section as is and make any changes afterwards (there is a lot of great content here that users will benefit from with no changes)?

jwagenet:
If you want to merge now I won't complain! However, I think I could implement most of the suggestions quickly (a few days) and roll the rest into an issue. My thoughts:

* "Topological Selection and Exploration" is great. I think "Feature Selection.." is more approachable, but per your link "topology" is already in the lexicon, so to speak. I can include the link.
    * Are you happy to keep using the titles/descriptions "Selectors" and "Operators" for now? Operators could change to ShapeList operators/methods for more specificity.
*  Mainly excluded from the initial description to avoid confusion between builder selector use and shape selector use. (I hadn't noticed `Select` only worked in builders until I worked on this!)
* I tried to mention the edge reuse with `Select.NEW`, again dont want to confuse in the overview. I'm averse to digging to deep on OCCT behavior to keep the scope tight here. Open to your insight.
* `new_edges` seems like a no brainer to add. I'll look into it.
* An extended cheat sheet sounds great, but I think that will be out of scope for this first round. The properties specifically are a driver for #935.
* Easy to add a note about printing.

gumyr:
It's probably easier for you if I let you make the changes so I'll hold off merging.  Looking forward to improved docs!

jwagenet:
I've pushed the updates, ready to merge.

gumyr:
Thank you!  This is a very welcome addition to the docs.

Issue #778, opened 2024-11-11
Adding EXTRABOLD to fon


startOfStars:
Hello

At the moment EXTRABOLD seems not be supported in fontStyle ?
I find REGULAR, BOLD, ITALIC but not EXTRABOLD.

Thank you !

gumyr:
`EXTRABOLD` is not supported by the OpenCascade CAD kernel that is used by build123d.  However, there are some unused values that could be added - these are the OCCT options:

- Font_FontAspect_UNDEFINED
- Font_FontAspect_Regular
- Font_FontAspect_Bold
- Font_FontAspect_Italic
- Font_FontAspect_BoldItalic
- Font_FA_Undefined
- Font_FA_Regular
- Font_FA_Bold
- Font_FA_Italic
- Font_FA_BoldItalic

while these are the options exposed by build123d:

- REGULAR
- BOLD
- ITALIC
 
At this point it isn't clear what the difference between the `FontAspect` and `FA` versions are but it seems like `BoldItalic` would be good to add.

startOfStars:
Thank you for your answer.
To be honest I don't understand all of what you wrote... 

I was asking this because for exemple for Noto Sans Google Font, 18 styles are available, including "ExtraBold".

PS : Thank you for this amazing build123d !

gumyr:
The fonts are provided by the CAD kernel not Google or anywhere else so we're restricted to what OCCT provides.

jwagenet:
@startOfStars A little late, but you can directly call the font with the `font_path` kwarg eg. `Text("Hello World", 10, font_path="C:/Windows/Fonts/NotoSans-ExtraBold.ttf")` on Windows. This seems to override both `font` and `font_style`. The docs don't really indicate this behavior.

jwagenet:
> At this point it isn't clear what the difference between the FontAspect and FA versions are

It seems like the FA versions are just aliases for the FontAspect versions

Issue #459, opened 2024-01-05
Text does not factor in line height


MatthiasJ1:
The text alignment bounding box height should be determined by the font's line height rather than the rendered text's bounding box. The current implementation leads to errors like this:

```py
Text("A car.", 5)
```
<img width="316" alt="Screenshot" src="https://github.com/gumyr/build123d/assets/149267773/733710ce-5070-40b3-892c-808781d7d3f0">

```py
Text("A", 5) + Pos(6)*Text("car", 5) + Pos(10)*Text(".", 5)
```
<img width="312" alt="Screenshot2" src="https://github.com/gumyr/build123d/assets/149267773/ac3ec58c-0dda-4c28-9bc4-c6e7f538ffc7">

gumyr:
Align applies to Text like all objects:
```py
t = (
    Text("A", 5, align=Align.MIN)
    + Pos(5) * Text("car", 5, align=Align.MIN)
    + Pos(12) * Text(".", 5, align=Align.MIN)
)
show(t)
```
![image](https://github.com/gumyr/build123d/assets/85638518/a794993d-af41-4da9-8028-f818fd99a533)

The default is `Align.CENTER,Align.CENTER`.  Do you still think there is a problem here?

MatthiasJ1:
```py
t = (
    Text("ye", 5, align=Align.MIN)
    + Pos(5.5) * Text("s", 5, align=Align.MIN)
)
show(t)
```

<img width="364" alt="Screenshot" src="https://github.com/gumyr/build123d/assets/149267773/c27c0612-73c4-4f7e-9b01-3a7410728cb6">

gumyr:
If alignment worked based on the font's line height, vertically centering a single lower case letter would always be wrong which would be very frustrating.  build123d isn't a typesetting tool. Users are primarily going to be embossing very small amounts of text onto their parts so control over single characters is more important then sentences spanning multiple `Text` blocks. 

Won't fix.

MatthiasJ1:
@gumyr if you want to center text based on its bounding box rather than the text parameters you could easily realign the object using its bounding box (`add` could have an align parameter) or you could have an `alignmode=[Text, Geometry]` argument for `Text`.

Where this is an issue is say you have a ruler or panel with text underneath some feature. If the text includes different anchors ("mode" vs "43" vs "yaw") there is currently no way to align those without manually specifying offsets for each label.

gumyr:
[...]

Issue #458, opened 2024-01-05
Text(align) does not affect multiline text 


MatthiasJ1:
```py
Text("center\nalso center", 5, align=(CENTER, CENTER))
```
gives
<img width="316" alt="Screenshot" src="https://github.com/gumyr/build123d/assets/149267773/b2bb0f0f-c9cb-4f96-8877-acd9051f51aa">

`Font_BRepTextBuilder::Perform` is missing `Graphic3d_HorizontalTextAlignment` and `Graphic3d_VerticalTextAlignment`. These should probably be a separate argument: `Text(..., align=..., text_align=...)`

PR #960, opened 2025-04-11
Add OCCT text alignment to `Text`/`make_text` 

jwagenet:
`Text` now has a `text_align` tuple arg to control how text is aligned within the bounding box by implementing unused `Graphic3d_` enums. New enum `TextAlign` to control the text alignment as a tuple pair in `text_align`. `align` still controls how the bounding box, but is now default `None`.

**I'm looking for feedback on the naming and implementation of `TextAlign` and the `text_align` arg.**
I'm open to alternate ways of implementing text alignment: 
* Splitting `TextAlign` into two enums, eg `HorizTextAlign`
* Assign separate args
* Implement `TextAlign` differently `text_align` takes a single value, eg `LEFTBOTTOM`, `RIGHTCENTER`
* etc.

### Changes
- Add `FontStyle.BOLDITALIC` font aspect
- Add`TextAlign` enum
- Add `text_align` tuple arg to `Text` and` make_text` which resolves to `(horiz_align, vert_align)`. Maps to `Graphic3d_HTA_*` and `Graphic3d_HTA_*` enums for `Font_FontMgr.Perform()`
- Use `text_align=(CENTER, CENTER)` as default, although I think (LEFT, BOTTOM) is more traditional
- Set `align=None` as default to align by text alignment by default. align still aligns the bounding box
- Minimal test coverage for addition of `text_align`

### Resolves
* #778
* #458 
* #459

### Examples
TextAlign: LEFT, CENTER, RIGHT
image

TextAlign: BOTTOM, CENTER, TOP, TOPFIRSTLINE
image

Align: MAX, CENTER, MIN
image


jwagenet:
~~The TextAlign enum needs to be added to the cheat sheet and I should probably tuplify `text_align`.~~

Also reverted fuse commit as the change broke a lot of tests. Saving for later.

jwagenet:
OCCT has `Font_TextFormatter`, which can do a bit more formatting like setting bounding box size and text wrapping, so newline isn't necessary for multiline text. Could be useful in the future.

gumyr:
I didn't expect that multi-line text would be needed as build123d isn't a text processing tool but it appears that there is a need for it. Given that, I don't see any way to support it without both Align and TextAlign capabilities. Thanks for looking into this.

jwagenet:
I agree. Significant text processing support doesn't seem super necessary (and there is more font/textformatting to be exposed from OCCT), but this seemed like a quick fix for open issues.

If the `TextAlign` is satisfactory as implemented, I'm calling this PR done.

gumyr:
Thanks!  Some of the documentation for this could be written with the feature itself :)

PR #964, opened 2025-04-17
Fix `revolve` direction and size with negative `revolution_arc`

jwagenet:
Fix modulo logic for converting negative `revolution_arc` so  `angle == revolution_arc`. I assume this was an oversight as the modulo result will match sign of the modulo divisor. Previously `revolution_arc = -190` corresponded to `angle = 170`.

- Add sign for modulo divisor
- Keep sign for consistency with `angle == 360` case even though solid is the same
- Add new test for range of `revolution_arc` values

### Example

````python
with BuildPart() as part:
    with BuildSketch() as sketch:
        RegularPolygon(10, 5, align=(Align.CENTER, Align.MIN))
    revolve(axis=Axis.X, revolution_arc=-90)
````

Now
image

Previously
image

### DIFF
--- a/src/build123d/operations_part.py
+++ b/src/build123d/operations_part.py
@@ -464,8 +464,9 @@ def revolve(
 
     # Make sure we account for users specifying angles larger than 360 degrees, and
     # for OCCT not assuming that a 0 degree revolve means a 360 degree revolve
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #984, opened 2025-05-06
Add timestamp argument to STEP export

jgraichen:
Allow passing a timestamp value when exporting STEP files, to generate STEP files with a specific timestamp value in the file header.

For example, a null timestamp (`0000-00-00T00:00:00`), or a static timestamp can be used when generated files should be equal if there are no visual changes, such as for file versioning.

This commit extends the `#export_step` function, to accept a `timestamp` keyword argument, that can be a string or a `datetime` object. Accepting a `datetime` object makes it easier to use from Python, like a source file modification date.

---

In [`jgraichen/someline`](https://github.com/jgraichen/someline/blob/3058617e9de72d2eff38fc3e1c9779d8c4b350ae/someline/util.py#L167-L178) I store the exported step files in the Git repository, but the current timestamp in the file always results in changes. I use them there in a CI pipeline to ensure that the code always generates the same models as before.

Currently, I have to strip the timestamp after the export. Passing a `timestamp` directly when exporting would make it much easier and directly allow reproducible builds.

Some formatting changes are from `black`, some imports were not used anymore.

### DIFF
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -29,10 +29,10 @@ license:
 # pylint has trouble with the OCP imports
 # pylint: disable=no-name-in-module, import-error
 
-from io import BytesIO
+from datetime import datetime
 import warnings
+from io import BytesIO
 from os import PathLike, fsdecode, fspath
-from typing import Union
 
 import OCP.TopAbs as ta
 from anytree import PreOrderIter
@@ -47,7 +47,11 @@ from OCP.RWGltf import RWGltf_CafWriter
 from OCP.STEPCAFControl import STEPCAFControl_Controller, STEPCAFControl_Writer
 from OCP.STEPControl import STEPControl_Controller, STEPControl_StepModelType
 from OCP.StlAPI import StlAPI_Writer
-from OCP.TCollection import TCollection_AsciiString, TCollection_ExtendedString, TCollection_HAsciiString
+from OCP.TCollection import (
+    TCollection_AsciiString,
+    TCollection_ExtendedString,
+    TCollection_HAsciiString,
+)
 from OCP.TColStd import TColStd_IndexedDataMapOfStringString
 from OCP.TDataStd import TDataStd_Name
 from OCP.TDF import TDF_Label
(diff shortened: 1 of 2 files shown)


src/build123d/exporters3d.py
```diff
@@ -302,6 +308,11 @@ def export_step(
     header = APIHeaderSection_MakeHeader(writer.Writer().Model())
     if to_export.label:
         header.SetName(TCollection_HAsciiString(to_export.label))
+    if timestamp:
```
gumyr: if timestamp is not None:

gumyr (changes requested):
Thanks, this looks like a good feature.  It can be irritating when things like timestamps make files artificially different.

src/build123d/exporters3d.py
jgraichen: Done. 👍

tests/test_exporters3d.py
```diff
@@ -144,6 +146,29 @@ def test_export_step_unknown(self):
         os.chmod("box_read_only.step", 0o777)  # Make the file read/write
         os.remove("box_read_only.step")
 
+    def test_export_step_timestamp_datetime(self):
+        b = Box(1, 1, 1)
+        t = datetime(2025, 5, 6, 21, 30, 25, tzinfo=ZoneInfo("Europe/Amsterdam"))
```
jdegenstein: I don't think we need to test the datetime module itself, can we just test by `t = datetime(2025, 5, 6, 21, 30, 25)` without specifying `tzinfo=` and thus eliminating the platform specific dependency and related changes in `pyproject.toml`?

pyproject.toml
```diff
@@ -70,6 +70,7 @@ development = [
     "pytest-benchmark",
     "pytest-cov",
     "pytest-xdist",
+    "tzdata; sys_platform == 'win32'",
```
jdegenstein: I would like to avoid having platform specific dependencies, please see my earlier comment.

tests/test_exporters3d.py
jgraichen: Done. :+1:

pyproject.toml
jgraichen: Done. :+1:

gumyr (commented):
+1

jdegenstein:
Merged. Thanks for your contribution!

jgraichen:
Thank you both, for your continuous development and maintenance! It has been a pleasure.

PR #988, opened 2025-05-13
deglob.py -> add ability to write deglobbed change back to target file

jdegenstein:
This PR adds:
```
python deglob.py --write target_file.py
```
Which will replace the first occurrence of a glob import in the target file and warn if more than one instance was found. The default behavior (without `--write`) is still to print to stdout and open the file as read-only.


tools/deglob.py
```diff
@@ -16,8 +16,9 @@
     an import statement listing only those names. This practice can help
     prevent polluting the global namespace and improve clarity.
 
-    Example:
-        deglob.py my_build123d_script.py
+    Examples:
+        python deglob.py my_build123d_script.py
```
gumyr, line 20: The `python deglob` isn't required as `#!/usr/bin/env python` is at the first line of the file and this should cover the user's python version. Is this not working in your environment?

tools/deglob.py
```diff
@@ -16,8 +16,9 @@
...
+        python deglob.py -h
```
gumyr, line 21: --write` should probably be in the examples

tools/deglob.py
```diff
@@ -152,15 +173,41 @@ def main():
...
     # Create the import statement to replace the glob import
     import_line = f"from build123d import {', '.join(actual_imports)}"
-    print(import_line)
+
+    if args.write:
+        # Replace only the first instance, warn if more are found
+        updated_code = re.sub(r"from build123d import\s*\*", import_line, code, count=1)
+
+        # Write code back to target file
+        with open(args.build123d_file, "w", encoding="utf-8") as f:
```
gumyr: Probably should handle the case of a non-writable file more gracefully:
```py
try:
    with open(args.build123d_file, "w", encoding="utf-8") as f:
        f.write(updated_code)
except (PermissionError, OSError) as e:
    print(f"Error: Unable to write to file '{args.build123d_file}'. {e}")
    sys.exit(1)
```

tools/deglob.py
```diff
@@ -152,15 +173,41 @@ def main():
...
+            f.write(updated_code)
+
+        if glob_count:
```
gumyr: Maybe there should be a `--verbose` flag controlling this output?  I'm assuming the primary use case for the `--write` option is automated scripts where verbose output like this might be unwelcome.

tools/deglob.py
```diff
@@ -152,15 +173,41 @@ def main():
...
+            print(f"Replaced build123d glob import with '{import_line}'")
+
+        if glob_count > 1:
```
gumyr, line 227: I'm not sure it's a good idea to update the file when the script will not have solved the problem of glob imports.  It seems as though we could:
1) warn and abort the script once two glob imports are found
2) change one of the glob imports and remove any others
3) change one and warn about the others (behavior of this PR)

1 is safe but the provides the least value. 2 is more dangerous (?) but fixes the problem. 3) is safer but doesn't fix the problem.  I think 2 could be implemented with:
```py
# Replace all instances of glob imports
updated_code = re.sub(r"from build123d import\s*\*", import_line, code)

# Remove any remaining glob imports
while re.search(r"from build123d import\s*\*", updated_code):
    updated_code = re.sub(r"from build123d import\s*\*", "", updated_code)
```
What do you think?

gumyr (changes requested):
I didn't expect this script to be used by anyone but me :)

tools/deglob.py
jdegenstein: I don't think the shebang syntax is supported on Windows, I chose to keep the examples more specific but leave the shebang for more convenient use on platforms that support it (did not seem to cause issues on Windows when I tested).

tools/deglob.py
jdegenstein: Sure, perhaps it would be even better to have the entire argparse `-h` output in this region?

tools/deglob.py
jdegenstein: This is a great point, will add it

tools/deglob.py
jdegenstein: I agree

tools/deglob.py
jdegenstein: I don't like 2 in context of the verbose flag being false. I am also concerned with 2 if there were imports inside of functions (and not at the top level). Option 4 is to keep the behavior of (3) but always print the warning (even when verbose is set to false).

If not then I prefer 1.

tools/deglob.py
gumyr: That's a better idea.  The -h output is actually quite good (without your changes):
```
usage: deglob.py [-h] build123d_file

Find all the build123d symbols in module.

positional arguments:
  build123d_file  Path to the build123d file

options:
  -h, --help      show this help message and exit
```

tools/deglob.py
gumyr: I'm fine with 1 or 4 whatever you think is best.

tools/deglob.py
gumyr: Maybe having two example then?  The shebang isn't really a big deal though, users can always explicitly reference python as you've shown.

jdegenstein: Chose to leave off shebang example per discussion below

tools/deglob.py
jdegenstein: Added usage section from argparse -h output, as discussed.

tools/deglob.py
jdegenstein: Chose option 4 as described above.

tools/deglob.py
jdegenstein: Added this, as requested

tools/deglob.py
jdegenstein: Done.

jdegenstein:
Merged, thanks for review and sorry for lengthy delay

PR #992, opened 2025-05-16
Rework Location overloads, eliminate RotationLike

### LINKED ISSUE #441: Review the Location constructor and optimize the valid overloads
Some of the Location constructor's overloads seem confusing and may not be optimal.  Specifically,
```py
    @overload
    def __init__(
        self, translation: VectorLike, direction: VectorLike, angle: float
    ):
```
seems misleading at best.  The term `translation` and `direction` should also be reconsidered as there are `position` and `orientation` properties.

jdegenstein:
This is related to PR #991 but includes changes that allow this to pass mypy and pytest. There might be a few changes needed to make sure code coverage as at 100% so I am opening as draft for now.

This PR also eliminates `RotationLike` and replaces with `Rotation | VectorLike` across the codebase.

EDIT: here is the summary of location overloads that I ended up with:
```py
class Location:
    "@overloads" #omitting future @overload for brevity
    def __init__(self, position: VectorLike | None = None, orientation: VectorLike | None = None, ordering: Extrinsic | Intrinsic | None = None):
    def __init__(self, plane: Plane):
    def __init__(self, location: Location | Rotation):
    def __init__(self, top_loc: TopLoc_Location):
    def __init__(self, gp_trsf: gp_Trsf):

    #actual constructor \/
    def __init__(self, *args, **kwargs):
        ...
```
What changed from the discussion at issue #991 is that the first overload now only accepts `orientation: VectorLike | None`. This clarifies the behavior significantly because it means that if one passes `Location(Rotation())` it will be handled by the later revised `__init__` overload (`location: Location | Rotation`).


src/build123d/geometry.py
```diff
@@ -126,6 +126,51 @@ def _parse_intersect_args(*args, **kwargs):
     return axis, plane, vector, location, shape
 
 
+def _parse_location_args(*args, **kwargs):
+    position, orientation, ordering, plane, location, top_loc, gp_trsf = (None,) * 7
```
gumyr, line 130: This line isn't required as the `pop` method will either assign the `kwargs["position"]` or if there is no "position" key, `None`.

src/build123d/geometry.py
```diff
@@ -1376,149 +1421,80 @@ class Location:
...
     def __init__(
         self,
-        translation: VectorLike,
-        rotation: RotationLike,
-        ordering: Extrinsic | Intrinsic,
-    ):  # pragma: no cover
-        """Location with translation with respect to the original location.
-        If rotation is not None then the location includes the rotation (see also Rotation class)
-        ordering defaults to Intrinsic.XYZ, but can also be set to Extrinsic
+        position: VectorLike | None = None,
```
gumyr, line 1427: Why `None` here?  What should happen if the user passed `None` as all three parameters?

src/build123d/geometry.py
```diff
@@ -1376,149 +1421,80 @@ class Location:
...
+    ):
+        """Location with position that defaults to the global origin.
+        If orientation is not None then the location includes the orientation
+        (see also Rotation class). Ordering defaults to Intrinsic.XYZ,
+        but can also be set to members of the Extrinsic enum.
         """
 
     @overload
-    def __init__(self, plane: Plane):  # pragma: no cover
+    def __init__(self, plane: Plane):
```
gumyr, line 1438: The `# pragma: no cover` was to instruct the coverage calculator to ignore these lines as even if they were covered in the tests they are reported as not covered.  Has this changed now?

src/build123d/geometry.py
```diff
@@ -1376,149 +1421,80 @@ class Location:
...
+            quaternion = gp_Quaternion()
+            quaternion.SetEulerAngles(rot_order, *o_radians)
+            transform.SetRotation(quaternion)
+        if position:
+            # translation part must be done after SetRotation
+            transform.SetTranslationPart(Vector(position).wrapped)
 
-            transform.SetTranslationPart(Vector(translation).wrapped)
-        self.wrapped = TopLoc_Location(transform)
+        if hasattr(self, "wrapped"):
```
gumyr, line 1494: I don't think having a return in the constructor is recommend. How about just using a `_wrapped` or similar variable in the processing and check `if _wrapped not None:` assign `self.wrapped = _wrapped` else use the `transform`?

gumyr (changes requested):
I'm having doubts about removing the `angle` variant - this would be a loss in backwards compatibility for something that does work and is used in our code and might be used in user code. Would you be okay will leaving it?

All of the other changes are `RotationLike` aren't they?  I'm going to think about this some more before reviewing those changes.

src/build123d/geometry.py
jdegenstein: Code coverage appears to be fine for these lines, so I left them off. (I still have to add 2 more tests to get to 100% coverage, unrelated to this).

src/build123d/geometry.py
jdegenstein: Yeah great point, what I did here was based on what init did earlier. I agree with your idea.

src/build123d/geometry.py
jdegenstein: Good point, I forgot about the functionality of the dict pop default.

src/build123d/geometry.py
jdegenstein: I was thinking this captures the empty init case as well, `Location()` -> `Location((0,0,0),(0,0,0))`. I will look at the other point you raised

jdegenstein:
> I'm having doubts about removing the angle variant - this would be a loss in backwards compatibility for something that does work and is used in our code and might be used in user code. Would you be okay will leaving it?

@gumyr I am ok with having the functionality kept, I would possibly recommend that the signature would be different and accessible via kwargs only (which greatly simplifies the input parsing, reduces the likelihood of bugs, and makes the functionality a bit clearer to the user):
```py
    def __init__(self, position: VectorLike, *, direction: VectorLike, angle: float): 
    #                                        ^note the star, which enforces kwarg-only for the later parameters
```
Then calling this version would look like `Location((1,2,3),direction=(0,1,1),angle=45)`

The downside of this kwarg-only approach is of course this is a breaking change. I feel that it is probably worth it because currently on dev the distinction between `Location(pos1a, pos2a)` and `Location(pos1b, pos2b, pos3b)` is not that clear i.e.:
```
pos2a: VectorLike # is a rotation
pos2b: VectorLike # is a direction
```
see also: https://peps.python.org/pep-0570/#how-to-teach-this

> All of the other changes are RotationLike aren't they? I'm going to think about this some more before reviewing those changes.

I believe so, there are some other bug fixes that I discovered as well.

src/build123d/geometry.py
jdegenstein: RE: what happens when user passes None as all three parameters, this is what happens currently.
```py
Location(position=None, orientation=None, ordering=None) # Returns: Location((0,0,0),(0,0,0))
Location(None, None, None) # Returns: TypeError: Unexpected input type: [<class 'NoneType'>, <class 'NoneType'>, <class 'NoneType'>]

PR #1000, opened 2025-05-28
Bugfix: correct `ArcArcTangentArc` tangency with reversed start_arc in algebra mode

### LINKED ISSUE #983: ArcArcTangentArc placing result at wrong spot sometimes
I can't figure out how to reproduce more concisely, and I originally believed it was related to edges being reversed, but that does not appear to be the case in my investigation. cc @jwagenet since you know more about these new objects than anyone.

Image

```py

with BuildPart() as p:
    with BuildLine() as l:
        m1 = JernArc((160/2,0),(1,0),220,-39.327)
        print(m1@1)
    with BuildSketch() as s:
        with PolarLocations(0,4):
            SlotArc(m1,80)
    extrude(amount=30)

    with BuildSketch() as s2:
        add(s.sketch)
        offset(amount=-20)
    extrude(amount=30,mode=Mode.SUBTRACT)

ax = Axis((0,0,0),(1,1,0))
e1 = s.sketch.edges().filter_by_position(ax,0,30).sort_by(Axis.X)
zz = ArcArcTangentArc(e1[0],e1[1],25,side=Side.LEFT) # placed incorrectly, length is same as later uses, swapping sides does not fix it but create a different wrong result

b1 = CenterArc((80,0),40,90,180)
b2 = CenterArc((0,80),40,0,-180)
aata = ArcArcTangentArc(b1,b2,25,side=Side.RIGHT) # works fine

r1 = Edge(b1.wrapped.Reversed())
r2 = Edge(b2.wrapped.Reversed())
raata = ArcArcTangentArc(r1,r2,25,side=Side.RIGHT) # also works fine
```

jwagenet:
When `start_arc` has a reversed direction resulting in a negative plane normal in algebra mode, the resulting arc was drawn with the center facing the wrong way. `ArcArcTangentArc` now checks the tangency of the resulting arc and flips the arc if necessary. 

Resolves #983

black did some unrelated formatting I left in.

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -32,7 +32,7 @@ import copy as copy_module
 from collections.abc import Iterable
 from math import copysign, cos, radians, sin, sqrt
 from scipy.optimize import minimize
-import sympy # type: ignore
+import sympy  # type: ignore
 
 from build123d.build_common import WorkplaneList, flatten_sequence, validate_inputs
 from build123d.build_enums import (
@@ -456,7 +456,7 @@ class Helix(BaseEdgeObject):
     defined by cone_angle.
 
     If cone_angle is not 0, radius is the initial helix radius at center. cone_angle > 0
-    increases the final radius. cone_angle < 0 decreases the final radius. 
+    increases the final radius. cone_angle < 0 decreases the final radius.
 
     Args:
         pitch (float): distance between loops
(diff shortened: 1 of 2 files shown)


jdegenstein:
Looks good to me, thanks for working on this!

gumyr:
Thank you!  You got #1000 - congratulations!

PR #1003, opened 2025-05-29
Bugfixes: Slot objects

### LINKED ISSUE #948: SlotCenterPoint validates minimum width incorrectly
`SlotCenterPoint` raises `ValueError: Slots must have width > height. Got: {height=} width={half_line.length * 2} (computed)` when the slot width is too small, however the minimum width calculation is incorrect. Also, the fstring was broken by the formatter.

The minimum width should actually be `half_line.length * 2 + height` or validated as `half_line.length > 0` or `center != point`.

### Example
`SlotCenterPoint((0, 0), (1.5, 0), 3)` but this would actually produce a slot with overall length 6.

jwagenet:
- Fix `SlotCenterPoint` width check for min width of 0 instead of h. Fixes #948 
- Remove unnecessary circle conditional in `SlotOverall`
- Remove rotation step in `SlotArc` which doubled final rotation (likely no one would have encountered since the arc can be fully defined without additional rotation)

### DIFF
--- a/src/build123d/objects_sketch.py
+++ b/src/build123d/objects_sketch.py
@@ -426,10 +426,10 @@ class SlotCenterPoint(BaseSketchObject):
 
         half_line = point_v - center_v
 
-        if half_line.length * 2 <= height:
+        if half_line.length <= 0:
             raise ValueError(
-                f"Slots must have width > height. "
-                "Got: {height=} width={half_line.length * 2} (computed)"
+                "Distance between center and point must be greater than 0 "
+                f"Got: distance = {half_line.length} (computed)"
             )
 
         face = Face(
--- a/src/build123d/objects_sketch.py
+++ b/src/build123d/objects_sketch.py
@@ -521,17 +521,14 @@ class SlotOverall(BaseSketchObject):
         self.width = width
         self.slot_height = height
 
-        if width != height:
-            face = Face(
-                Wire(
-                    [
-                        Edge.make_line(Vector(-width / 2 + height / 2, 0, 0), Vector()),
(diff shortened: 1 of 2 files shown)


src/build123d/objects_sketch.py
```diff
@@ -521,17 +521,14 @@ def __init__(
...
-                Wire(
-                    [
-                        Edge.make_line(Vector(-width / 2 + height / 2, 0, 0), Vector()),
-                        Edge.make_line(Vector(), Vector(+width / 2 - height / 2, 0, 0)),
-                    ]
-                ).offset_2d(height / 2)
-            )
-        else:
-            face = cast(Face, Circle(width / 2, mode=mode).face())
+        face = Face(
```
gumyr: IIRC there was a user request to allow this slot to degenerate to a circle.

jwagenet: Gotcha, I didn't go back in the history. A value check blocks the width == height circle degeneration [at line 513](https://github.com/gumyr/build123d/pull/1003/files/811dd569d3bedabe26685f432edc49c1ca8b52be#diff-54e18cc1907978baadc97d042a4471892a9670078a8dcd95450f46236a68ba0fR513-R516), so I took out the branch:

Personally, it doesn't make sense semantically for a slot object to degenerate to a circle. None of the other slot objects do this and it's trivial for the user to do it. However, if you'd like to keep this behavior I can revert the change and update the value check.

gumyr: IIRC the user was generating slots from some automated code where a circle was a valid outcome and they convinced me to allow a circle. So is a circle a degenerate, but valid, slot?  I don't really know but it didn't really cause any problems so let's allow it.

gumyr:
Thank you

PR #1029, opened 2025-07-08
Remove subprocess usage in test_examples.py for Pyodide support

yeicor:
As discussed on Discord, the `subprocess` module is not supported in browser environments like `Pyodide`. To address this, I replaced the `subprocess` calls in `test_examples.py` with `exec`-based evaluation.

While `exec` does not provide the same isolation as spawning a new `Python` process, it's acceptable here because we control the example scripts being tested.

I already verified that the current examples run as expected in both `Pyodide` and standard `Python`.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -14,7 +14,6 @@ import os
 import sys
 import tempfile
 import unittest
-import codecs
 
 
 _examples_dir = Path(os.path.abspath(os.path.dirname(__file__))).parent / "examples"
@@ -51,7 +50,7 @@ def generate_example_test(path: Path):
             oldwd = os.getcwd()
             try:
                 os.chdir(cwd)
-                with codecs.open(path, 'r', 'utf-8') as f:
+                with open(path, 'r', encoding='utf-8') as f:
                     example_source = f.read()
                 exec(_MOCK_OCP_VSCODE_CONTENTS + example_source, {})
             finally: # Best-effort restore to previous state (examples should be safe)
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -11,10 +11,10 @@ desc: Unit tests for the build123d examples, ensuring they don't raise.
 from pathlib import Path
 
 import os
-import subprocess
 import sys
 import tempfile
 import unittest
+import codecs
 
 
 _examples_dir = Path(os.path.abspath(os.path.dirname(__file__))).parent / "examples"
@@ -22,8 +22,6 @@ _ttt_dir = Path(os.path.abspath(os.path.dirname(__file__))).parent / "docs/asset
 
 _MOCK_OCP_VSCODE_CONTENTS = """
 from pathlib import Path
-
-import re
 import sys
 from unittest.mock import Mock
 mock_module = Mock()
@@ -50,22 +48,16 @@ def generate_example_test(path: Path):
             # future that wants to both read assets from the examples
             # directory and write output files, deal with it then.
             cwd = tmpdir if 'benchy' not in path.name else _examples_dir
-            mock_ocp_vscode = Path(tmpdir) / "_mock_ocp_vscode.py"
-            with open(mock_ocp_vscode, "w", encoding="utf-8") as f:
-                f.write(_MOCK_OCP_VSCODE_CONTENTS)
-            got = subprocess.run(
-                [
-                    sys.executable,
-                    "-c",
-                    f"exec(open(r'{mock_ocp_vscode}').read()); exec(open(r'{path}').read())",
-                ],
-                capture_output=True,
-                cwd=cwd,
-                check=False,
-            )
-            self.assertEqual(
-                0, got.returncode, f"stdout/stderr: {got.stdout} / {got.stderr}"
-            )
+            oldwd = os.getcwd()
+            try:
+                os.chdir(cwd)
+                with codecs.open(path, 'r', 'utf-8') as f:
+                    example_source = f.read()
+                exec(_MOCK_OCP_VSCODE_CONTENTS + example_source, {})
+            finally: # Best-effort restore to previous state (examples should be safe)
+                os.chdir(oldwd)
+                sys.modules.pop("ocp_vscode", None)
+
 
     return assert_example_does_not_raise
 


tests/test_examples.py
```diff
@@ -50,22 +48,16 @@ def assert_example_does_not_raise(self):
...
-                cwd=cwd,
-                check=False,
-            )
-            self.assertEqual(
-                0, got.returncode, f"stdout/stderr: {got.stdout} / {got.stderr}"
-            )
+            oldwd = os.getcwd()
+            try:
+                os.chdir(cwd)
+                with codecs.open(path, 'r', 'utf-8') as f:
```
jdegenstein: can you change this line to `with open(path, 'r', encoding='utf-8') as f:`

(and remove the associated import)

yeicor: Sure, done and retested.

gumyr (commented):
This PR drops the checking of the return code.  Copilot suggests that the following would capture the output and check like the original did:
```py
exit_code = 0
try:
    with contextlib.redirect_stdout(stdout_buffer), contextlib.redirect_stderr(stderr_buffer):
        exec(_MOCK_OCP_VSCODE_CONTENTS + example_source, {})
except SystemExit as e:
    exit_code = e.code if isinstance(e.code, int) else 1
finally:
    os.chdir(oldwd)
    sys.modules.pop("ocp_vscode", None)

self.assertEqual(
    0, exit_code, f"stdout/stderr: {stdout_buffer.getvalue()} / {stderr_buffer.getvalue()}"
)
```
What do you think, is it worth doing the checking?

yeicor:
I ran a quick test to check how pytest handles sys.exit() and stdout/stderr during test execution.

If I add the following to the end of e.g. `vase_algebra.py`:

```python
import sys
print("This is some error message that is useful for debugging")
sys.exit(1)
```

And now I run all example tests with `python -m pytest tests/test_examples.py`, I get the following results:

    [...]
    collected 77 items                                          
    
    tests/test_examples.py .............................. [ 38%]
    .............................................F.                          [100%]
    
    =================================== FAILURES ===================================
    ______________________ TestExamples.test_vase_algebra_py _______________________
    [...]
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    
    >   ???
    E   SystemExit: 1
    
    <string>:47: SystemExit
    ----------------------------- Captured stdout call -----------------------------
    This is some error message that is useful for debugging
    [...]

As you can see, it includes the stdout and the exit code of the failed test. Hence, it seems that Copilot's suggestion adds unnecessary "complexity" that is already handled by pytest.

fischman:
FWIW the use of `subprocess` in the checked-in state was intentional, in order to run each example in isolation, to ensure that they remained individually valid (i.e. an example didn't silently rely on a symbol from another example, or any other observable state of the python process such as imported modules, etc). I think it would be valuable to retain this property on platforms that support it (i.e. non-WASM).

jdegenstein:
@fischman if that seems valuable to you, I am ok with it -- can add a platform specific branch like:
```py
if sys.platform == 'emscripten':
    # pyodide specific
else:
    # existing behavior with subprocess
```
What do you think?

yeicor:
Thanks for the review @fischman. I believe that the `exec` isolation problem is not as bad as you make it seem: the trick is that I use an empty dictionary as the globals so that the examples do not modify the global namespace. I relied on chatgpt to find possible issues and I discuss them at the end.

====================

<p>Yes, you <strong>can</strong> pass an empty dictionary (<code>{}</code>) to the <code>globals</code> parameter of Python 3’s <code>exec()</code> function <strong>to avoid modifying the actual global state</strong>, but with a few caveats and nuances you should be aware of:</p>
<hr>
<h3>✅ <strong>What Happens When You Pass <code>{}</code> to <code>globals</code>:</strong></h3>
<pre><code class="language-python">code = "a = 42"
exec(code, {})
</code></pre>
<ul>
<li>
<p>The code is executed with the given dictionary as the global namespace.</p>
</li>
<li>
<p>This <strong>prevents pollution of your actual global namespace</strong> (e.g., variables or functions defined in your module).</p>
</li>
<li>
<p>Any variables defined by the executed code go into that empty dictionary, not your global scope.</p>
</li>
</ul>
<hr>
<h3>⚠️ But There Are Important Details:</h3>
<ol>
<li>
<p><strong>Minimal Built-ins by Default</strong>:</p>
<ul>
<li>
<p>If you just pass <code>{}</code> to <code>globals</code>, there will be <strong>no <code>__builtins__</code></strong>, meaning things like <code>print</code>, <code>len</code>, <code>range</code>, etc., will be undefined unless you include them manually:</p>
<pre><code class="language-python">safe_globals = {"__builtins__": __builtins__}
exec("print('Hello')", safe_globals)  # Works
exec("print('Hello')", {})  # Raises NameError
</code></pre>
</li>
</ul>
</li>
<li>
<p><strong>Local Scope Defaults to Global Scope If Omitted</strong>:</p>
<ul>
<li>
<p>If you only pass <code>globals</code>, then <code>locals</code> defaults to the same dict. So the code will use the same dict for both local and global variables.</p>
</li>
</ul>
</li>
<li>
<p><strong>Does Not Prevent Side Effects</strong>:</p>
<ul>
<li>
<p>Even if you isolate the namespace, the executed code <strong>can still perform side effects</strong> like:</p>
<ul>
<li>
<p>Writing to disk</p>
</li>
<li>
<p>Importing modules</p>
</li>
<li>
<p>Modifying objects you pass in</p>
</li>
</ul>
</li>
<li>
<p>So you’re only isolating <em>namespace pollution</em>, not necessarily security or system safety.</p>
</li>
</ul>
</li>
<li>
<p><strong>Code Can Still Modify Objects You Put in the Globals</strong>:</p>
<ul>
<li>
<p>For example, if you pass in a global variable that refers to a mutable object, <code>exec()</code> code can mutate it.</p>
</li>
</ul>
</li>
</ol>

====================

And these are my answers to the caveats:

1. This is not true (see the [docs](https://docs.python.org/3/library/functions.html#exec)).
2. This is not a problem for the current use case.
3. Writing to disk would still be a problem with `subprocess`, importing modules seems to be ok as they are forgotten after exec() finishes (see demo of python repl below).
4. We are not sending any global that can be modified.

```
>>> exec("import math; a = math.sin(1)", {})
>>> a
  File "<python-input-1>", line 1, in <module>
    a
NameError: name 'a' is not defined
>>> math.sin(1)
  File "<python-input-2>", line 1, in <module>
    math.sin(1)
    ^^^^
NameError: name 'math' is not defined. Did you forget to import 'math'?
```

According to the demo above, I shouldn't even have to forget the ocp_vscode module as I did just in case in the patch. I can update this and write a comment about exec's behavior.

Assuming that the examples code is well-intentioned, using an isolated globals dict with exec() seems sufficient to avoid unintended global state changes, but correct me if I'm wrong. I’d personally prefer to avoid including Pyodide-specific code in build123d, but it’s not my call, and I’m fine with it if it's ultimately the chosen direction.

fischman:
Thanks for looking into it, @yeicor. I think the overview generated by chatgpt is incomplete. For example imported modules are singletons in Python, and importing a module inside the exec context gets you a (mutable!) handle on the same module object outside the exec context, enabling leakage.

I sympathize with wanting to avoid pyodide-specific code in b123d.
One way to accomplish this would be to patch subprocess in the pyodide test runner to swap the stdlib's (C) fork/exec impl of subprocess with your (py) exec impl.

yeicor:
Ok, I get it, things like this could have unintended effects:

```
>>> exec("import math; math.b = 'c'", {})
>>> import math
>>> math.b
'c'
```

I'm ok with any of the suggested solutions so far. I'll let the maintainers decide and update this pull request if required.

yeicor:
I ended up patching subprocess in the OCP.wasm tests instead, so no change is needed here. That may make more sense anyway, since that project is already full of hacky patches.

PR #1040, opened 2025-07-20
fix PolarLine with angles that make the length negative

### LINKED ISSUE #1041: PolarLine does not work properly with LengthMode.VERTICAL or HORIZONTAL
Currently `PolarLine((0,0,0), 10, -90, length_mode=LengthMode.VERTICAL).position_at(1)` yields `(0,10,0)` which is incorrect because it should give `(0,-10,0)`.

`LengthMode.HORIZONTAL` has a similar issue while `LengthMode.DIAGONAL` works fine.

danieledapo:
Without this patch `PolarLine((0,0,0), 10, -90, length_mode=LengthMode.VERTICAL).position_at(1)` yields `(0,10,0)` which is incorrect because it should give `(0,-10,0)`.

`LengthMode.HORIZONTAL` has the same issue.

Closes https://github.com/gumyr/build123d/issues/1041

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -779,9 +779,9 @@ class PolarLine(BaseEdgeObject):
         if length_mode == LengthMode.DIAGONAL:
             length_vector = direction_localized * length
         elif length_mode == LengthMode.HORIZONTAL:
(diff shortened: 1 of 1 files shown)


gumyr:
Thanks!

Issue #654, opened 2024-06-27
Add `BRepFill_Filling` based constrained surface creation


jdegenstein:
Currently the surface construction methods in build123d do not provide great control over adjacent face continuity requirements. I propose to add `BRepFill_Filling` as a build123d direct API `Face` method.

Further reading: https://neweopencascade.wordpress.com/2019/03/15/build-constrained-surfaces-with-brepoffsetapi_makefilling-part-i/

I was able to create a rough proof of concept, and it does function as expected:

![image](https://github.com/gumyr/build123d/assets/16868537/c9fb5b14-64fc-4e8e-b366-6fa4d7d1b279)

jdegenstein:
Working prototype:
```py
from typing import Iterable, Union, Tuple
from enum import Enum, auto
from OCP.BRepFill import BRepFill_Filling
from OCP.GeomAbs import *  # import specific

# put in build_enums.py
class Continuity(Enum):
    """geometric continuity"""

    C0 = auto()
    G1 = auto()
    G2 = auto()

    def __repr__(self):
        return f"<{self.__class__.__name__}.{self.name}>"

# put in topology.py ?
_continuity_dict = {
    Continuity.C0: GeomAbs_C0,
    Continuity.G1: GeomAbs_G1,
    Continuity.G2: GeomAbs_G2,
}

# put in topology.py as a classmethod of Face
@classmethod
def make_surface_patch(
    cls,
    edge_face_constraint: Iterable[
        Union[
            Tuple[Edge, Face, Continuity],
            Tuple[Edge, Continuity],
        ]
    ],
) -> Face:
    # input validation
    for efc in edge_face_constraint:
        if not isinstance(efc[0], Edge):
            raise ValueError(f"Provided {type(efc[0])} is not an Edge or Wire")

    patch = BRepFill_Filling()
    for efc in edge_face_constraint:
        if isinstance(efc[1], Continuity) and len(efc) == 2:
            patch.Add(efc[0].wrapped, _continuity_dict[efc[1]])
        else:
            patch.Add(efc[0].wrapped, efc[1].wrapped, _continuity_dict[efc[2]])
    patch.Build()
    result = patch.Face()

    return cls(result)


Face.make_surface_patch = make_surface_patch
```

gumyr:
This looks good to me.  At some point we're going to need some good surface modelling documentation with all of these different methods to choose from.

jdegenstein:
[...]

PR #1043, opened 2025-07-21
Add `Face.make_surface_patch` constrained surface modeling method

jdegenstein:
This is a PR to finally resolve issue https://github.com/gumyr/build123d/issues/654

I have changed a few things since the discussion was last active including changing the interface to separate out the 3 input types (edge/face/continuity constraints, edge constraints, and point constraints). The reason for this change was to simplify mypy type checking. I also switched to `BRepOffsetAPI_MakeFilling` which appears to be the method that OCCT created as the user-facing side of this, and it has almost the same API as the method in the linked issue above.

As discussed at the issue above, edge constraints no longer have the ability to specify continuity as OCCT seems to ignore anything other than C0. I also added several tests which provide some basic usage examples as well.

Example screenshot with result:
image


gumyr:
Nice!  This was one of the Roadmap items - happy to have it done.

Issue #793, opened 2024-11-17
docs: add complete build123d + visualization demo via ocp_vscode "standalone" mode


jdegenstein:
Now that ocp_vscode supports use in a regular browser (and not strictly bound with VSCode), it is much easier to explain how one could simply:

1. `pip install build123d ocp_vscode`
2. execute `python -m ocp_vscode`
3. open `http://127.0.0.1:3939/viewer` in a suitable browser
4. execute a python file with necessary build123d and ocp_vscode imports and `show` statement

As a way to get users "up and running quickly" I propose something similar to the above is added to the build123d docs (with appropriate links/caveats back to the `ocp_vscode` repo/docs). Further I propose that a screenshot is added to show what this could look like (maybe with vim + chrome, or something similar).

gumyr:
+1

jdegenstein:
Screenshot for the docs [using Ubuntu 20.04, gedit, gnome-terminal, ipython and firefox]:

![image](https://github.com/user-attachments/assets/be3474d3-710f-44d6-aceb-2e0a42e9a80d)

gumyr:
Nice

codingandmore:
Yes nice! I suggest to consider a few more things:

- If  you follow [the installation instruction](https://build123d.readthedocs.io/en/latest/installation.html)s there is no hint that you have to `pip install ocp_vscode`. For me it was confusing to see that the show()/show_all() functions can't be resolved. And I had no idea how to import them (you can find it much later somewhere). I understand that this is not part of build123d itself but every user needs to visualize his models somehow. 
- Beside the standalone mode there is an option "Move to new Window" in VSCode. Not sure that everyone will find this immediately. It is useful if you still want to work within VSCode but with limited screen space, like a laptop. If you assign a keyboard shortcut to toggle between the windows it can be quite convenient. This still supports debug mode. Yes I know this is not a VSCode documentation....

Would it make sense to start a discussion what else could get "users up and running quickly"? 

- if people have a programming background but not necessary a Python one explain environments, pip?
- Provide a VSCode profile template installing everything needed for build123d
- Provide an app template or example for frequently used functions in a command line using argparse (e.g. `python mymodel.py --export stl` , `python mymodel --show` , ...
Probably there are many more ideas. Might also be placed elsewhere in a wiki or so.

gumyr:
Here is a screen grab from the installation instructions:
[...]

PR #1047, opened 2025-07-24
[doc/external.rst] Explain different uses of ocp-vscode

sheltered-depths:
After reading [this issue](https://github.com/gumyr/build123d/issues/793), I attempted to clarify the different ways of using ocp-vscode. I'm pretty new to build123d, so please scrutinize.

### DIFF
--- a/docs/external.rst
+++ b/docs/external.rst
@@ -4,16 +4,39 @@
 External Tools and Libraries
 ############################
 
-The following sections describe tools and libraries external to build123d
-that extend its functionality.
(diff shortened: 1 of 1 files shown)


gumyr:
Thanks for taking the time to improve the documentation! This is a tricky area since ocp_vscode is a separate project from build123d, even though it’s our recommended viewer. There’s an ongoing discussion about how to better organize and expand the documentation overall, so I think it’s best to revisit how we present ocp_vscode as part of that broader effort.

Your contribution is appreciated; it helps highlight where users see room for improvement, which is really valuable for planning those updates.

PR #1048, opened 2025-07-25
Add missing docstring to `Face.make_surface_patch` and enhance error checking / handling and related tests

jdegenstein:
Method was previously missing docstring, I found that the error handling was insufficient for practical use as well especially when the returned result was empty. Enhanced both and added a few basic tests for the error handling.


jdegenstein:
Merged.

Issue #1042, opened 2025-07-21
ArcArcTangentArc generating wrong result


jdegenstein:
Desired result is the radius arc (m4), as far as I know m3 should generate the same result, but seems to find an incorrect solution that is tangent to neither of the input arcs.

```py
with BuildSketch() as s:
    Circle(35)
    with Locations((200/2,40)):
        Circle(35)
    make_hull()

sel = s.sketch.edges().sort_by(Axis.X)[-1]

m2 = RadiusArc((-32/2,38.8+7.5+7),(32/2,38.8+7.5+7),-200)
m3 = ArcArcTangentArc(m2, sel, 200)
m4 = RadiusArc(m2@1,sel@1,-200)
```

<img width="681" height="375" alt="Image" src="https://github.com/user-attachments/assets/e717fb8a-c212-4940-adbd-d0af51617672" />

jwagenet:
Initial observations:
1. You've revealed that I neglected (forgot? ignored?) to handle a non-overlapping circle case for tangent arcs crossing the midline between the arcs (see `ArcArcTangentArc` for linear crossing). I could probably add this, but would require an additional side/keep parameter. Your m3 is tangent to m2 at those points, but the arc center is on the wrong side (supposed to be fixed in #1000).
2. Technically, I didn't intend to support connecting arc segments which are already tangent with the appropriate arc, so this particular case would receive a ValueError accordingly. Currently, the tangent arc (or line) created will be tangent to the whole arcs with no checking for segment tangency.
3. In testing this I discovered another bug in `BuildLine` where two tangent arcs with centers on each side of the end points is created! I don't remember seeing this in testing, but may also be related to changes in #1000.

Thanks for the stress testing!

jwagenet:
@jdegenstein does #1058 resolve this issue?

This change should make sure you don't get the invalid arc, however it does not complete arc m2 with m4 as you were attempting to do. I'm not strictly opposed to "completing the arc", but it basically becomes a new branch with some other considerations.

Your snippet has a max_radius ValueError because they are actually slightly overlapping
```py
with BuildSketch() as s:
    Circle(35)
    with Locations((200/2,40)):
        Circle(35)
    make_hull()

sel = s.sketch.edges().sort_by(Axis.X)[-1]

m2 = RadiusArc((-32/2,38.8+7.5+7),(32/2,38.8+7.5+7),-200)
[...]

PR #1058, opened 2025-08-01
Feature: `ArcArcTangentArc` add missing tangent cases and fix bad objects in BuildLine 

jwagenet:
`ArcArcTangentArc` should now be complete as I think all tangent cases are now covered. If you think I missed a case please let me know.

As before, this object does not check if the final tangent arc joins incomplete arc segments or not and does not complete/join the arc segments if they are already tangent/coincident.

* Add inside crossing tangent case for non-overlapping arc as noted in #1042
* Add cases for partial and full overlapping arcs
* Add `short_sagitta` option
* Expand `keep` to a tuple to cover all tangency cases:
    * tried to make this easy enough to guess correct arc without referencing 
    * position 1 'placement' defines whether start_arc is internally (INSIDE) tangent to the tangent arc or externally tangent (except special BOTH case)
    * position 2 'type' defines whether the tangent arc is crossing/internal (INSIDE) or not
    * **breaking change** as default (INSIDE. INSIDE) does not correspond to previous INSIDE
* Set construction objects to mode.PRIVATE to avoid being added to BuildLine context as noted in [comment](https://github.com/gumyr/build123d/issues/1042#issuecomment-3134794146)
* Added non-exhaustive tests to spot check all min/max radius limits and tangent arc tangency

arcarctangentarc_keep_table


gumyr:
@jwagenet thank you for all the detailed work that has gone into making these arc objects - there are so many cases to consider!

bernhard-42:
Just out of curiosity, is there a reason not to use from [`OCP.GccAna.GccAna_Circ2d2TanRad`](https://dev.opencascade.org/doc/refman/html/class_gcc_ana___circ2d2_tan_rad.html)
I had played around with it a few weeks ago and it seems to work well for my few tests
image

See also https://dev.opencascade.org/doc/overview/html/occt_user_guides__modeling_algos.html#occt_modalg_2_4,

jwagenet:
@bernhard-42  Mostly because I didn't know about it when I made the initial TangentArc classes. I hadn't seen this overview doc before so I should skim through and save it, but I've generally found the OCCT docs mind-numbing to read so I wouldn't have seen this otherwise. 

`OCP.GccAna` seems like it could replace tangent objects I added + some others I've partially completed in addition to adding some other tangent-based objects.

PR #1062, opened 2025-08-06
add ability to do perspective projection

Mx-El:
I use build123d for making 2d art for pen plotting, and it looks weird without the ability to do perspective projection. Default is still orthographic projection as before, so should be completely backwards compatible. When calling, you just provide a focal length if you want perspective.


gumyr:
I can't think of a way to test this other than by looking at it (or some complex math) so let's hope that HLRAlgo_Projector is working correctly.  Thanks for the addition!

Mx-El:
Yeah, I just eyeballed it and confirmed that "thing farther away looks smaller" and given that my use case is for art, that was plenty good enough for me, lol

Issue #966, opened 2025-04-18
Consider adding Color ENUMs or switching to Web Color names


gumyr:
The Color names correspond to OCCT defined Enum values.  A Python Enum could be dynamically generated from the OCCT Enum as follows:
```py
from enum import Enum
import OCP.Quantity as Quantity
import inspect

# Use inspect.getmembers to access enum attributes
enum_members = [
    (name.replace("Quantity_NOC_", ""), value)
    for name, value in inspect.getmembers(Quantity.Quantity_NameOfColor)
    if name.startswith("Quantity_NOC_")
]

# Dynamically create the Python Enum
Color = Enum("Color", {name: value for name, value in enum_members})

# Example usage - in geometry.py
color = Color.BLACK
...
self.wrapped = Quantity_Color(color.value)
```

Alternatively, the https://webcolors.readthedocs.io/en/latest/ package provides a simple interface to the Web Colors which could replace the OCCT names entirely.

This issue is related to https://github.com/gumyr/build123d/issues/856

KarthikDani:
Amazing

jwagenet:
If we want an Enum, but not a massive list, the basic colors could be used: https://en.wikipedia.org/wiki/Web_colors#Basic_colors
However, I think using `webcolors` and the extended color string list is fine here.

PR #1072, opened 2025-08-13
Feature: add `ColorLike` handling to `Color`

### LINKED ISSUE #656: Add ColorLike with Color and CSS color string
As an enhancement, add `ColorLike = Union[Color, str]` and allow CCS strings as `.color` attribute, e.g. `obj.color="aliceblue"`

**Note:** [Open cascade color names](https://dev.opencascade.org/doc/refman/html/_quantity___name_of_color_8hxx.html#aadc2a1b709b06601041190b0a00c926a) and [CSS color names](https://www.w3schools.com/cssref/css_colors.php) are not identical. Open cascade has most of the CSS names, but e.g. not "crimson", but it does have the same color with different prefixes ("1", "2", "3", "4") that have different vales

### LINKED ISSUE #856: Add ColorLike type alias
Creating a type alias like this:

```py
from typing import TypeAlias
ColorLike: TypeAlias = (
    _.Color | # build123d color
    _.Quantity_ColorRGBA | # OCP color
    str | # name, ex: "red"
    tuple[str, int] | # name + alpha, ex: ("red", 0.5)
    tuple[float, float, float] | # rvb, ex: (1, 0, 0)
    tuple[float, float, float, int] | # rvb + alpha, ex: (1, 0, 0, 0.5)
    int | # hexa, ex: 0xff0000
    tuple[int, int] # hexa + alpha, ex: (0xff0000, 0x80)
)
```

and a convertion method like this:

```py
class Color:
    @classmethod
    def from_color_like(color_like: ColorLike) -> Color: ...
```

and replacing Color method parameters by a ColorLike might be practical.

It will allow for instance this syntax:

```py
obj = Solid(obj=box.wrapped, color="red")
```

Instead of explitly instantiating the Color object:

```py
obj = Solid(obj=box.wrapped, color=Color("red"))
```

I can create a PR if you agree.

### LINKED ISSUE #957: `_create_xde` in `export_step` does not accept color as string
This fails for `AttributeError: 'str' object has no attribute 'wrapped'`.

````python
part = Box(1, 1, 1)
part.color = "black" # or "#000000"
part.label = "box"
export_step(part, part.label + ".step")
````

Using `Color` instead of string works, however string colors should be supported everywhere if supported anywhere.

````python
part.color = Color(0, 0, 0)
````

### PR DESCRIPTION
**Pull request by jwagenet**
Credit to @njourdane for the initial `ColorLike` definition

* Add `ColorLike` alias (see below).
* Add `ColorLike` to `Color` with some internal restructuring
* Remove/replace explicit `Quantity_ColorRGBA` and `color_tuple` overloads
* Update Shape.color setter to accept ColorLike eg. `Solid(obj=box.wrapped, color="red")`
* Add support for css3 colors as a primary name string, falls back to OCCT/X11 colors, then fails if neither exist
* Maintain support for incomplete color formats eg. `Color(1., 2.)`
* Change str color hint from "~" to "is/near" depending if a css3 color is found or it falls back to OCCT fuzzy compare

### `ColorLike` types:
All non-object types explicitly accept complete form with alpha or not
```py
name, ex: "red" (Also supports "#ff0000"),
name + alpha, ex: ("red", 0.5),
rgb, ex: (1., 0., 0.),
rgb + alpha, ex: (1., 0., 0., 0.5),
hex, ex: 0xff0000,
hex + alpha, ex: (0xff0000, 0x80),
Color,
Quantity_ColorRGBA
```

### Issues
* Resolve #656 by adding `ColorLike` and adding css3 color name support
* Resolve #856 by adding `ColorLike` and updating `Shape.color`
* Fix #957 by updating `Shape.color` to fix downstream failure
* Partially addresses #966 by adding webcolor/css3 names, but does not add enums from any color set (IMO string literals are sufficient here).

### Notes
* webcolor dependency version is set the same as for ocp_tesselate
* Can also follow through with removing `to_tuple`
* repr floating point could probably be truncated to 6 figs or something
* Getting into the weeds: adding color name to repr is kinda half-baked since OCCT `Quantity_Color.Name()` is always a nearest fuzzy match (without further comparisons). I tried the OCCT nearest match method (convert to sRGB and find nearest neighbor) with webcolor and the results were qualitatively mediocre, so maybe `Quantity_Color.Name()` isn't great either. Not a fan of the name in repr, IMO using string hex (`"#ff0000"`) in an editor like VSCode with a color picker will give best UX for non-named colors. Exposing `name` (if given as a valid string name) and `hex` as properties could be a superior value add.


src/build123d/geometry.py
```diff
@@ -1292,14 +1308,61 @@ def __deepcopy__(self, _memo) -> Color:
...
+        red, remainder = divmod(triplet, 256**2)
+        green, blue = divmod(remainder, 256)
+        return red / 255, green / 255, blue / 255
+
+    @staticmethod
+    def _rgb_from_str(name: str) -> tuple:
+        if "#" not in name:
+            try:
+                # Use css3 color names by default
+                triplet = webcolors.name_to_rgb(name)
```
gumyr, line 1339: I don't think `webcolors.name_to_rgb` strips excess white space from the string - adding a `.strip()` here could save some user frustration.

src/build123d/topology/shape_core.py
```diff
@@ -281,15 +284,15 @@ def __init__(
         self,
         obj: TopoDS_Shape | None = None,
         label: str = "",
-        color: Color | None = None,
+        color: Color | ColorLike | None = None,
```
gumyr: `Color | ColorLike` is redundant as `Color` is included in `ColorLike` (just like with `VectorLike`).

src/build123d/topology/shape_core.py
```diff
@@ -281,15 +284,15 @@ def __init__(
...
         parent: Compound | None = None,
     ):
         self.wrapped: TOPODS | None = (
             tcast(Optional[TOPODS], downcast(obj)) if obj is not None else None
         )
         self.for_construction = False
         self.label = label
-        self._color = color
+        self.color = color
```
gumyr, line 295: Why is `_color` removed? Doesn't this conflict with the `color` property?

gumyr (commented):
Thank you, a lot of effort has gone into this!

jdegenstein:
I requested bernhard to take a look -- seems like making these changes should keep ocp_vscode in consideration as well.  IIRC some of the problems in the past have been related to differences between build123d and ocp_vscode.

src/build123d/topology/shape_core.py
jwagenet: AFAIK `self.color = color` calls the color setter instead of overwriting the property, in this case to assure `self._color` is always set the same way.

jwagenet:
> I requested bernhard to take a look -- seems like making these changes should keep ocp_vscode in consideration as well. IIRC some of the problems in the past have been related to differences between build123d and ocp_vscode.

Sounds good. `Color.wrapped = Quantity_ColorRGBA` has not changed, so hopefully there are no issues there. This change primarily expands accepted input formats for `Color`.

src/build123d/geometry.py
jwagenet: Added a strip step in [e6cc2c6](https://github.com/gumyr/build123d/pull/1072/commits/e6cc2c6c0eaf5b87a60a7c4715c2f566053796cd). Applies to `ColorFromName_s` too.

src/build123d/topology/shape_core.py
jwagenet: Fixed where applicable in [e6cc2c6](https://github.com/gumyr/build123d/pull/1072/commits/e6cc2c6c0eaf5b87a60a7c4715c2f566053796cd).

gumyr:
Timed out on waiting for more review.  Thanks @jwagenet, this should resolve problems where users were assigning colors incorrectly but getting things to look like they were working due to the flexibility of ocp_vscode.

PR #1081, opened 2025-08-28
Add Mixin1D.discretize

paulftw:
Discretize to a 1d shape is like tessellate to a 2d shape, but only the latter exists in build123d


gumyr:
Thanks for this PR. It took me a while to appreciate the difference between `Mixin1D.discretize` and `Mixin1D.positions`.  Given that they both produce a list of points from a 1D object I think the `discretize` algorithm should be integrated into the existing `positions` code.  What do you think of adding and optional `deflection` parameter which when set uses the `discretize` algorithm. To avoid some issues with users selecting the wrong algorithm with `quasi`, maybe GCPnts_QuasiUniformDeflection should be used with Wires and GCPnts_UniformDeflection with Edges as Wires can be non C2 continuous but Edges are always C2 continuous (unless the user uses the restricted `_to_bspline` method)?

Issue #328, opened 2023-09-22
Intersect Everything


gumyr:
As mentioned in Issue #327 it should be possible to intersect the geometric objects with each other but also the topological objects. Much of this functionality is already present but it should be reorganized to be consistent across the whole project.  It may be desirable to have specific names for some operations, for example a `section` is just the intersection of a `Plane` and a `Shape` so the logic should be in the `intersect` method(s) wrappers for well understood names.

A survey of the related existing intersection related methods needs to be done to assess what is missing.

gumyr:
The existing intersection related methods are as follows:

```py
class Plane:
    def find_intersection(self, axis: Axis) -> Union[Vector, None]:
        """Find intersection of axis and plane"""

class Shape:
    def intersect(self, *to_intersect: Shape) -> Self:
        """Intersection of the positional arguments and this Shape.

        Args:
            toIntersect (sequence of Shape): shape to intersect

        Returns:

        """

    def faces_intersected_by_axis(
        self,
        axis: Axis,
        tol: float = 1e-4,
    ) -> ShapeList[Face]:
        """Line Intersection

        Computes the intersections between the provided axis and the faces of this Shape

        Args:
            axis (Axis): Axis on which the intersection line rests
            tol (float, optional): Intersection tolerance. Defaults to 1e-4.

        Returns:
            list[Face]: A list of intersected faces sorted by distance from axis.position
        """

    def find_intersection(self, axis: Axis) -> list[tuple[Vector, Vector]]:
        """Find point and normal at intersection

        Return both the point(s) and normal(s) of the intersection of the axis and the shape

        Args:
            axis (Axis): axis defining the intersection line

        Returns:
            list[tuple[Vector, Vector]]: Point and normal of intersection
        """

class Compound:
    def do_children_intersect(
        self, include_parent: bool = False, tolerance: float = 1e-5
    ) -> tuple[bool, tuple[Shape, Shape], float]:
        """Do Children Intersect

        Determine if any of the child objects within a Compound/assembly intersect by
        intersecting each of the shapes with each other and checking for
        a common volume.

        Args:
            include_parent (bool, optional): check parent for intersections. Defaults to False.
            tolerance (float, optional): maximum allowable volume difference. Defaults to 1e-5.

        Returns:
            tuple[bool, tuple[Shape, Shape], float]:
                do the object intersect, intersecting objects, volume of intersection
        """

    def intersect(self, *to_intersect: Shape) -> Compound:
        """Construct shape intersection

[...]

PR #1090, opened 2025-09-16
Intersect Everything: Geometry and 0D, 1D Shape

jwagenet:
**Supposed to just be Geometry and 0D Shapes, staging as draft due to type conflict with 1D**

Adding test framework for a matrix of intersection tests + tweaks to geometry and `Vertex` intersect methods. See https://github.com/gumyr/build123d/issues/328#issuecomment-3300257613 for change objectives. This change focuses primarily on input and output conditions, not intersection finding methods.

### Geometry
* `Location` intersections can return `Vector` or `Location` depending on target orientation
* Coplanar `Plane` intersection returns `self`

### 0D Shape
* `Vertex` gets `intersect` method
* `Vertex` intersection return a `ShapeList`, even if one intersection is expected (to conform to future of other Shape dimensions)

### 1D Shape
* `intersect` moved from `Edge` to `Mixin1d`
* Intersection returns a `ShapeList`, even if one intersection is expected (to conform to future of other Shape dimensions)
* `Wire` is a valid `self` and `to_intersect` type
* `Shape` is a valid `to_intersect` type
* Restructured intersection loops to intersect across all items instead of just `self + edge`. 
* Remove Vertex intersections coincident Edge intersections (only return higher order objects)
* `Wire` and `to_intersect >= 2` not exhaustively tested for sanity

### Test Matrix
* Add test_intersection.py. Some tests may duplicate existing tests
* Known failing tests are skipped by default. A flag can be set to fail for dev
* Included tests for known issues
* Non-exhaustive tests for geometry and 0d `Shapes` to at least test all input combinations for expected output


gumyr:
Thank you! My grandson is visiting so I won’t be able to look at this for a few days.

jwagenet:
Added updated 1d.intersect

Not sure why codcov is looking at a bunch of shape_core stuff I didnt touch, but I'll see if I can clear up the relevant warnings.

mypy was clean on my system (cached maybe?), so I need to dig into those errors. **Edit**: The mypy errors are on the dev branch

jwagenet:
Considering how `Edge.intersect(Plane, Plane)` should work for testing for this case (represented with Faces):
image
The methodology is wrong as the correct result is `[Vertex]`, but currently would get `None` with the final set intersection. Method should be updated to `((A∩B)∩C)...∩n`, were any empty result exits the loop and returns `None`

jwagenet:
Updated methodology as described above (`((A∩B)∩C)...∩N`)to intersect each intersect result from `to_intersect` sequentially and exit if a result is `None` for both `Vertex` and `Mixin1D` (more important for 1d).

gumyr:
Finally getting to a point where I can start to review this but first a question.  I've got the 1st drop of the `Edge.make_constrained_arcs` functionality ready to merge into `dev`.  As part of this I've fixed up a whole bunch of `mypy` typing problems when using the OCP headers (we don't use them in GitHub so those problems don't show). Unfortunately this has churned up `one_d.py` quite a bit.  Does it matter to you if I merge this now?  I didn't touch any of the intersect code so an automatic merge seems likely to work.

jwagenet:
@gumyr Please feel free to merge your work. I can review merge conflicts that arise.

gumyr:
> @gumyr Please feel free to merge your work. I can review merge conflicts that arise.

Done.  I've found that getting mypy happy with OCP is painful but once it's done it there should only be small updates required to keep things up-to-date.

gumyr:
This work is looking very good; I'm learning from your use of pytest.  How did you want to merge? Wait to the "end" when you're happy or sections like this?

jwagenet:
I'd say its best to merge. I don't think there is a downside to merging this section now (and future sections as they are complete). My goal is for each chunk to be self contained.

gumyr:
Merged.  A huge thank you for this work!  It isn't necessarily exciting but moves us that much closer to the v1.0.0 release.

PR #1096, opened 2025-09-28
Fix for non-associative multiplications when using GridLocations

Dmytry:
Currently, build123d multiplication is unexpectedly non ~~commutative~~ associative when using lists like GridLocations: 

`(Rot(45) * GridLocations(10, 10, 2, 2)) * Cylinder(2, 3)` works but `Rot(45) * (GridLocations(10, 10, 2, 2) * Cylinder(2, 3))` throws an exception.

I'm not entirely sure how best to do the type hints for new multiplication, though, since the changed code would also accept nested iterables.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1705,9 +1705,12 @@ class Location:
     @overload
     def __mul__(self, other: Iterable[Location]) -> list[Location]: ...
 
+    @overload
+    def __mul__(self, other: Iterable[Shape]) -> list[Shape]: ...
+
     def __mul__(
         self, other: Shape | Location | Iterable[Location | Shape | Iterable]
-    ) -> Shape | Location | list[Location | Shape | list]:
+    ) -> Shape | Location | list[Location] | list[Shape]:
         """Combine locations"""
         if self.wrapped is None:
             raise ValueError("Cannot move a shape at an empty location")
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1706,8 +1706,8 @@ class Location:
     def __mul__(self, other: Iterable[Location]) -> list[Location]: ...
 
     def __mul__(
-        self, other: Shape | Location | Iterable[Location]
-    ) -> Shape | Location | list[Location]:
+        self, other: Shape | Location | Iterable[Location | Shape | Iterable]
+    ) -> Shape | Location | list[Location | Shape | list]:
         """Combine locations"""
         if self.wrapped is None:
             raise ValueError("Cannot move a shape at an empty location")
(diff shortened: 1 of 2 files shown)


gumyr:
@bernhard-42 I'd like to get your opinion on this one.  TIA

bernhard-42:
Sorry for missing it.

The reason is that `Location.__mul__` only vectorizes the multiplication of `Location` with `list(Location)` but not with `list(Shape)`
This is the needed change to make it associative:

```diff
❯ git diff src/build123d/geometry.py
diff --git a/src/build123d/geometry.py b/src/build123d/geometry.py
index ad04b0d..3d9cc0b 100644
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1843,14 +1843,21 @@ class Location:
                 raise ValueError("Can't multiply by empty location")
             return Location(self.wrapped * other.wrapped)

-        # other is a list of Locations
         if isinstance(other, Iterable):
             others = list(other)
-            if not all(isinstance(o, Location) for o in others):
-                raise ValueError("other must be a list of Locations")
             if any(o.wrapped is None for o in others):
                 raise ValueError("Can't multiple by empty Locations")
-            return [Location(self.wrapped * loc.wrapped) for loc in others]
+            if all(isinstance(o, Location) for o in others):
+                return [Location(self.wrapped * loc.wrapped) for loc in others]
+            elif all(
+                hasattr(o, "wrapped") and isinstance(o.wrapped, TopoDS_Shape)
+                for o in others
+            ):
+                return [self * shape for shape in others]
+            else:
+                raise ValueError(
+                    "other must be a list of Locations or a list of shapes"
+                )

         raise ValueError(f"Invalid input {other}")
         
```

bernhard-42 (commented):
I think I would be a bit more restrictive:

```python
           elif all(
                hasattr(o, "wrapped") and isinstance(o.wrapped, TopoDS_Shape)
                for o in others
            ):
```
instead of only `if isinstance(other, Iterable):`

**EDIT: that should have gone into the review - struggling with review feature, sorry**

gumyr:
Thanks for the PR and for identifying this inconsistency in the positioning syntax.

We’ve decided not to merge this change. While applying a `Location` to a homogeneous list of shapes is geometrically reasonable, the proposed implementation broadens `Location.__mul__` to recursively accept arbitrary iterables. That would allow mixed and nested collections with inconsistent return types and less predictable error behavior.

Because `*` is overloaded for transforms, shapes, location collections, and Python sequences, it is not generally associative—especially in longer expressions. We do not want to expand the core operator semantics for this alternative parenthesization.

The existing form remains supported:
```py
Rot(45) * GridLocations(10, 10, 2, 2) * Cylinder(2, 3)
```
If a strong need emerges later, we could consider a narrowly defined operation for a homogeneous collection of shapes, with explicit semantics and broader tests. Closing this PR for now, with thanks again for the report and implementation.

PR #1106, opened 2025-10-11
fix: gradient error in gordon surface intersect

gongfan99:
modified:   pyproject.toml

The intersect point detection algorithm is more reliable now.
Also the bspline with kink can be the input to make_gordon_surface.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
     "ipython >= 8.0.0, < 10",
     "lib3mf >= 2.4.1",
     "ocpsvg >= 0.5, < 0.6",
-    "ocp_gordon >= 0.1.13",
+    "ocp_gordon >= 0.1.14",
     "trianglesolver",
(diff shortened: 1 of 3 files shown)


gumyr:
I didn't know that a zero length Geom_BSplineCurve was possible - interesting.

It seems as though allowing points as profile/guides in the middle would be problematic - doesn't that create two Faces? If it is possible I'm concerned that the Face won't be really valid.  How about restricting the point(s) to the first or last of the profiles/guides?

gongfan99:
@gumyr That makes sense. I have limited the point to be the first or last of profile/guide.

The argument `profiles` or `guides`  in `Face.make_gordon_surface` does not need to be in order because `ocp_gordon` has automatic curve sorting function. Hence, the limitation (for the point being the first or last of profile/guide) can only be implemented on the `ocp_gordon` side.

gumyr:
Thank you!

PR #1110, opened 2025-10-21
Improve shapes typing

snoyer:
This should improve the type checking of Shapes by:

- renaming the `.wrapped` member that can be `None` to `._wrapped` to be used internally only
- exposing the public `.wrapped` as a property that is asserted `not None`
- implementing `__bool__` so we can do `if shape` instead of `if shape.wrapped is not None`
- making the `Mixin*D` classes properly subclass `Shape[...]` (instead of `Shape` without the generic argument)

before and after from a user point of view:
b123d-typing-before
b123d-typing-after

### DIFF
--- a/src/build123d/topology/two_d.py
+++ b/src/build123d/topology/two_d.py
@@ -213,7 +213,7 @@ class Mixin2D(ABC, Shape):
 
     def __neg__(self) -> Self:
         """Reverse normal operator -"""
-        if self.wrapped is None:
+        if self._wrapped is None:
             raise ValueError("Invalid Shape")
         new_surface = copy.deepcopy(self)
         new_surface.wrapped = downcast(self.wrapped.Complemented())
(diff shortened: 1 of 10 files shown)


gumyr:
We're in the process of creating a new release so we're putting a hold on new content like this for a little while.  Thanks and stay tuned...

src/build123d/topology/two_d.py
```diff
@@ -2359,7 +2360,7 @@ def __init__(
             obj = obj_list[0]
 
         if isinstance(obj, Face):
-            if obj.wrapped is None:
+            if not obj.wrapped:
```
gumyr: Shouldn't this be `if not obj:`

snoyer: nice catch, will fix :)

gumyr:
Thank you!

PR #1113, opened 2025-10-21
Replace `Face.make_plane()` with `Face(Plane)`

jwagenet:
Replace `Face.make_plane()` with `Face(Plane)` to make geometry -> `Shape` conversion consistent with `Vertex(Vector)` and `Edge(Axis)`.

### DIFF
--- a/src/build123d/topology/two_d.py
+++ b/src/build123d/topology/two_d.py
@@ -449,7 +449,7 @@ class Face(Mixin2D, Shape[TopoDS_Face]):
     @overload
     def __init__(
         self,
-        obj: TopoDS_Face,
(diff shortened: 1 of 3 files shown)


gumyr:
We're in the process of creating a new release so we're putting a hold on new content like this for a little while. Thanks and stay tuned...

gumyr:
This is a nice change which makes the API more Pythonic and improves consistency.  However, `make_plane` shouldn't be entirely removed yet, let's put a deprecation warning in there and give users a chance to fix their code.

jwagenet:
No problem. Readded the method with the warning, but didnt revert the tests (in case codecov complains)

gumyr:
Thank you!

Issue #1119, opened 2025-10-29
Incorrect `extrude` `Until.LAST` example due to target mirrored revolve creating invalid Solid


jwagenet:
`extrude.py` `Until.LAST` example `ex26` produces incorrect result due to invalid target. Invalid target is a mirrored revolve. 

The example should be fixed, but I think a complete resolution requires updating `extrude` to check for an invalid target.

### Incorrect example
```py
with BuildPart() as ex26:
    with BuildSketch() as ex26_sk:
        with Locations((0, rev)):
            Circle(rad)
    revolve(axis=Axis.X, revolution_arc=90)
    mirror(about=Plane.XZ)
    with BuildSketch() as ex26_sk2:
        Rectangle(rad, rev)
    ex26_target = ex26.part
    extrude(until=Until.LAST, clean=False, mode=Mode.REPLACE)
```

<img width="252" height="225" alt="Image" src="https://github.com/user-attachments/assets/62ce204f-5d42-4797-9437-a3981eb0e034" />

### Expected

Replace `revolve/mirror` with `revolve(axis=Axis.X, revolution_arc=180)`

<img width="240" height="222" alt="Image" src="https://github.com/user-attachments/assets/822e16aa-b36b-4feb-bc2d-f40927b96c9e" />

gumyr:
This example is # 36 now:
```py
rad, rev = 6, 50

with BuildPart() as ex36:
    with BuildSketch() as ex36_sk:
        with Locations((0, rev)):
            Circle(rad)
    revolve(axis=Axis.X, revolution_arc=180)
    with BuildSketch() as ex36_sk2:
        Rectangle(rad, rev)
    extrude(until=Until.NEXT)
```
and works fine (there was a change in extrude until).  I'm going to close the issue - hopefully it doesn't recur.

PR #1120, opened 2025-10-29
Intersect Everything: 2D, 3D, Composite Shapes

### LINKED ISSUE #697: Intersection of an edge with a face or a wire doesn't work anymore
I have quite a few scripts that worked with build123d 0.5.x that do something like this (with more complex shapes than just a circle):
```py
Edge.make_line((-1, 0), (1, 0)).intersect(Wire.make_circle(0.5))
Edge.make_line((-1, 0), (1, 0)).intersect(Face(Wire.make_circle(0.5)))
```
I have just upgraded to 0.6.1 and both calls [now return `NotImplemented`](https://github.com/gumyr/build123d/blob/235260a8000fa9d5174df0efea100d7cff332ec6/src/build123d/topology.py#L4692). A bummer.
IIUC the change was to include intersection points/vertices when the input is an edge or an axis. But can we just fall back to the old behaviour when other shapes are passed to the intersect function?

### LINKED ISSUE #918: Basic face-face intersection does not work
I'm pretty sure this used to work.

```py
a = Rectangle(10,10)
b = Plane.XZ * a
i = a.intersect(b)
```

Image

### LINKED ISSUE #945: 1d intersect anything is broken
```py
s = Circle(10)
l = Line(-20, 20)
i = l & s
```

```
ValueError: Unknown object type: <class 'build123d.objects_sketch.Circle'>
```

Reordering the operation (`s & l`) works

jwagenet:
Add a matrix of intersection tests + higher order Shape intersect methods. See https://github.com/gumyr/build123d/issues/328#issuecomment-3300257613 for change objectives. This change focuses on input and output conditions, but adding `BRepAlgoAPI_Section` as a second `_bool_op` operation has proven to produce good results.

### 2D, 3D, Composite Shapes

- Intersection returns a ShapeList, even if one intersection is expected
- `BRepAlgoAPI_Section` is added as a second operation to `BRepAlgoAPI_Common`, but sometimes both produce duplicate intersections
- ~Wire, Shell, and~ Compound are dissolved to Vertex, Edge, Face, Solid ahead of intersections. Compound intersections can produce extra, duplicate Shapes
- Remove lower order shapes from intersection list which _appear_ to be part of a higher order shape using a lazy distance check
- All 3 methods follow a similar, duplicated structure

### 1D Shapes
- Methodology updated to use `BRepAlgoAPI_Section` and  `BRepAlgoAPI_Common` 
- Wire is now dissolved after intersection

### Test Matrix

* Some Solid/Edge combinations are skipped due to producing duplicate, but correct Edges
* Non-exhaustive tests for 2d, 3d, and Composite Shapes

### Other Fixes
I'm relying on the existing tests to capture indirect breakages
* Correct `extrude.py` example so intersection produces correct result for test (see #1119)
* Update a handful of tests and methods to correctly handle `None | ShapeList` result
* Update `mode == Mode.INTERSECT` to produce `A ∩ B  + A ∩ C` result instead of `A ∩ B ∩ C`. Shape.intersect(B, C) does not produce `A ∩ B ∩ C`. Discovered in `test_ppp_0108`. **Edit**: this is now accomplished by wrapping objects in Compound 
* Update `_bool_op` to ignore Shapes without `order` when determining highest order. Allows for passing Shape(Topods_Face) for Plane from Mixin1D

### Todo
* Some intersect combinations produce duplicate Edge intersections. No existing method to detect duplicates
* ~`_bool_op` restructure logic to dissolve Wire and Shell results after intersection should be possible~ Done
* Attempt to refactor duplicated code in 2D, 3D, Composite methods
* ~`Vertex` and `Mixin1D` intersect methods may be able to follow logic of higher order methods leading to potential refactor back into `Shape.intersect`~ Done. Doesn't seem necessary for Vertex unless refactored into `Shape.intersect`
* Consider updating `Shape._bool_op` output to ~~match intersect result types~~ `ShapeList` (largeish blast radius)
* Update `Shape.intersect` output to return `A ∩ B ∩ C` for `to_intersect = [B, C]`

### Resolves (see issue tests)
* Resolves #945
* Resolves #918
* Resolves #697


gumyr:
Awesome work!  I don't know how to thank you enough for this.  

As @jdegenstein is doing a release at the moment I think we'll hold off on merging this until that is complete.  Maybe we can do another release soon afterward with this as it will be a large improvement of what this there.

jwagenet:
Interested in fielding any architectural feedback about the current (and 1d, 2d) changes and items on the todo list. Some specific thoughts:
* RE: code duplication. Is it worth attempting to coerce as much as possible into `Shape.intersect` (seems each dim level may still need a thin wrapper for type casting to work) 
* Is `filter_shapes_by_order` robust enough  (tests useful here) and should its functionality be a `Shape` method?

Less on topic:
* Is there any reason (A.X, A.Y, A.Z) == (B.X, B.Y, B.Z) is insufficient for a `Vertex` equality check. Is there a technical reason why matching identity may be preferred here?
* Whats the history behind `Edge.intersect` using the `find_intersection_points` instead of `BRepAlgoAPI_Section`, which seems to work just as well (possible risk of duplicates when checking Edge + Edge)?

gumyr:
I'd like to know more about: 
> Update mode == Mode.INTERSECT to produce A ∩ B  + A ∩ C result instead of A ∩ B ∩ C. Shape.intersect(B, C) does not produce A ∩ B ∩ C. Discovered in test_ppp_0108
What is `test_ppp_0108`? I'm trying to understand this change and need an example.

jwagenet:
Sure, I appreciate the scrutiny here. 

[PPP_0108](https://build123d.readthedocs.io/en/latest/tttt.html#party-pack-01-08-tie-plate) and other TooTallTobys/examples are tested and benchmarked; if the mass assertion fails the test fails. For this particular part, the shape of the rib feature is made by `extrude(amount=30, both=True, mode=Mode.INTERSECT)`. The `both` flag passes two unfused solids to `Builder._add_to_context` which handles the intersection between self._obj and objects. 
image

In 0.10.0, this calls `Shape.intersect` which in turn passes self and objects to `_bool_op`. This produces the desired/expected result of context self with the union of the extrude objects:
https://github.com/gumyr/build123d/blob/e6d98de840cffe89accf8fbdf36d564010a36c6f/src/build123d/build_common.py#L466-L469
https://github.com/gumyr/build123d/blob/e6d98de840cffe89accf8fbdf36d564010a36c6f/src/build123d/topology/shape_core.py#L1394-L1395
As far as I can tell, the items in `args` and `tools` (self and objects (also to_intersect)) passed to `BRepAlgoAPI_Common` are unioned before intersection: `self ∩ objects`. 
image

However this change introduces the update to `Solid.intersect()` to independently intersect each `to_intersect` item with self._obj: `self._obj ∩ object[0] ∩ object[1] ...`.  Reverting to 0.10.0 `Builder._add_to_context` produces the only the result as the intersection of self with separate extrude objects instead of the unioned objects:
image

My solution is to call `self._obj.intersect` for each item in objects:
https://github.com/gumyr/build123d/blob/b049e6a8cee277f4239b100cedfe51ace33ff12d/src/build123d/build_common.py#L466-L475

Two alternate solutions are to fuse objects or combine in a compound before calling the intersection

gumyr:
Thanks for the details, I was having a hard time to come up with a way to create this circumstance (and for the pointer to the TTT tests, I forgot about those). From a performance point of view merging to a compound and doing a single intersection would be more efficient. However, there might not be many ways to create a situation like this so maybe it doesn't really matter.  What do you think?

You have a TODO list above, did you want to merge then complete those items or wait until everything is done?

gumyr:
>  RE: code duplication. Is it worth attempting to coerce as much as possible into Shape.intersect (seems each dim level may still need a thin wrapper for type casting to work)

I would say no, intersection is quite complex so having class specific implementations is totally appropriate.  If there are some common algorithms they could be factored out in the further to compact the code a little.

> Is filter_shapes_by_order robust enough (tests useful here) and should its functionality be a Shape method?

Since the shapes being compared are all being extracted from the original shape maybe we can just compare the TopoDS objects or use the OCCT history mechanism to determine if they are from the original?  I've prototyped this for extrude until:
```py
    # 2: Intersect the extrusion with the target to find the target's modified faces
    intersect_op = BRepAlgoAPI_Common(target.wrapped, extrusion.wrapped)
    intersect_op.Build()
    intersection = intersect_op.Shape()
    face_exp = TopExp_Explorer(intersection, TopAbs_FACE)
    if not face_exp.More():
        raise ValueError("No intersection: extrusion does not contact target")

    # Find the faces from the intersection that originated on the target
    history = intersect_op.History()
    modified_target_faces = []
    face_explorer = TopExp_Explorer(target.wrapped, TopAbs_FACE)
    while face_explorer.More():
        target_face = TopoDS.Face_s(face_explorer.Current())
        modified_los: TopTools_ListOfShape = history.Modified(target_face)
        while not modified_los.IsEmpty():
            modified_face = modified_los.First()
            modified_los.RemoveFirst()
            modified_target_faces.append(modified_face)
        face_explorer.Next()
```
History also keeps track of `Generated` (as well as `Modified`).  It seems like this could be used to check if objects are originating from the same place.

> Is there any reason (A.X, A.Y, A.Z) == (B.X, B.Y, B.Z) is insufficient for a Vertex equality check. Is there a technical reason why matching identity may be preferred here?

A Vertex has a Location so depending on what is being done just checking the x,y,z values might not be enough.

> Whats the history behind Edge.intersect using the find_intersection_points instead of BRepAlgoAPI_Section, which seems to work just as well (possible risk of duplicates when checking Edge + Edge)?

IIRC `find_intersection_points` was created before I knew about BRepAlgoAPI_Section.  In general I like to refactor custom implementations for OCCT versions (if they work :) ), OCCT is so huge that just knowing everything it can do it near impossible although LLM really, really help with this.

jwagenet:
> You have a TODO list above, did you want to merge then complete those items or wait until everything is done?

This change is intended to stand alone, so it can be merged when the content is satisfactory (although its unfortunate the intersect update is split over the 0.10.0 release), and also attempt to reduce review burden of many changes. In light of the following this merge can wait until resolved:

>  From a performance point of view merging to a compound and doing a single intersection would be more efficient.

Agree. I don't have knowledge in this area. For what its worth, creating the compound just cleans up the call in `Builder._add_to_context`. I currently split Compounds before intersecting so its just rearranging where the logic is happening. The todo to move splitting Wires and Shells after `_bool_op` rather than before may also apply to Compounds to tidy things up.

> filter_shapes_by_order

I'll have to take a look your snippet. Based on `BRepAlgoAPI_Section` and `BRepAlgoAPI_Common` returning duplicate Shapes with different internal ids, I'm not convinced `_bool_op` results retain child/parent relationships. The history aspect is interesting and worth exploring.

> BRepAlgoAPI_Section

Noted.

jwagenet:
Pushed changes to:
- Convert 1D to 2D+ methodology
- Intersect Wire and Shell Directly
- Update _bool_op to take Shapes without `order` attribute
- Revert `mode == Mode.INTERSECT` iteration and pass Compound instead

gumyr:
Thank you! A large step towards the 1.0 release.

PR #1121, opened 2025-10-31
added support for passing an iterable of radii to FilletPolyline.

x0pherl:
inspired by 
image

tests pass; the touched method has 100% coverage.
docs have been updated to reflect the change
linting, type annotations, and formatter have all been executed

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -824,12 +824,13 @@ class FilletPolyline(BaseLineObject):
             radius_list = [radius] * len(points)  # Single radius for all points
         else:
             radius_list = list(radius)
-            if len(radius_list) != len(points):
+            if len(radius_list) != len(points) - int(not close) * 2:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


x0pherl:
note: this needs to be updated to properly handle closed=False conditions; discussed in Discord; I'll submit an update tomorrow

Jojain:
I'm not a fan of having 2 separate lists for vertex and radii with the expectation that they have both the same amount of items, its easy to not have them in sync.
Usually we would prefer using a dict as the source of data.

But what do to with the existing radius input then...

x0pherl:
> I'm not a fan of having 2 separate lists for vertex and radii

I don't love that either, but maintaining backwards compatibility seems important here. However, I'm already back to the board to handled closed=False cases (where there are fewer vertices than points) so I'll consider if there are nice clean ways to address this

gumyr:
Thanks!  @jdegenstein is working on a release now so we'll delay merging this for a little while.

gumyr:
Thank you!

PR #1129, opened 2025-11-08
changes to make development more friendly on MacOS

x0pherl:
- Adding .DS_Store to .gitignore
- adding quotes to pip commands in the CONTRIBUTING markdown to make them work universally (including on MacOS zshell). you can see [https://github.com/mu-editor/mu/issues/852](https://github.com/mu-editor/mu/issues/852) for a discussion on why this change is needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ venv.bak/
 
 # Profiling debris.
 prof/
+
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #1131, opened 2025-11-09
Use normal.cross(Vector(0,0,-1)) for Plane(face) x_dir to ensure consistent orientation

javimixet:
### Summary
Use a consistent geometric rule for determining `x_dir` in `Plane(face)` instead of relying on the face surface’s U-derivative.

### Details
When `Plane` is created from a `Face` without an explicit `x_dir`, the X direction is now computed as:

```python
x_dir = face.normal_at().cross(Vector(0, 0, -1))
```

- For horizontal faces, X aligns with global +X.  
- For vertical/sloped faces, X lies parallel to the global XY plane and points **right** to an observer standing in front of the face.  

This produces predictable, human-intuitive orientations and avoids inconsistent X directions between adjacent faces.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -2814,13 +2814,11 @@ class Plane(metaclass=PlaneMeta):
                 raise ValueError("Planes can only be created from planar faces")
             face_normal = Plane.get_topods_face_normal(arg_face.wrapped)
             face_z_dir = Vector(face_normal).normalized()
-
-            z_threshold = 1 - TOLERANCE
-            if abs(face_z_dir.Z) > z_threshold:
+            candidate = face_z_dir.cross(Vector(0, 0, -1))
+            if candidate.length < TOLERANCE:
                 face_x_dir = Vector(1, 0, 0)
             else:
-                face_x_dir = face_z_dir.cross(Vector(0, 0, -1)).normalized()
-
+                face_x_dir = candidate.normalized()
             self._origin = arg_face.center()
             self.x_dir = Vector(arg_x_dir) if arg_x_dir else Vector(face_x_dir)
             self.x_dir = Vector(round(i, 14) for i in self.x_dir)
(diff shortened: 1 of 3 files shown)


src/build123d/geometry.py
```diff
@@ -2810,25 +2810,21 @@ def __init__(self, *args, **kwargs):
...
-                surface.D1(0.5, 0.5, point, face_x_dir, tangent_v)
+            face_normal = Plane.get_topods_face_normal(arg_face.wrapped)
+            face_z_dir = Vector(face_normal).normalized()
+
+            z_threshold = 1 - TOLERANCE
+            if abs(face_z_dir.Z) > z_threshold:
+                face_x_dir = Vector(1, 0, 0)
             else:
-                face_x_dir = surface.Position().XDirection()
+                face_x_dir = face_z_dir.cross(Vector(0, 0, -1)).normalized()
```
gumyr: I don't think this works.  If the `face_z_dir` is (0,0,-1) the cross product is (0,0,0) which will cause this to fail.

gumyr (commented):
I'm not a fan of trying to fix this "problem". If there is not enough information to create the plane then the user shouldn't expect a known outcome.  One can provide an x_dir when creating a Plane from a Face or an Axis which allows the user total control over the resulting Plane.

src/build123d/geometry.py
javimixet: Don't lines 2819 and 2820 handle that?

javimixet:
> I'm not a fan of trying to fix this "problem". If there is not enough information to create the plane then the user shouldn't expect a known outcome. One can provide an x_dir when creating a Plane from a Face or an Axis which allows the user total control over the resulting Plane.

That makes sense, and I agree this isn’t a bug in the current behavior. My intent wasn’t to “fix” anything, but to suggest a slightly more predictable default when x_dir isn’t provided.

Users already have full control by explicitly passing an x_dir, and this wouldn’t change that. I see it as a small default enhancement, in line with how build123d already handles many defaults—but of course, it’s your call, and I’m completely fine if you’d rather keep things as they are.

src/build123d/geometry.py
gumyr: Not mathematically but practically maybe yes.  I'm concerned that there might be magic number stuff happening when there are small differences in the z direction calculation.

PR #1140, opened 2025-11-16
FilletPolyLines allow 0-radius

AlexVerschoot:
Hello,

So i'd like to enable zero-radius items in FilletPolyLines. Since it's just very handy to be able to mix a normal polyline (a fillet value 0), and a fillet polyline.

I got the code working, but got stuck on the linter and mypy. So it's not production ready. 

I also was unable to update the docs (      File "/Users/alex/build123d/.venv/lib/python3.12/site-packages/sphinx/events.py", line 415, in emit
        raise ExtensionError(
    sphinx.errors.ExtensionError: Handler <function merge_typehints at 0x124f5b740> for event 'object-description-transform' threw an exception (exception: unhashable type: 'list')
)

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -787,20 +787,22 @@ class Helix(BaseEdgeObject):
 
 class FilletPolyline(BaseLineObject):
     """Line Object: Fillet Polyline
-
     Create a sequence of straight lines defined by successive points that are filleted
     to a given radius.
 
     Args:
         pts (VectorLike | Iterable[VectorLike]): sequence of two or more points
-        radius (float | Iterable[float]): radius to fillet at each vertex or a single value for all vertices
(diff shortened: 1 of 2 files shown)


jwagenet:
For what its worth, it looks like you are only updating the docstring, so you don't need to rebuild the docs (though it can be good to confirm the formatting is correct).

gumyr:
Thank you!

AlexVerschoot:
> Thank you!

I actually wasn't fully done yet, i was thinking of reworking the code, to use a polyline, and fillet that, while rebuilding the "closed mechanism". This to see if the linter would be happier, and to get the code better.

Is that something you would like me to do?

gumyr:
Yes, code improvements are welcome. I didn't change anything significant, just fixed up the typing.

Issue #738, opened 2024-10-16
Mesher / 3MF crashes when run as a file upon exit


jdegenstein:
This is a non-critical issue since the crash occurs AFTER all of the script has run. I currently believe the crash occurs while resources are being freed up as the python interpreter exits. Here is a test-case e.g. `crashtest.py`:
```py
from build123d import *

with BuildPart() as board:
    Box(1, 2, 3)

exporter = Mesher() # generates one exception
exporter.add_shape(board.part) # generates a second exception
```
Run like:
```
$ python crashtest.py
Exception ignored in: <function Base.__del__ at 0x7f0458f61120>
Traceback (most recent call last):
  File "/home/oem/miniconda3/envs/ocp_vscode2/lib/python3.11/site-packages/py_lib3mf/Lib3MF.py", line 4327, in __del__
  File "/home/oem/miniconda3/envs/ocp_vscode2/lib/python3.11/site-packages/py_lib3mf/Lib3MF.py", line 4091, in Release
  File "/home/oem/miniconda3/envs/ocp_vscode2/lib/python3.11/site-packages/py_lib3mf/Lib3MF.py", line 4026, in checkError
AttributeError: 'NoneType' object has no attribute 'SUCCESS'
Exception ignored in: <function Base.__del__ at 0x7f0458f61120>
Traceback (most recent call last):
  File "/home/oem/miniconda3/envs/ocp_vscode2/lib/python3.11/site-packages/py_lib3mf/Lib3MF.py", line 4327, in __del__
  File "/home/oem/miniconda3/envs/ocp_vscode2/lib/python3.11/site-packages/py_lib3mf/Lib3MF.py", line 4091, in Release
  File "/home/oem/miniconda3/envs/ocp_vscode2/lib/python3.11/site-packages/py_lib3mf/Lib3MF.py", line 4026, in checkError
AttributeError: 'NoneType' object has no attribute 'SUCCESS'
```
I currently suspect that some logic can be added to `class Mesher` that will solve this problem (such as to free up resources as the script exits somehow). It also could be a problem with the lib3mf python bindings.

gautaz:
Hello,

I just ran into this issue with the following source code:

```python
from build123d import *

length, width, thickness = 80.0, 60.0, 10.0
center_hole_dia = 22.0

with BuildPart() as ex2:
    Box(length, width, thickness)
    Cylinder(radius=center_hole_dia / 2, height=thickness, mode=Mode.SUBTRACT)

[...]

PR #1141, opened 2025-11-18
Add BytesIO export support to SVG, DXF, STEP, and STL/3MF via a lib3mf/Mesher workaround

### LINKED ISSUE #1024: add BytesIO support to remaining export functions (GLTF/GLB, STL via native OCCT)
Currently `export_brep` supports BytesIO export, consider adding to other export functions. It may make upcoming wasm in-browser usage easier.

ref: https://github.com/yeicor/OCP.wasm/issues/4

jdegenstein:
Opening as draft for now.

Adds:
1. SVG export to BytesIO
2. DXF ''
3. STEP ''
4. STL/3MF via a lib3mf/Mesher workaround. STL is not possible to do in-memory using OCCT native methods
5. Necessary tests to cover these changes

### DIFF
--- a/src/build123d/mesher.py
+++ b/src/build123d/mesher.py
@@ -542,7 +542,7 @@ class Mesher:
         """write
 
         Args:
-            file_name Union[Pathlike, str, bytes, BytesIO]: file path
+            file_name Union[Pathlike, str, bytes]: file path
 
         Raises:
             ValueError: Unknown file format - must be 3mf or stl
--- a/src/build123d/mesher.py
+++ b/src/build123d/mesher.py
@@ -83,6 +83,7 @@ license:
 # pylint: disable=no-name-in-module, import-error
 import copy as copy_module
 import ctypes
+from io import BytesIO
 import math
 import os
 import sys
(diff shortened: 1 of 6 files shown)


gumyr:
Looks good to me.  Just out of curiosity, does the 3MF SUCCESS error message occur when writing to BytesIO?

src/build123d/mesher.py
```diff
@@ -540,7 +542,7 @@ def write(self, file_name: PathLike | str | bytes):
         """write
 
         Args:
-            file_name Union[Pathlike, str, bytes]: file path
+            file_name Union[Pathlike, str, bytes, BytesIO]: file path
```
jdegenstein: TODO: revert this

src/build123d/mesher.py
```diff
@@ -551,3 +553,8 @@ def write(self, file_name: PathLike | str | bytes):
             raise ValueError(f"Unknown file format {output_file_extension}")
         writer = self.model.QueryWriter(output_file_extension[1:])
         writer.WriteToFile(file_name)
+
+    def write_stream(self, stream: BytesIO, file_type: str):
```
jdegenstein, line 557: @gumyr do you have any opinion on the naming for this new `Mesher.write_stream` method? The backstory for why I chose to do this is that the `Mesher.write` method infers the file format (STL or 3MF) from the output file extension, which is not applicable to BytesIO.

gumyr: Why can't write do both?  if the `file_name` is a BytesIO stream go down the stream path you've already created?

jdegenstein: @gumyr it technically CAN but I want to support exporting to BOTH 3MF and STL and with BytesIO there is no mechanism for the current write method to do it (again it infers the file format from the file extension of the file_name).

gumyr: Got it.  `write_stream` is fine by me.

jdegenstein:
> Looks good to me. Just out of curiosity, does the 3MF SUCCESS error message occur when writing to BytesIO?

@gumyr yes that issue still occurs although it is not triggered by write itself, see here https://github.com/gumyr/build123d/issues/738

PR #1144, opened 2025-11-19
Fix Example 14 header in introductory_examples.rst

Birdulon:
Header had "1." instead of "14."


jdegenstein:
Thank you! I just noticed this today and was thinking about fixing it

PR #1160, opened 2025-12-06
Improve Geometry class string representations

### LINKED ISSUE #1104: Improve Axis __repr__
The Axis __repr__ just shows a tuple of vectors which can be confusing.

jwagenet:
Improve and make consistent string representations for `Vector`, `Location`, `Axis`, and `Plane`.
All classes now follow format:
```py
ax = Axis((1.11, 2.22, 3.33), (4.44, 5.55, 6.66))
f"{ax:.3f}": "Axis((1.110, 2.220, 3.330), (0.455, 0.569, 0.683))"
str(ax): "Axis: (position=(1.11, 2.22, 3.33), direction=(0.455842, 0.569803, 0.683763))"
repr(ax): "Axis((1.11, 2.22, 3.33), (0.455842, 0.569802, 0.683763))"
```

### Changes
* Add `__format__` to simplify formatting `Vector` tuples and `Location`, `Axis`, `Plane` tuple of tuples
* `__repr__` follows format `class_name(attribute1, attribute2, ...)` where float format is `.6g` (`TOL_DIGITS`). repr() is intended to explicitly reproduce the class if executed
* `__str__` follows format `class_name: ("attribute1"=attribute1, "attribute2"=attribute2, ...)` where float format is `.6g` (`TOL_DIGITS`)
* Use `.6g` for `Vector` formats in `show_topology`
* Correct `Location` kwarg names and make docstrings more concise
    * Explicitly `position, direction, angle` constructor in branches

Resolves #1104


jwagenet:
I'm not married to `.6g` precision for str and `.13g` precision for repr. I think both could be smaller, especially repr, since effective precision is generally limited by TOLERANCE

gumyr:
This is a very nice quality of life improvement - thank you.  I really like the `__format__` for `Vector`, something I've been meaning to improve for a long time.  I want the `__format__` method for the `Plane`, `Axis` and `Location` classes but that can happen later.

> I'm not married to .6g precision for str and .13g precision for repr. I think both could be smaller, especially repr, since effective precision is generally limited by TOLERANCE

I'm not sure what the right thing is to do here - but maybe explicitly using `TOL_DIGITS` as the default format precision would be appropriate (instead of 6)?  Anything beyond `TOLERANCE` is just noise but I wonder if there are negative side effects of limiting precision (the user would have to copy the values from the string to see this - so maybe I'm being paranoid).

bernhard-42:
My tests on linux arm showed one very small issue that nevertheless brings the show_topology test case to fail:
arm linux returns as `__repr__` string for a `Location` for one value `0.00`, while the expected string is `-0.00` (all other OS return `-0.00`) So the test is successful, but the rounding error in the string let it fail.

We could blame the test for string comparison, but that's the only way one can test `show_topology`, or should `__repr__` map the interval `[-tolerance, tolerance]` to positive `0.00`. 

From looking at the commits this seems to be covered in this PR, right?

gumyr:
Thanks @jwagenet !

@bernhard-42  did you run your test after @jwagenet's latest change (a couple hours before your comment)?  The float formatting doesn't change the sign:
```py
>>> print(f"{-0.00000000001:0.4f}")
-0.0000
```
so if different architectures are creating values around zero we probably have to change the test to avoid this.

jwagenet:
@bernhard-42 Is the `-0.0` issue issue persistent after the PR? When I first only applied the `g` spec alone to test values with `0.0` like `Location`, I had the same failures for only the macOS-14 and -15 tests, but adding the rounding step in the `Vector` formatter fixed this issue. Format specs other than `f` and `g` will still exhibit this as they fall back on `tuple` string formatting.

jwagenet:
I pushed the commit last night and didn't update, but I made the following changes:
* `Axis`, `Plane`, and `Location` now have `__format__` which return a string tuple of tuples for each class
* `repr` uses format directly, while `str` does not to avoid extra string operations to add the attribute names
* Both `str` and `repr` use `.TOL_DIGITSg` precision and formatting

I *did not* update `show_topology` to use the formats, but its worth noting the Location used to use `.2f`. I think `.3f` is reasonable for Loc, Center, Pos in the topo tree if fixed-width formatting is desirable.

Codecov continues to be difficult.

gumyr:
@bernhard-42 Does the linux arm issue occur with the latest version of this PR?

bernhard-42:
Sorry, this got lost during the holidays.

```bash
(build123d) bernhard@ubuntu2504:~/Development/CAD/build123d$ uname -nom
ubuntu2504 aarch64 GNU/Linux
```

Using cadquery_ocp_novtk 7.9.3 (with aarch64 supports)

```bash
(build123d) bernhard@ubuntu2504:~/Development/CAD/build123d$ uv pip list | grep -E '(build123|cadquery)'
build123d               0.10.1.dev16+gc38cd45cb.d20260108 /home/bernhard/Development/CAD/build123d
cadquery-ocp-novtk      7.9.3.0
cadquery-ocp-proxy      7.9.3.0
```

All show_topology tests pass with this PR

```bash
(build123d) bernhard@ubuntu2504:~/Development/CAD/build123d$ git branch
  dev
* geom_repr


(build123d) bernhard@ubuntu2504:~/Development/CAD/build123d$ pytest tests/test_direct_api/test_assembly.py -W ignore -v
============================================================ test session starts ============================================================
platform linux -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0 -- /home/bernhard/Development/CAD/build123d/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/bernhard/Development/CAD/build123d
configfile: pyproject.toml
collected 6 items                                                                                                                           

tests/test_direct_api/test_assembly.py::TestAssembly::test_attributes PASSED                                                          [ 16%]
tests/test_direct_api/test_assembly.py::TestAssembly::test_do_children_intersect PASSED                                               [ 33%]
tests/test_direct_api/test_assembly.py::TestAssembly::test_remove_child PASSED                                                        [ 50%]
tests/test_direct_api/test_assembly.py::TestAssembly::test_show_topology_compound PASSED                                              [ 66%]
tests/test_direct_api/test_assembly.py::TestAssembly::test_show_topology_shape PASSED                                                 [ 83%]
tests/test_direct_api/test_assembly.py::TestAssembly::test_show_topology_shape_location PASSED                                        [100%]

============================================================= 6 passed in 0.95s =============================================================
```

gumyr:
Thank You!

PR #1163, opened 2025-12-08
Fix Matrix constructor to create gp_GTrsf from nested list correctly

### LINKED ISSUE #1162: Matrix constructor does not create GTrsf correctly for `transform_shape`
Minimal test case:

```python

from build123d import *
box = Solid.make_box(1, 1, 1)
translate_matrix = [
        [1.0, 0.0, 0.0, 1.0],
        [0.0, 1.0, 0.0, 2.0],
        [0.0, 0.0, 1.0, 3.0],
        [0.0, 0.0, 0.0, 1.0],
]
box.transform_shape(Matrix(translate_matrix))
```

Expected result:
- transformation of some kind

Actual result:
```

File ~/devel/build123d/src/build123d/topology/shape_core.py:2189, in Shape.transform_shape(self, t_matrix)
   2186     return self
   2187 new_shape = copy.deepcopy(self, None)
   2188 transformed = downcast(
-> 2189     BRepBuilderAPI_Transform(self.wrapped, t_matrix.wrapped.Trsf()).Shape()
   2190 )
   2191 new_shape.wrapped = tcast(TOPODS, transformed)
   2193 return new_shape

Standard_ConstructionError: gp_GTrsf::Trsf() - non-orthogonal GTrsf
```

The `translate_matrix` above is copied from the existing test suite for `test_empty_shape`. But the same happens with the identity matrix of:

```python
[
        [1.0, 0.0, 0.0, 0.0],
        [0.0, 1.0, 0.0, 0.0],
        [0.0, 0.0, 1.0, 0.0],
        [0.0, 0.0, 0.0, 1.0],
]
```

If I instead create an identity matrix as follows:

```python
import OCP

gtrsf = OCP.gp.gp_GTrsf()
matrix = Matrix(gtrsf)
```

...then it works (or at least, I get no exception). When I examine the elements of the wrapped `GTrsf` objects in the two cases, they seem to be identical, but the one using the `Matrix([[...]])` doesn't work. It seems like doing `SetValue()` is not enough to set up the object correctly.

spookylukey:
Fixes #1162

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -2489,8 +2489,8 @@ class Matrix:
                 raise TypeError(f"{args[0]} is of an unexpected type")
 
         # Handle kwargs
-        trsf = kwargs.pop("trsf", default_trsf)
-        matrix = kwargs.pop("matrix", default_matrix)
+        trsf: gp_GTrsf = kwargs.pop("trsf", default_trsf)
+        matrix: Sequence[Sequence[float]] = kwargs.pop("matrix", default_matrix)
 
         # Handle unexpected kwargs
         if kwargs:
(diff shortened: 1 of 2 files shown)


spookylukey:
This doesn't work - it fixes the added test, but fails a bunch of existing tests for Matrix.

I don't understand what it is going on with `dp_GTrsf`, it seems to have some extra state that isn't the 4x3 matrix.

gumyr:
I'm going to close this PR. If you'd like to improve this in the future you're welcome to open another PR.

PR #1169, opened 2025-12-13
Feature: Support single line fonts and add FontManager Class

### LINKED ISSUE #1087: Consider adding support for single line fonts
Currently there are frequent issues reported by users of regular e.g. TTF fonts that are related to issues with the font face in CAD. One common issue is that the strokes of these fonts are self-intersecting, which can ultimately result in invalid shapes later in the CAD design process. Another approach I have been investigating for awhile is the use of "single line fonts". The idea is that the font face would be simply edges that are potentially unclosed (rather than faces like the existing build123d text approach). Fortunately OCCT has some direct support for this type of fonts (owing to their usefulness in a CAD context)

Now this raises an important question: how useful is a set of font characters composed of edges? The answer is varied, but for 3D printing it does not seem very useful (CNC and laser cutting might find this valuable directly though). From here these edges could be offset and fused to create actual `Face`s -- which would be useful in 3D printing. What is the difference with the existing build123d text approach? These characters should (in theory, not well tested) be more stable in build123d / OCCT because the faces were directly generated internally vs. imported.

Rough working prototype:
```py
from OCP.Font import (
    Font_FA_Bold,
    Font_FA_BoldItalic,
    Font_FA_Italic,
    Font_FA_Regular,
    Font_FontMgr,
    Font_SystemFont,
)
from OCP.gp import gp_Ax3
from OCP.Graphic3d import (
    Graphic3d_HTA_LEFT,
    Graphic3d_HTA_CENTER,
    Graphic3d_HTA_RIGHT,
    Graphic3d_VTA_BOTTOM,
    Graphic3d_VTA_CENTER,
    Graphic3d_VTA_TOP,
    Graphic3d_VTA_TOPFIRSTLINE,
)
from OCP.TCollection import TCollection_AsciiString
from OCP.StdPrs import StdPrs_BRepTextBuilder as Font_BRepTextBuilder, StdPrs_BRepFont
from OCP.NCollection import NCollection_Utf8String
from build123d.topology.utils import tuplify

@classmethod
def make_text_single(
    cls,
    txt: str,
    font_size: float,
    font: str = "Arial",
    font_path: str | None = None,
    font_style: FontStyle = FontStyle.REGULAR,
    text_align: tuple[TextAlign, TextAlign] = (TextAlign.CENTER, TextAlign.CENTER),
    align: Align | tuple[Align, Align] | None = None,
    position_on_path: float = 0.0,
    text_path: Edge | Wire | None = None,
    single_line_font: bool = False,
) -> Compound:
    """2D Text that optionally follows a path.

    The text that is created can be combined as with other sketch features by specifying
    a mode or rotated by the given angle.  In addition, edges have been previously created
    with arc or segment, the text will follow the path defined by these edges. The start
    parameter can be used to shift the text along the path to achieve precise positioning.

    Args:
        txt: text to be rendered
        font_size: size of the font in model units
        font: font name
        font_path: path to font file
        font_style: text style. Defaults to FontStyle.REGULAR
        text_align (tuple[TextAlign, TextAlign], optional): horizontal text align
            LEFT, CENTER, or RIGHT. Vertical text align BOTTOM, CENTER, TOP, or
            TOPFIRSTLINE. Defaults to (TextAlign.CENTER, TextAlign.CENTER)
        align (Union[Align, tuple[Align, Align]], optional): align min, center, or max
            of object. Defaults to None
        position_on_path: the relative location on path to position the text,
            between 0.0 and 1.0. Defaults to 0.0
        text_path: a path for the text to follows. Defaults to None (linear text)

    Returns:
        a Compound object containing multiple Faces representing the text

    Examples::

        fox = Compound.make_text(
            txt="The quick brown fox jumped over the lazy dog",
            font_size=10,
            position_on_path=0.1,
            text_path=jump_edge,
        )

    """
    # pylint: disable=too-many-locals

    def position_face(orig_face: Face) -> Face:
        """
        Reposition a face to the provided path

        Local coordinates are used to calculate the position of the face
        relative to the path. Global coordinates to position the face.
        """
        assert text_path is not None
        bbox = orig_face.bounding_box()
        face_bottom_center = Vector((bbox.min.X + bbox.max.X) / 2, 0, 0)
        relative_position_on_wire = (
            position_on_path + face_bottom_center.X / path_length
        )
        wire_tangent = text_path.tangent_at(relative_position_on_wire)
        wire_angle = Vector(1, 0, 0).get_signed_angle(wire_tangent)
        wire_position = text_path.position_at(relative_position_on_wire)

        return orig_face.translate(wire_position - face_bottom_center).rotate(
            Axis(wire_position, (0, 0, 1)),
            -wire_angle,
        )

    if sys.platform.startswith("linux"):
        os.environ["FONTCONFIG_FILE"] = "/etc/fonts/fonts.conf"
        os.environ["FONTCONFIG_PATH"] = "/etc/fonts/"

    font_kind = {
        FontStyle.REGULAR: Font_FA_Regular,
        FontStyle.BOLD: Font_FA_Bold,
        FontStyle.ITALIC: Font_FA_Italic,
        FontStyle.BOLDITALIC: Font_FA_BoldItalic,
    }[font_style]

    if text_align[0] not in [TextAlign.LEFT, TextAlign.CENTER, TextAlign.RIGHT]:
        raise ValueError(
            "Horizontal TextAlign must be LEFT, CENTER, or RIGHT. "
            f"Got {text_align[0]}"
        )

    if text_align[1] not in [
        TextAlign.BOTTOM,
        TextAlign.CENTER,
        TextAlign.TOP,
        TextAlign.TOPFIRSTLINE,
    ]:
        raise ValueError(
            "Vertical TextAlign must be BOTTOM, CENTER, TOP, or TOPFIRSTLINE. "
            f"Got {text_align[1]}"
        )

    horiz_align = {
        TextAlign.LEFT: Graphic3d_HTA_LEFT,
        TextAlign.CENTER: Graphic3d_HTA_CENTER,
        TextAlign.RIGHT: Graphic3d_HTA_RIGHT,
    }[text_align[0]]

    vert_align = {
        TextAlign.BOTTOM: Graphic3d_VTA_BOTTOM,
        TextAlign.CENTER: Graphic3d_VTA_CENTER,
        TextAlign.TOP: Graphic3d_VTA_TOP,
        TextAlign.TOPFIRSTLINE: Graphic3d_VTA_TOPFIRSTLINE,
    }[text_align[1]]

    mgr = Font_FontMgr.GetInstance_s()

    if font_path and mgr.CheckFont(TCollection_AsciiString(font_path).ToCString()):
        font_t = Font_SystemFont(TCollection_AsciiString(font_path))
        font_t.SetFontPath(font_kind, TCollection_AsciiString(font_path))
        mgr.RegisterFont(font_t, True)

    else:
        font_t = mgr.FindFont(TCollection_AsciiString(font), font_kind)

    logger.info(
        "Creating text with font %s located at %s",
        font_t.FontName().ToCString(),
        font_t.FontPath(font_kind).ToCString(),
    )
    if single_line_font:
        font_t.SetSingleStrokeFont(True)

    builder = Font_BRepTextBuilder()
    font_i = StdPrs_BRepFont(
        NCollection_Utf8String(font_t.FontName().ToCString()),
        font_kind,
        float(font_size),
    )

    if single_line_font:
        font_i.SetCompositeCurveMode(False)

    text_flat = Compound(
        builder.Perform(
            font_i, NCollection_Utf8String(txt), gp_Ax3(), horiz_align, vert_align
        )
    )
    
    tfe = text_flat.edges()
    # Align the text from the bounding box
    align_text = tuplify(align, 2)
    text_flat = text_flat.translate(
        Vector(*text_flat.bounding_box().to_align_offset(align_text))
    )
    if text_path is not None:
        path_length = text_path.length
        text_flat = Compound([position_face(f) for f in text_flat.faces()])
        return text_flat

    if not single_line_font:
        return text_flat
    else:
        return tfe

Compound.make_text_single = make_text_single
```

Rough usage example (could use a lot of refinement with some kind of built-in functionality):
```py
text = "the quick brown fox jumps over the lazy dog"
regular = Compound.make_text_single(text,10,"Relief SingleLine CAD")
singleline = Compound.make_text_single(text,10,"Relief SingleLine CAD", single_line_font=True)

offsets = [offset(e,amount=.5) for e in singleline.edges()]
offset_faces = [make_face(o) for o in offsets]
offset_faces_fused = Sketch() + offset_faces
```

With `offset_faces_fused` show in yellow and `singleline` show in brown:
Image

EDIT: I also forgot to attached the font that I used -- https://github.com/isdat-type/Relief-SingleLine which includes a so-called "open TTF" that enables this single line fonts (the contours are not closed).

jwagenet:
### FontManager class
* Rename utils.py to text.py for font/text objects
* Abstract `Font_FontMgr` into `FontManager`

OCP appears to persist the `Font_FontMgr` instance, so a user can register a font or all the fonts in a folder ahead of time:

```py
text = "the quick"
new_font_faces = FontManager().register_font("Roboto-VariableFont.ttf")
print(new_font_faces)
roboto = Compound.make_text(text, 10, "Roboto")
roboto_black = Compound.make_text(text, 10, "Roboto Black")

google_sans = Compound.make_text(text, 10, "Google Sans", "GoogleSans-VariableFont_GRAD,opsz,wght.ttf")
print([f.name for f in available_fonts() if "Google" in f.name])
google_sans_medium = Compound.make_text(text, 10, "Google Sans Medium")
```
```py
['Roboto Thin', 'Roboto ExtraLight', 'Roboto Light', 'Roboto', ...]
['Google Sans', 'Google Sans Medium']  
```

### Support single line fonts
* Bundle "Relief SingleLine CAD" as a built in single line font. This is OFL and I believe the requirements are satisfied. Has alias "singleline"
* Allow marking fonts as `single_line` in FontManager
* Extend `Compound.make_text` for single line/stroke fonts
    * `single_line_width` controls Face outline amount, 0 returns Wires
    * `offset_2d` and `_make_face` local method handle outlining
* Triad and BallJoint symbol use single line font

image

```py
singleline = Compound.make_text(text, 10, "Relief SingleLine CAD")
outline = Compound.make_text(text, 10, "Relief SingleLine CAD", text_path=Line((0, 10), (1, 10)), single_line_width=.5)
```

image

Resolves #1087


jwagenet:
### Draft Discussion
"Relief SingleLine CAD" is bundled is this a good decision?. I think this offers a solid default font option, although it has a limited character set (latin + accents). 

For the time being, there is no `make_text_line` as suggested by @jdegenstein, 1. to allow for discussion, 2. I'm not convinced a distinction matters at the direct method level because Compound can't semantically be composed of only Faces or Wires. Additionally, the outline step is easier to branch in process and _make_face is still necessary.

IMO, the outline step is required at some point to avoid passing only wires to `Text`

If `single_line_width` float control is undesirable, then 3 methods are required: `make_text`, `make_text_line`, `make_text_outline` or 3 objects `Text`, TextLine`, `TextOutline`. However, I suggest only 2 `Text`, TextLine` and 1 direct method as it stands.

Drafting labels may also benefit from "singleline" like the triad, although I dont know that the title block should necessarily use this font.

src/build123d/topology/composite.py
```diff
@@ -352,38 +340,30 @@ def position_face(orig_face: Face) -> Face:
...
         text_flat = Compound(
-            TopoDS.Compound_s(
-                builder.Perform(
-                    font_i,
-                    NCollection_Utf8String(txt),
-                    gp_Ax3(),
-                    horiz_align,
-                    vert_align,
-                )
+            builder.Perform(
```
gumyr: Doesn't mypy fail here without `TopoDS.Compound_s` to downcast the `TopoDS_Shape` to a `TopoDS_Compound`?  (OCP stubs will need to be active to see this.)

src/build123d/topology/composite.py
```diff
@@ -393,9 +373,24 @@ def position_face(orig_face: Face) -> Face:
...
+
+        def _make_face(edges: Iterable[Edge]) -> Face:
+            face = Face(Wire.combine(edges)[0])
+            if face.normal_at().Z < 0:  # flip up-side-down faces
+                face = -face  # pylint: disable=E1130
+            return face
+
+        # Outline single line text
+        if system_font.IsSingleStrokeFont() and single_line_width > 0:
+            outline = [e.offset_2d(single_line_width) for e in text_flat.edges()]
```
gumyr: I suspect that `trace` will be safer to use here than `offset_2d` and it makes a `Face` directly.

gumyr (commented):
The `Text` object needs some work doesn't it?  A `single_line_width` of `0.0` raises some interesting problems. I'm assuming that text as edges would be exported to SVG and have value - is that the intent?

I'd rather not break the convention that `Sketch` objects are a `Compound` of `Face` so it seems as though we should have a `BaseCompoundObject` that could contain either `Face` or `Edge` shapes and `Text` would be derived from that.

If `Text` is used in a `BuildLine` context then `Edges` are generated; however, if `Text` is used in `BuildSketch` then `single_line_width` must be greater than 0.

There are many generic operators but this would be the first generic object.

src/build123d/topology/composite.py
jwagenet: While I agree offset may introduce issues at large amounts, personally, I find the trace result poor for this application without rounded ends (as a matter of taste and the history of single line fonts for routed text). However, the `trace` result is also subpar: 
image
`offset_2d`:
image

src/build123d/topology/composite.py
jwagenet: @jdegenstein removed `TopoDS.Compound_s` from the issue example so I assumed it was fine. I'll investigate.

jwagenet: Also, if passing mypy with OCP stubs is critical perhaps they should be included in the GH type checker and CONTRIBUTING.md/pyproject.toml [development]...

jwagenet: Anyway, I see you added that as part of the typing fixes and it indeed failed with the stubs, so I've reverted that.

src/build123d/topology/composite.py
gumyr: `trace` can take a `Wire` which would help on the corners but I agree that rounded ends look better.  As we've seen, offset can be problematic but it looks like it's appropriate here.

src/build123d/topology/composite.py
gumyr: `mypy` checking with the OCP stubs was turned on in the middle of this PR so the timing is unfortunate.

jwagenet:
I'm completing work on this improvement:
* Add `single_line_width` to `Text` to control face size and ensure the object always returns a face using `outline_2d` as is
* Add `Text` reference docs section to `Objects` for a visual companion to capabilities. Could find a better home through a docs reorg.
* Do not create `TextLine` or similar object since there isn't a `BaseCurveObject` construction and text output is not great for sketch creation

gumyr:
Thank you!  This should make a big improvement in success when using text.

PR #1174, opened 2025-12-17
fix: Add tuple[int, int, int] to ExportSVG color type annotations

rawwerks:
## Summary

The `ExportSVG.add_layer()` method and `ExportSVG.__init__()` accept RGB tuples for `fill_color` and `line_color` parameters. This is documented in the docstrings and correctly handled by the internal `convert_color()` function:

```python
case tuple() as color_tuple:
    rgb_color = RGB(*color_tuple)
```

However, the public type annotations didn't include `tuple[int, int, int]`, causing type checkers to report errors for valid code like:

```python
exporter.add_layer("Hidden", line_color=(99, 99, 99), line_type=bd.LineType.ISO_DOT)
```

## How this was found

This discrepancy was discovered using [ty](https://astral.sh/blog/ty), Astral's new Python type checker (just released in beta). When type-checking code that passes RGB tuples, ty correctly reported:

```
error[invalid-argument-type]: Expected `ColorIndex | RGB | Color | None`, found `tuple[Literal[99], Literal[99], Literal[99]]`
```

## Changes

- Add `tuple[int, int, int]` to `fill_color` and `line_color` type annotations in:
  - `ExportSVG._Layer.__init__()`
  - `ExportSVG.__init__()`
  - `ExportSVG.add_layer()`
- Update internal `convert_color()` to use `tuple[int, int, int]` instead of bare `tuple` for consistency
- Update docstrings to reflect the full type union
- Reformatted with `black` per contribution guidelines

## Verification

- Ran `black --config pyproject.toml` ✓
- Ran `pylint` - score improved from 9.86 to 9.89 (fixed docstring line lengths)
- ty now reports "All checks passed!" for code using RGB tuples

### DIFF
--- a/src/build123d/exporters.py
+++ b/src/build123d/exporters.py
@@ -882,13 +882,13 @@ class ExportSVG(Export2D):
         def __init__(
             self,
             name: str,
-            fill_color: ColorIndex | RGB | Color | None,
-            line_color: ColorIndex | RGB | Color | None,
+            fill_color: ColorIndex | RGB | Color | tuple[int, int, int] | None,
+            line_color: ColorIndex | RGB | Color | tuple[int, int, int] | None,
             line_weight: float,
             line_type: LineType,
         ):
             def convert_color(
-                input_color: ColorIndex | RGB | Color | tuple | None,
+                input_color: ColorIndex | RGB | Color | tuple[int, int, int] | None,
             ) -> Color | None:
                 """
                 Convert various color representations into a `Color` object.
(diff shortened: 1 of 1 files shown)


gumyr:
`RGB` is already a  `NamedTuple`:
```py
class RGB(NamedTuple):
    """Named tuple representing an RGB color value.

    Attributes:
        r: red channel in range [0, 255]
        g: green channel in range [0, 255]
        b: blue channel in range [0, 255]
    """

    r: int
    g: int
    b: int
```
so this is a case where `ty` didn't understand that and the AI generated code for something that wasn't broken.

rawwerks:
sorry to waste your attention! 

based on the discourse conversation, i was curious to see what claude code + ty would come up with.

PR #1177, opened 2025-12-22
refactored FilletPolyline to use Polyline and Fillet2d

AlexVerschoot:
Hi,

So i refactored FilletPolyline. 
I removed a bug with non-closed fillet polylines (they need 2 fewer fillet radiuses, instead of one)

One item that's bugging me about it is that the behaviour is not clearly defined:
Take following code:
```
            p = FilletPolyline(
                (0, 0),
                (10, 0),
                (10, 10),
                (0, 10),
                radius=(1, 2),
                close=False,
            )

            p = FilletPolyline(
                (0, 0),
                (10, 0),
                (10, 10),
                (0, 10),
                radius=(1, 2,3,4),
                close=True,
                mode=Mode.PRIVATE
            )
            p.position+= (15,0,0)
```
This will generate the following (i added the the radius list index to them):
afbeelding

And i'm really wondering if we should make it so, that closed becomes like following:
afbeelding
This way, we would maintain consistency over close is or not is true.

### DIFF
--- a/src/build123d/objects_curve.py
+++ b/src/build123d/objects_curve.py
@@ -814,7 +814,6 @@ class FilletPolyline(BaseLineObject):
         close: bool = False,
         mode: Mode = Mode.ADD,
     ):
-
         context: BuildLine | None = BuildLine._get_context(self)
         validate_inputs(context, self)
         points = flatten_sequence(*pts)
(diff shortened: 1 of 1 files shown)


AlexVerschoot:
Closing this request, because i don't see a way for the refactored code to handle 3d polylines

PR #1178, opened 2025-12-25
Added project_line feature to ExtensionLine.

slobberingant:
The project_line feature allows an ExtensionLine to be projected in a direction. This is useful to provide horizontal or vertical dimensions of elements.
Tests have been added and the drafting tutorial has been modified to show this feature.


src/build123d/drafting.py
```diff
@@ -504,7 +504,7 @@ class ExtensionLine(BaseSketchObject):
         label_angle (bool, optional): a flag indicating that instead of an extracted length
             value, the size of the circular arc extracted from the path should be displayed
             in degrees. Defaults to False.
-        project_line (Vector, optional): Vector line which to project dimension against.
+        project_line (Vector, optional): Vector line which to project dimension against. Offset start point is the position of the start of border.
```
gumyr: `project_line` should be `VectorLike` as the code already handles a tuple here.

gumyr (changes requested):
I find the term `project_line` to be confusing (although technically correct).  How about `measurement_direction` or maybe `dimension_direction` instead?

In this example:
Image

the `project_line` is `(0,1,0)` and `(1,1,0)` for the left and right dimensions respectively which is the direction the extension line will follow.  By using the term `project` in there it isn't really clear what is being projected and the term `project` is overly used IMO (outside of this feature).

slobberingant:
`measurement_direction` is clearer. I think `measure_dir` is still understandable.
Let me know if you agree and I'll make the change as well as the type definition of `VectorLike`.

gumyr:
Okay, `measurement_direction` it is.  This will be a nice improvement to the dimensioning system (which still needs work).

slobberingant:
The changes have been made as discussed.

gumyr:
Thank you!

PR #1179, opened 2025-12-26
Improve documentation for introductory examples

DaWeidmann:


docs/introductory_examples.rst
```diff
@@ -275,7 +275,7 @@ create the final profile.
 
 .. _ex 9:
 
-9. Selectors, Fillets, and Chamfers
+9. Selectors, Fillets and Chamfers
```
gumyr: You don't like Oxford commas?

DaWeidmann: I don't have an opinion on them. But I prefer consistency in a document and I think either this heading or that of section 13 should be changed:

> 13. CounterBoreHoles, CounterSinkHoles and PolarLocations

If you prefer, I will adapt my PR to include them, or remove this change completely :)

gumyr: I definitely appreciate the feedback.  How about we use Oxford commas consistently if you're willing to make more changes?

DaWeidmann: Thank you for the reply. I just changed the PR to use Oxford commas.

gumyr:
Thank you!

PR #1181, opened 2025-12-29
feature: allow adding invalid meshes to Mesher

Gigahawk:
It's somewhat annoying that Mesher will sometimes prevent me from adding an invalid shape.
IMO it should be on the user to check if the mesh is valid if they absolutely need to guarantee that.

I've also noticed that these checks will report meshes are invalid even when they work fine.

Here are some files exported from b3d that throw both validity/manifold warnings (after this patch):
[export.zip](https://github.com/user-attachments/files/24372783/export.zip)
```
$ python export.py -f action_quick_release_protector.py
Cleaning export folder
Exporting sequentially
Importing action_quick_release_protector
Exporting build123d part to export/action_quick_release_protector-female.stl
/home/jasper/repos/objects/export.py:33: UserWarning: 3mf mesh is invalid
  exporter.add_shape(result)
/home/jasper/repos/objects/export.py:33: UserWarning: 3mf mesh is not manifold
  exporter.add_shape(result)
Exporting build123d part to export/action_quick_release_protector-female.step
Exporting build123d part to export/action_quick_release_protector-female.3mf
Exporting build123d part to export/action_quick_release_protector-female.brep
```

Build123d script for this part is at https://github.com/Gigahawk/objects/blob/6964d0eade729934548602dc91c04e8a2c799537/action_quick_release_protector.py

Yes I know that `add`ing the thread to the part is not recommended but I am just using this as a demonstration of a case where the outputs seem to be fine but the export will fail anyways due to the validity check failing.

### DIFF
--- a/src/build123d/mesher.py
+++ b/src/build123d/mesher.py
@@ -370,6 +370,7 @@ class Mesher:
         mesh_type: MeshType = MeshType.MODEL,
         part_number: str | None = None,
         uuid_value: UUID | None = None,
+        allow_invalid: bool = False,
     ):
         """add_shape
 
@@ -383,6 +384,7 @@ class Mesher:
             mesh_type (MeshType, optional): 3D printing use of mesh. Defaults to MeshType.MODEL.
             part_number (str, optional): part #. Defaults to None.
             uuid_value (uuid, optional): value from uuid package. Defaults to None.
+            allow_invalid (bool, optional): allow adding invalid meshes instead of throwing an error
 
         Raises:
             RuntimeError: 3mf mesh is invalid
@@ -439,7 +441,10 @@ class Mesher:
 
             # Check mesh
             if not mesh_3mf.IsValid():
-                raise RuntimeError("3mf mesh is invalid")
+                msg = "3mf mesh is invalid"
+                if not allow_invalid:
+                    raise RuntimeError(msg)
+                warnings.warn(msg, stacklevel=2)
             if not mesh_3mf.IsManifoldAndOriented():
                 warnings.warn("3mf mesh is not manifold", stacklevel=2)
 


src/build123d/mesher.py
```diff
@@ -449,7 +439,7 @@ def add_shape(
 
             # Check mesh
             if not mesh_3mf.IsValid():
-                raise RuntimeError("3mf mesh is invalid")
+                warnings.warn("3mf mesh is invalid", stacklevel=2)
             if not mesh_3mf.IsManifoldAndOriented():
```
snoyer, line 448:
```suggestion
            elif not mesh_3mf.IsManifoldAndOriented():
```
do you need the second warning if the first one has already been sent?

Gigahawk: I'm not sure, I don't really know what constitutes validity vs manifold but I have seen parts with just one and also both.

jdegenstein:
I am not sure I support this PR as-is. The backstory of Mesher/lib3mf is that the CQ implementation of 3MF support at the time (may still be true) is non-conformant with the expectation that all 3MFs are manifold -- "3MF provides a clear definition of manifoldness — it’s impossible to create a 3MF file with non-manifold edges" [1].

This is part of the reason why the native OCCT `export_stl` was left in build123d -- it allows for saving non-manifold objects.

What about making this non-manifold-is-allowed behavior optional behind a flag?

[1] https://blog.prusa3d.com/3mf-file-format-and-why-its-great_30986/

gumyr:
I agree with @jdegenstein here - 3MF is only supposed to create manifold meshes.  Is there value to having `Mesher` create non manifold STL files? Would that be substantially different than the `export_stl` option?

Gigahawk:
> I agree with @jdegenstein here - 3MF is only supposed to create manifold meshes.  Is there value to having `Mesher` create non manifold STL files? Would that be substantially different than the `export_stl` option?

I mean I'm also using it to export 3mf files. As for allowing it behind a flag, do you mean adding an optional param to `add_shape` to print a warning instead of raising an exception?

gumyr:
Creating non manifold 3MF files on purpose is invalid as per 3MF. Should the error message be changed to say something like "Invalid mesh - try export_stl instead"?

Gigahawk:
I feel like this is following the spec for compliances sake instead of for an actual practical reason. The files I've attached show that it is possible to export "invalid" parts that open with no errors, so while I do agree that it's important to notify users that the export will likely be broken, I don't see why we must explicitly forbid them from trying it at all. 

I think a possible compromise might be to refactor it so that the files get exported first before the exception gets thrown? That way the behavior remains largely the same and if users don't care about the error they can explictly catch and ignore it.

jdegenstein:
> As for allowing it behind a flag, do you mean adding an optional param to add_shape to print a warning instead of raising an exception?

Yes. Are you open to that? I want it to be very clear that the user needs to explicitly opt-in.

> I think a possible compromise might be to refactor it so that the files get exported first before the exception gets thrown?

I do NOT support that behavior because not all users watch warnings / errors. That is also why I think this needs to be opt-in at a minimum and still provide a warning.

Gigahawk:
Sounds good, I've put the warning behavior behind an `allow_invalid` parameter.

gumyr:
Thanks for working through the discussion and adding explicit opt-in behavior. We’ve decided not to merge this PR.

Mesher is primarily a 3MF exporter, and intentionally writing an invalid 3MF mesh is not a behavior we want to support, even behind an option. Some applications may accept such files, but that does not make the output valid or reliable for downstream users.

For intentionally non-manifold geometry, export_stl remains the appropriate export path. We may improve the existing invalid-3MF error message to suggest that alternative, but we will keep the current behavior of rejecting invalid 3MF meshes.

Closing with thanks again for the report, examples, and discussion.

PR #1185, opened 2025-12-31
allow exporting BREP and STEP to binary file objects

snoyer:
`OCP` is capable of exporting STEP and BREP to binary file-like streams but `build123d` currently does not allow it because of the way it checks if the destination is a stream vs a filename.

### DIFF
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -33,6 +33,7 @@ from datetime import datetime
 import warnings
 from io import BytesIO
 from os import PathLike, fsdecode, fspath
+from typing import BinaryIO
 
 import OCP.TopAbs as ta
 from anytree import PreOrderIter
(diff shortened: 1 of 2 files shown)


jdegenstein:
This is great, thank you for opening this PR. Would it be possible to add this functionality for more of the export file formats? (SVG, DXF, STL, 3MF)? I recently worked on BytesIO and have a related issue that may be helpful https://github.com/gumyr/build123d/issues/1024#issuecomment-3560267199

snoyer:
> Would it be possible to add this functionality for more of the export file formats? (SVG, DXF, STL, 3MF)?

I assume it would be possible but I haven't looked into it. This PR was only prompted by a usecase of piping out BREP and STEP files so that's the only ones I did. Would the other exporters need to be handled in this PR too?

gumyr:
Thank you! Expanding to other formats would be great but doesn't need to gate this PR.

PR #1194, opened 2026-01-09
fix: in tests use `open()` with explicit `utf-8` to avoid `cp1252` on windows

jdegenstein:
With a recent addition of an example there was a case of windows runners failing to execute them in tests correctly. Ultimately the cause of this issue was defaulting to `cp1252` in the `open` method of python on windows instead of `utf-8`. This PR makes an explicit choice of `utf-8` to avoid these issues in the future. The same issue also applied to `test_benchmarks.py` so I made the same adjustment there too.

black also changed some formatting in these files.

### DIFF
--- a/tests/test_benchmarks.py
+++ b/tests/test_benchmarks.py
@@ -4,6 +4,7 @@ from build123d import *
 from pathlib import Path
 
 from unittest.mock import Mock
+
 mock_module = Mock()
 mock_module.show = Mock()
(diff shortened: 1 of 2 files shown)


gumyr:
This looks like a great change.  That utf-8 problem was scary because I thought it was something serious.

jdegenstein:
Merged.

PR #1201, opened 2026-01-15
Generate artifacts during docs build

### LINKED ISSUE #1167: Consider removing many docs assets and rely on on-the-fly regeneration via new bernhard-42/tcv_screenshots
Given the release of https://github.com/bernhard-42/tcv_screenshots it now appears possible to generate docs assets "on the fly" in CI/CD with ocp_vscode (really it is three-cad-viewer, I think but just calling it ocp_vscode for now). This has quite a few potential benefits;

1. reduce size of repo and number/size of binary files that are tracked here over time
2. improve quality of docs assets (e.g. potentially replace SVGs in intro examples with ocp_vscode screenshots)
3. move existing screenshots of ocp_vscode to be regenerated each time docs are built (reduces work and errors when updating examples)

FYI it is almost certainly possible to run docs rebuilds (with images) locally as well. Most likely would add `tcv_screenshots` to build123d's optional `[docs]` category to streamline this.

jwagenet:
[Demo of Built Docs](https://jwagenet.github.io/build123d/html/index.html)

Process and toolchain for building documentation artifacts at sphinx build time:
- minimal doc artifact load on repo going forward
- reproducible raster imagery (.png) with tcv_screenshots, images with missing source would be recreated
- build other artifacts from files used for docs (.svg, .glb, etc)
- default settings promote a certain degree of uniformity across the docs

## Overview

Artifact creation for different cases:
- svgs as found in `general_examples`
- pngs from inline `save_model` examples as found in `topology_selection`
- pngs from `save_model` loader where whole example file is used in code sample as found in `examples`
- png thumbnails from processed `save_model`s for galleries like `ttt`
- glbs like `tea_cup` for 3D model views
- Copy required build resources like benchy stl for `benchy` example

Process has multiple entry points:
- `docs/make(.bat) html` to run build with normal config
- `py docs/build_artifacts.py -c static_config.json -l readme -d _static/assets` produces readme artifacts
- `py docs/build_artifacts.py -s /docs/objects/examples/objects_1d.py` to test output from a single file (using tools requires some path prep)
- Additional usage in `CONTRIBUTING.md`

Drawbacks as implemented:
- Increases friction for contributing changes to docs as testing artifact contributions and changes less simple than running py file
- **API changes that break an example break build process**

### Config and process

Sphinx `conf.py` calls from `build_artifacts.py` during `docs/make(.bat) html`
- Generated artifacts placed in artifact folder by default: `_build/assets`
- `asset_conf.json` to defines where and how to copy/generate artifacts from to `label`
- Config defines additional `sources` for assets/artifact generation (eg `../examples`)
- Required `resources` in sources required for an import are copied to artifact folder.
- Modules defined in `build` are "imported" to add models to tcv_screenshots queue and generate other artifacts into artifact folder (.svg, .glb, etc)
- `get_saved_models()` to feed to tcv_screenshots, crop screenshots to object, and resize with `process_images`
- From `thumbnails` make optional square thumbnails for grids
- svgs, glbs, and other binary formats are built in file
- A hash is made from sources and config to check on next run if source content has changed as a caching measure for development
- Screenshot saved models runner and caching are on a per-folder/config basis for now
- `static_conf.json` is building assets which need to be stored in repo like readme images or expensive assets

Local built assets may be cleaned with `py docs/build_artifacts.py --clean` or `./docs/make(.bat) clean` (if located in `_build` ).

### Organization

Source and destination folders are not set in stone, but I do think more rigid organization is helpful long term, but the focus of this change is just on the asset side. eg:

```
docs/
    conf.py
    top_level_topic.rst
    topic_content/
        sub_topic.rst
        examples/
            topic_example.py
            example_asset.stl
    _static/assets/
        logos/
            logo.svg
        topic_content_assets/
            toy_truck.jpg

    _build/assets/
        topic_content/
            toy_truck.png
```

### Content Changes

Some changes to image appearances:

- Use "thick" line for "visible"/"primary" svg layers in most cases to align with some previous example styles
- Most images/figures height limited to 300px and centered
- Most examples use isometric view if previously had near-iso viewpoint
- Add heart token image in summary so its available from `_build`
- Missing examples added as similarly as possible
- Various corrections to examples due to api changes or improved visibility (notably Ellipses and construction lines)

### Prebuilt (static) artifacts

Some assets cannot be built with the build process such as reference images or screenshot, however in some cases it may be necessary to prebuild artifacts
- readme images need to be available in repo
- `reference_assembly.svg` takes a *very* long time (5 min on my system) to render and causes the tcv_screenshot browser to time out, so it is prebuilt for the sake of other users
- Renders from `heat_exchanger` and `bicycle_tire` are very slow, so prebuilding is worth considering (increases current build time by >50%)

### tcv_screenshots

tcv_screenshots does the heavy lifting for making screenshots ocp_vscode-style. Allowing for sizing the viewer is a key feature that improves screenshot control. I am however not using the intended interface of the current/pinned version (`0.3.0`).
- Not using cmdline entry
- Not using `main()` paradigm
- Global `_saved_models` persists if `get_saved_models` is not called, so I am loading a longer queue of models/views to screenshot before running for each "folder"
- Customized version of `process_examples` from `tcv_screenshots.render` to make this happen: https://github.com/gumyr/build123d/blob/3dc7bd00cd34752dc79d458fa835d47718e37d38/docs/build_artifacts.py#L152C1-L182C28
- A default model config is merged in at runtime

### svg writer

- Added `write_svg` tool for docs uniformity
- `project_shapes` wraps `camera_projection` and `project_to_viewport` to produce visible and hidden layers
- Layers provided as dict with shapes and svg settings
- Line thickness is normalized to canvas size

### Other tools

`process_image`:
- Crop rendered png around foreground and resize
- Position foreground in square thumbnail and resize

`camera.camera_projection`:
- Rotate shapes to view orientation as seen from front view
- Specify ortho and iso presets + dimetric and trimetric, or
- Define viewpoint looking at shape center (or target) with horizontal coordinate system (elevation and azimuth in degrees)

## Out of scope future work

- Consider removing artifacts from git history (this one-time destructive process could be part of 1.0.0 release)
- Drive majority to doc code samples from py files

Resolves: https://github.com/gumyr/build123d/issues/1167

Tagging @jdegenstein, @bernhard-42


jwagenet:
I've completed a process that builds most doc assets as runtime artifacts. I can envision improvements but I am "done" for now.

**Pros:** reduces artifact burden on repo moving forward, png/svg content is always updated with changes to api, easier to make sweeping changes to image style
**Cons:** api changes implemented without building docs first can break docs build (gh action to check this?), increases friction for docs contributions 

### Process changes since original draft

- Prebuilt `assets` moved to `_static/assets`
- Consolidated to single build config and drive process by target config rather than target folder
- Improve svg writer

### Content changes

- Add built content from before and after draft
- Add missing asset creation scripts
- Remove built content from repo
- Some content like readme built to static

jdegenstein:
I have been following this PR for awhile and just want to say that this will take awhile for me to go through. Per our earlier conversation, I generally like the idea of building the assets in CI. Of course the reality of actually doing this, and expecting others to understand how to use it correctly can be another matter. I think some of cons you raise are worth considering carefully. 

> Consider removing artifacts from git history (this one-time destructive process could be part of 1.0.0 release)

Yes, and there is one accidental `.stp` file in the git history that contributes significantly to the size of the entire repo. I would love to find the "right opportunity" to perform this painful process for any files that are eligible for this.

This is a can of worms -- but in your opinion should we consider splitting the docs into a separate repo? The collective effort put into the docs is absolutely huge relative to the size of the codebase for this project (owing to the difficulty of documenting a CAD framework). The massive downside of this is the need to sometimes open two separate PRs, one to add a feature and another to document it.

Here is a list of the top deleted files sorted by packed size:
```
=== Deleted paths by reverse accumulated size ===
Format: unpacked size, packed size, date deleted, path name(s)
    20584065    4051111 2025-11-20 examples/assembly/RC_Buggy_2_front_suspension.stp
     1100771     345969 2023-02-09 docs/tutorial_joint.svg
      608500     267797 2023-02-09 docs/tutorial_joint_box_outer_inner_lid.svg
    27389375     188597 2025-01-22 tests/test_direct_api.py
      188263     187019 2024-04-13 docs/assets/examples/mini_nuke.png
      584837     128140 2023-02-09 docs/tutorial_joint_box_outer_inner.svg
      419463      92148 2023-02-09 docs/tutorial_joint_box_outer.svg
      190788      68115 2023-02-09 docs/tutorial_joint_outer_leaf.svg
      219823      59622 2023-07-25 docs/tutorial_lego.svg
      182090      57382 2023-07-25 docs/tutorial_step11.svg
      139792      55467 2023-07-25 docs/tutorial_step9.svg
      126522      42889 2023-02-09 docs/tutorial_joint_inner_leaf.svg
```

jwagenet:
Yes this is a very big change, so I understand this will take some time to get through. Due to the all or nothing nature of this change its hard to break into smaller chunks.

Some additional review considerations I didnt previously note:
* I haven't tested this with an actual sphinx/RTD build pipeline to know if everything works
* There may be a better home for the docs tools to make them easier to use
* Forgot to add link to [Demo Docs](https://jwagenet.github.io/build123d/html/index.html)

Regarding splitting docs to a separate repo, my gut says this sounds nice on paper: a separate review pipeline and lower weight of main repo. However, two LLMs I consulted advise otherwise for reasons including those you suggest:
* Docs PRs stay in sync with code PRs — no cross-repo coordination
* A single git blame / history covers both
* Simpler for contributors (one place, one clone)

I think for now it doesn't make sense to split the repo for the docs for the moment. This is getting OT, but some ideas to help the repo and docs contributors:
* figuring out how to remove the heavy artifacts with the lowest impact 
* add a precommit hook or gh action to monitor for adding larges file to repo, in addition to .gitignore
* gh action or something to build docs to gh-pages or something to make demo viewing easier + check for errors

gumyr:
Firstly, don't "fix" the branch conflict, I need to do some work with `make_brake_formed` to make it more predictable so in the mean time I've disabled the checks while the model is broken. The specific problem is that `make_brake_formed` creates a Face then Solid from an open Wire with no control of the Face normal/extrude direction which makes it very unpredictable. Some difference that I can't find in the way fillets work now reverse the winding order of the the offset of the open Wire, causing the Face to be inverted, causing the extrusion to happen in he wrong direction, causing the "corner" fillets to be in the middle of the part.

With respect to this PR - thanks for all the work that has gone into it. Doing something about these large images in the repo is a good objective.

I fear that the documentation is already beyond what many users can update and this type of thing will increase that complexity significantly. Maybe we need to have a contributions section to the docs that describe the documentation process and how this image building system works?

The process of removing images (and step models) from the git history worries me as well but maybe that's unfounded.

jwagenet:
@gumyr Thanks for the reminder on the CONTRIBUTING.md addition. That was part of my todo list that didn't make it on the list. I've added a section outlining basic usage which should meet the needs for most contributions.

I'm holding off on making corrections for upstream changes and other PR checks until this is reviewed to avoid extra merges.

I agree, making this process as user friendly as possible is a priority. Perhaps it makes sense to implement this change without requiring contributors to prepare artifacts this way so contributors can evaluate the process (including you and @jdegenstein) or ignore it if its too onerous, and look at improvements from there. The impetus for this change were complaints about the build artifacts in repo (and tcv_screenshots), but until there is a plan to remove those artifacts it doesn't really matter if images keep getting added to the repo. 

I believe an automated artifact builds have benefits beyond reducing repo size for reasons mentioned in the pros above (reproducibility, another vector for catching breaking changes).

bernhard-42:
@jwagenet If you could open an issue in tcv_screenshots ans explain what your preferred behavior would be, happy to adapt it.

jwagenet:
https://github.com/gumyr/build123d/pull/1376 Has related fixes to some examples that also generate assets like `spitfire`

gumyr:
Thank you for the substantial work on this PR. The effort to make documentation assets reproducible, keep generated binaries out of future commits, and improve visual consistency is appreciated.

After discussion, we have decided not to merge this approach. The documentation is already challenging for contributors to update, and generating screenshots, SVGs, GLBs, thumbnails, and related assets during the Sphinx build would add a significant amount of tooling, environment sensitivity, and failure modes to that process. The PR itself necessarily retains static/prebuilt exceptions for expensive or non-reproducible assets, so it would also leave us maintaining both paths.

There are useful ideas here that we may revisit independently: separating source and generated assets more clearly, avoiding unnecessary regeneration with caching, and potentially improving tcv_screenshots upstream. However, these do not justify adopting the complete on-the-fly artifact system at this time.

We will also keep any Git-history cleanup for large historical artifacts as a separate, carefully planned maintenance task; it should not be coupled to a documentation-build change.

Closing this PR with thanks again for the investigation and implementation.

jwagenet:
No complaints, I understand this would be a major (and likely imperfect) change to the docs process. When I have time and if there is an appetite, I may contribute separate, smaller ideas from this PR such as added/updated scripts for making assets and organization.

Issue #1198, opened 2026-01-13
glTF export and node names


voneiden:
When exporting glTF, it appears that currently only meshes receive names based on label defined in build123d. 

<img width="470" height="346" alt="Image" src="https://github.com/user-attachments/assets/c48dd1cb-8312-496d-acd2-00bafea0c442" />

This is unfortunate, because Compound assemblies that are node-only in glTF (as they do not have any associated mesh) and will therefore lose their label/name entirely. 

All nodes have instead OCCT link names like `=>[0:1:1:2]`. I did a few attempts to override this generated name, but did not succeed. Upon reviewing OCCT code I noticed that the link name is set only if auto naming is enabled ( https://github.com/Open-Cascade-SAS/OCCT/blob/bfd3fcae2ee694cfbc4105db1a86c2e2b2e9bc8a/src/DataExchange/TKXCAF/XCAFDoc/XCAFDoc_ShapeTool.cxx#L432 ). 

So I tried adding `shape_tool.SetAutoNaming_s(False)` right after the instantiation of the shape tool in [exporters3d.py](https://github.com/gumyr/build123d/blob/dev/src/build123d/exporters3d.py#L106) and to my surprise, my own labels appeared on the nodes:

<img width="479" height="350" alt="Image" src="https://github.com/user-attachments/assets/c61bb43c-82e6-4461-9c49-5e5162931b8c" />

Thoughts? I don't necessarily think that disabling the auto naming is the way forward, it might be still worth to try to investigate if those auto generated names can be overriden somehow in build123d code.

voneiden:
[seven_segment_auto_name.tar.gz](https://github.com/user-attachments/files/24600722/seven_segment_auto_name.tar.gz)

For demo purposes, attachment contains gltf+bin and glb files for both auto name enabled and disabled cases.

jdegenstein:
I tested your files in multiple software packages and the `shape_tool.SetAutoNaming_s(False)` files generated results consistent with your screenshots. I support this being added to build123d. Do you want to open a PR for this?

voneiden:
Can we think of any use case when enabling auto naming is desirable? If yes I suppose we'd want it to be configurable by the user? Also should probably check how missing labels behave for nodes when auto naming is off. I didn't check this yet and I wonder if it results in a blank node name or something else.


But I can check it later today and also roll out a PR.

voneiden:
Also need to review the impact on step export as it shares the same `_create_xde` method.

[...]

PR #1202, opened 2026-01-16
Disable auto naming for glTF export

voneiden:
Auto naming is an OCCT feature that is enabled by default. For glTF export it is problematic, because users would expect to see their labels be used on glTF nodes however with auto naming enabled OCCT overwrites the user labels.

For STEP files auto naming does not appear to overwrite any user labels.

Refs: #1198

### DIFF
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -68,7 +68,9 @@ from build123d.geometry import Location
 from build123d.topology import Compound, Curve, Part, Shape, Sketch
 
 
-def _create_xde(to_export: Shape, unit: Unit = Unit.MM) -> TDocStd_Document:
+def _create_xde(
+    to_export: Shape, unit: Unit = Unit.MM, auto_naming: bool = False
+) -> TDocStd_Document:
     """create_xde
 
     An OpenCASCADE Technology (OCCT) XDE (eXtended Data Exchange) document is a
@@ -86,6 +88,8 @@ def _create_xde(to_export: Shape, unit: Unit = Unit.MM) -> TDocStd_Document:
     Args:
         to_export (Shape): object or assembly
         unit (Unit, optional): shape units. Defaults to Unit.MM.
+        auto_naming (bool, optional): whether to use ShapeTool AutoNaming.
+            Defaults to False.
 
     Returns:
         TDocStd_Document: XDE document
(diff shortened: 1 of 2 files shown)


src/build123d/exporters3d.py
```diff
@@ -234,7 +241,7 @@ def export_gltf(
             node.mesh(linear_deflection, angular_deflection)
 
     # Create the XCAF document
-    doc: TDocStd_Document = _create_xde(to_export, unit)
+    doc: TDocStd_Document = _create_xde(to_export, unit, auto_naming=True)
```
jdegenstein: It looks like this may be swapped gltf/step :: true/false instead of false/true. Can you check?

voneiden: Oh boy, I had one job. Good catch, thank you.

src/build123d/exporters3d.py
```diff
@@ -86,6 +88,7 @@ def _create_xde(to_export: Shape, unit: Unit = Unit.MM) -> TDocStd_Document:
     Args:
         to_export (Shape): object or assembly
         unit (Unit, optional): shape units. Defaults to Unit.MM.
+        auto_naming (bool, optional): whether to use ShapeTool AutoNaming
```
jdegenstein: Just noticed this, should read this way when a default is set:
`auto_naming (bool, optional): whether to use ShapeTool AutoNaming. Defaults to False.`

jdegenstein:
This is a larger point (I don't expect you to address in this PR) but I think we need to consider how to do more round trip testing for these import/export formats. In the case of this change we don't have an import GLTF and STEP is functionally not changing -- so there really is no simple way (that I am aware of) to check that the exported labels are represented in the GLTF itself.

In summary once you make the small docstring change above, this looks good to me.

voneiden:
@jdegenstein Fair point. I believe this particular PR scope would be fairly trivial to test: export_gltf with binary=False creates a JSON file that could be loaded and the post export node labels could be then asserted in a test. I can add such a test if you have no objections to the approach.

src/build123d/exporters3d.py
voneiden: Updated, thanks.

voneiden:
Actually I see there is already a test https://github.com/gumyr/build123d/blob/203f8bfc909d7762e1df5d03b9b0d40272241edf/tests/test_exporters3d.py#L176-L186  that asserts that labels exist in meshes, that could be extended to assert the nodes as well. Perhaps wrap the Box into a Compound to prove that Compound labels persist as well.

gumyr:
This PR looks good to me.  Is it ready to merge now?

voneiden:
Yes, good to go on my behalf. Thanks.

gumyr:
Thank you!

jdegenstein:
Thank you for this, will be a nice addition

PR #1212, opened 2026-01-21
.gitignore -> add stp, STP, brep, gltf, bin, glb

jdegenstein:
This is to avoid the issue where a large stp file was accidentally committed. I have also added other file formats as I think it makes sense to avoid committing large binary files. This can always be overridden by e.g. `git add --force somefile.glb` if necessary for e.g. use in the docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,18 +16,25 @@ docs/_build/
 # User generated debris
 *.swp
 *.log
+
+# User generated debris (3D)
 *.stp
 *.STP
 *.step
 *.STEP
 *.stl
-*.svg
-*.dxf
 *.3mf
 *.brep
 *.gltf
 *.bin
 *.glb
+
+# User generated debris (2D)
+*.dxf
+*.jpeg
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -16,12 +16,18 @@ docs/_build/
...
 *.step
 *.STEP
 *.stl
 *.svg
 *.dxf
 *.3mf
+*.brep
+*.gltf
+*.bin
+*.glb
```
gumyr, line 30: Good idea. How about adding some image file types in here like: png, jpg, jpeg, svg, etc.?

jdegenstein:
Another thing worth noting is this:
```
# Allow (un-ignore) .png files inside the /docs/ folder and its subfolders
!docs/**/*.png
```
The idea is that we can blanket ignore *.png and then un-ignore within a certain (recursive) subfolder.

gumyr:
Cool, looks good to me.

Issue #987, opened 2025-05-12
Solid.extrude_taper might flip the direction or extrusion


gumyr:
This code shows that the `LocOpe_DPrism` call in `Solid.extrude_taper` can invert the direction of extrusion.
```py
from build123d import *
from ocp_vscode import *

from dataclasses import dataclass, field
from typing import Tuple, Optional


@dataclass
class BaseCap:
    width: float
    length: float
    height: float
    wall: float
    roof: float
    taper: float = 0
    stem_rotation: float = 0
    align: Tuple[Align, Align] = (Align.CENTER, Align.CENTER)
    mode: Mode = Mode.ADD

    _part: Optional[Part] = field(init=False, default=None, repr=False)

    left: Optional[Edge] = field(init=False, default=None)
    right: Optional[Edge] = field(init=False, default=None)
    top: Optional[Edge] = field(init=False, default=None)
    bottom: Optional[Edge] = field(init=False, default=None)

    @property
    def part(self) -> Part:
        if self._part is None:
            self._part = self._build().mirror(Plane.XY)
        return self._part

    @part.setter
    def part(self, value: Part) -> None:
        self._part = value

    def _build(self) -> Part:
        with BuildPart() as bp:
            sketch = self._draw()
            print(sketch.face().normal_at())
            extrude(sketch, amount=self.height, taper=self.taper)
        return bp.part

    def _draw(self) -> Sketch:
        raise NotImplementedError


@dataclass
class AsymmetricCap(BaseCap):
    angle: float = 0
    flip: bool = False

    def _draw(self) -> Sketch:
        d = -1 if self.flip else 1

        with BuildSketch() as sk:
            with BuildLine():
                self.right = Line((0, 0), (0, -self.length))
                self.bottom = PolarLine(
                    start=self.right @ 1,
                    length=self.width * d,
                    angle=180 + self.angle * d,
                )
                self.left = PolarLine(
                    start=self.bottom @ 1,
                    length=(self.right @ 0).Y - (self.bottom @ 1).Y,
                    angle=90 + self.angle * d,
                    length_mode=LengthMode.VERTICAL,
                )
                self.top = Line(self.left @ 1, self.right @ 0)
            make_face()

        if self.flip:
            self.left, self.right = self.right, self.left

        return sk.sketch


c1 = AsymmetricCap(
    width=10,
    length=20,
    height=5,
    wall=2,
    roof=1,
    angle=4,
    taper=1,
)

c2 = AsymmetricCap(
[...]

Issue #746, opened 2024-10-21
Consider using dprism for extrude_taper instead of loft


gumyr:
The surface of the sphere is this example is not correct; however, using a Cone does work.  
```py
from build123d import *
from ocp_vscode import show_all

ball_dia = 100
base_dia = 90
base_thickness = 10
base_fillet = 5
base_chamfer = 5
hole_dia = 20
hole_taper = 2

with BuildPart() as ball:
    Sphere(ball_dia / 2)
    with Locations((0, 0, -ball_dia / 2)):
        Cylinder(
            base_dia / 2, base_thickness, align=(Align.CENTER, Align.CENTER, Align.MIN)
        )
    top_surface_coord = -ball_dia / 2 + base_thickness
    fillet_edge, chamfer_edge = (
        ball.edges(Select.LAST)
        .filter_by_position(Axis.Z, top_surface_coord, top_surface_coord)
        .sort_by(SortBy.RADIUS)
    )
    fillet(fillet_edge, base_fillet)
    chamfer(chamfer_edge, base_chamfer)
    # with Locations(Plane.XZ.offset(ball_dia / 2)):
    #     Cone(
    #         hole_dia * 0.5,
    #         hole_dia * 0.4,
    #         ball_dia,
    #         mode=Mode.SUBTRACT,
    #         align=(Align.CENTER, Align.CENTER, Align.MAX),
    #     )
    with BuildSketch(Plane.XZ) as hole_sketch:
        Circle(hole_dia / 2)
    extrude(
        hole_sketch.sketch,
        # until=Until.LAST,
        amount=ball_dia,
        # both=True,
        mode=Mode.SUBTRACT,
        # mode=Mode.INTERSECT,
        taper=hole_taper,
    )
    extrude(
        hole_sketch.sketch,
        amount=-ball_dia,
        mode=Mode.SUBTRACT,
        taper=-hole_taper,
    )
```
![image](https://github.com/user-attachments/assets/6e3538c0-d370-42bb-94b8-2920b1bb4257)

The `dprism` method might be a good alternative to the loft currently used in `extrude_taper`.

jdegenstein:
Here is a comparison @jrmobley made a long time ago when we were discussing something related. The differences in behavior are not intuitively obvious between the two methods.

![image](https://github.com/user-attachments/assets/e4563623-ca80-45b2-a583-fcea0276632a)

EDIT: give proper credit to jrmobley

Issue #568, opened 2024-02-26
Tapered extrude only works with N<=2 polar locations despite equivalent topology


MatthiasJ1:
```py
N = 2 # ok
N = 3 # fails
with BuildPart() as p:
    with BuildSketch() as s:
        with PolarLocations(0, N, 0, 180):
            Rectangle(100, 1)
    extrude(amount=-2, taper=-15)
```

Oddly, if you fillet the outside vertices even slightly, this will work:
```py
N = 20
with BuildPart() as p:
    with BuildSketch() as s:
        with PolarLocations(0, N, 0, 180):
            Rectangle(100, 1)
        fillet(s.vertices().sort_by_distance((0,0,0))[N*2:], 0.1)
    extrude(amount=-2, taper=-15)
```

gumyr:
The bug seems to be deep within the OCCT BRepOffsetAPI_MakeOffset code.  This is the result of the offset:
![image](https://github.com/gumyr/build123d/assets/85638518/617d337f-4686-48a8-ac39-fb0b452f4fe2)
which isn't very useful for the creation of the taper.

Issue #553, opened 2024-02-17
Extrude faces from SVG sometimes changes direction when applying taper


qwelyt:
```py
import build123d as bd

svg = bd.import_svg("bd_test.svg")
# taper = 0
taper = 10

with bd.BuildPart() as prt:
    with bd.BuildSketch():
        for k in svg:
            with bd.BuildLine():
                bd.add(k)
            bd.make_face()
    bd.extrude(amount=10, taper=taper, dir=bd.Axis.Z.direction)

show(prt)
```
Depending on what shapes are in the imported SVG, some objects change direction when applying a taper.  

This is without taper. Things extrude as expected.
![20240218_005027](https://github.com/gumyr/build123d/assets/3884820/efa55cbd-2c90-4e09-8be2-bd48e60b71a9)

When applying a taper primitives like `rect` and `cylinder` change direction of the extrusion while `path` behaves like they should.
![20240218_005046](https://github.com/gumyr/build123d/assets/3884820/27ff9ee2-6176-498f-a0d8-3173c7684bbb)
![bd_test](https://github.com/gumyr/build123d/assets/3884820/b9a2570f-62c9-428f-9079-4f083e45ced7)

gumyr:
Thanks for isolating the problem. After removing all of the Python code that I can, here is what happens:
```py
from build123d import *
from ocp_vscode import show
from OCP.LocOpe import LocOpe_DPrism
from math import radians, cos

svg = import_svg("issue-553.svg")
face_to_extrude = -Face(svg[2])
taper = 10

prism_builder = LocOpe_DPrism(
    face_to_extrude.wrapped,
    10 / cos(radians(taper)),
    radians(taper),
)
new_solid = Solid(prism_builder.Shape())

show(face_to_extrude, new_solid)
```
![image](https://github.com/gumyr/build123d/assets/85638518/936e7130-5ef7-4adc-a474-066e08888c3c)
the extrude is in the wrong direction and this direction is independent of the face normal (same result with `face_to_extrude = Face(svg[2])`.

Need to figure out why `LocOpe_DPrism` is flipping the direction or avoid using this class at all.

gumyr:
There are a couple solutions to this.  One is this:
```py
svg = import_svg("issue-553.svg")
profile = Face(svg[2])
taper = 10

taper_law = Law_Linear()
taper_law.Set(0.0, 1.0, 1.0, 1 - sin(radians(taper)))
outer = profile.outer_wire().wrapped

[...]

PR #1221, opened 2026-02-03
`Solid.extrude_taper` switch to `Solid.draft` approach instead of `LocOpe_DPrism`

jdegenstein:
The current approach for `Solid.extrude_taper` has a few issues:

1. Falls back to a loft approach in several common cases; negative taper angle and profile has holes
 2. The loft fallback often generates complex surfaces that affect the stability of downstream operations
 3. The LocOpe_DPrism approach seems to have its own issues such as https://github.com/gumyr/build123d/issues/987
 
This PR alters the approach from LocOpe_DPrism + loft fallback to `Solid.draft` + loft fallback. The internal method works by first creating a linear extrusion and then selecting the "side faces" and applying a draft angle to those. Here are the reasons why this may be a better approach than the current one:

1. Draft seems to have no issues with reasonable positive and negative taper angles
2. Draft seems to support some types of holes (composed of lines or circular arcs).
3. When it is used draft seems generates "clean" results, i.e. planar, cylindrical, conical faces instead of BSpline surfaces (loft)
4. I verified that this fixes issue #987 

Self Critique: I am wondering if rather than `Solid.draft` + loft fallback it should instead be: `Solid.draft` + LocOpe_DPrism/dprism fallback-1 + loft fallback-2. Also -- maybe I should also add some new tests to cover the issue 987 or 553 failures.

Related issues:
(linked above) #987 
#553  (similar to the above 987)
https://github.com/gumyr/build123d/issues/543 (need to check how corners are affected in this new approach)
https://github.com/gumyr/build123d/issues/568 (likely an issue in offset, not taper extrude)
https://github.com/gumyr/build123d/issues/746 (alternative approach proposed using dprism instead of draft)

### DIFF
--- a/src/build123d/topology/three_d.py
+++ b/src/build123d/topology/three_d.py
@@ -940,14 +940,13 @@ class Solid(Mixin3D[TopoDS_Solid]):
 
         Extrude a cross section into a prismatic solid in the provided direction.
 
-        Note that two difference algorithms are used. If direction aligns with
-        the profile normal (which must be positive), the taper is positive and the profile
-        contains no holes the OCP LocOpe_DPrism algorithm is used as it generates the most
-        accurate results. Otherwise, a loft is created between the profile and the profile
-        with a 2D offset set at the appropriate direction.
+        Note that two different algorithms are used. If direction aligns with
+        the profile normal (which must be positive), the OCP BRepOffsetAPI_DraftAngle
+        algorithm is tried. Otherwise, a loft is created between the profile and the
+        profile with a 2D offset set at the appropriate direction.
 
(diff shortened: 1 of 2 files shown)


MatthiasJ1:
Awesome! It doesn't quite address the linked issue but it at least makes the behavior consistent (before the type of extrusion would yield either `arc` or `intersection` results as shown below).

I think it would be great if we add the `Kind` parameter where:
- `kind=Intersect` (default) => `draft` with fallback of `loft<Intersect>`
- `kind=Arc` => `prism` with fallback of `loft<Arc>`

out

PR #1227, opened 2026-02-06
feat: enable STL rescale in `import_stl`, add tests

### LINKED ISSUE #874: Feature Request: add support for scaling imported STLs
Most 3d scans and software like Blender use meters as the base unit while B3D uses mm. This causes most models to be imported way too small. This can be fixed by either pre-processing which is an unnecessary extra step that needs to be done every time, or by using Mesher.import() which can take minutes to import the model.

jdegenstein:
Closes https://github.com/gumyr/build123d/issues/874

Thanks to @snoyer for the suggestion on performance improvement at the above.


gumyr (approved):
Looks good

PR #1234, opened 2026-02-18
fix: ShapeList.filter_by callable argument should match list type

tobywf:
Changes `ShapeList.filter_by` to take a `Callable[[T], bool]`.

Currently, `filter_by` takes a `ShapePredicate`, where the argument is just a `Shape`:

https://github.com/gumyr/build123d/blob/cb155f79d18f8247eb8e951617d0a3056be0ceec/src/build123d/topology/shape_core.py#L2657-L2662

Unlike `sort_by`:

https://github.com/gumyr/build123d/blob/cb155f79d18f8247eb8e951617d0a3056be0ceec/src/build123d/topology/shape_core.py#L2949-L2953

This means the following example passes type-checking:

```python
from build123d import *

def filter_face(_face: Face) -> bool:
    return False

def filter_vertex(_vertex: Vertex) -> bool:
    return False

with BuildSketch() as sk:
    Rectangle(1, 1)
    v = sk.vertices().filter_by(filter_vertex)
    v = sk.vertices().filter_by(filter_face)  # should fail type checking
```

I think `group_by` has a similar issue?


snoyer:
> I think `group_by` has a similar issue?

I think so too.

I was the one who added `ShapePredicate` way back in #177 and either it made sense at the time or I didn't know any better, but I'm not sure it's still necessary now... there's also been a big refactor in the meantime that may have affected thing? @gumyr do you know of any reason to keep `ShapePredicate`?


Also worth mentioning that fixing these types doesn't just catches errors like in @tobywf's example, it also allows inferring the argument types in lambdas, for example in:
```py
Box(1,2,3).edges().group_by(lambda e: e.length)
```
`e` would be inferred as `Edge` and `e.` would have auto-completion and whatnot.

gumyr:
It seems that all three should be unified/fix and I don't see the need for `ShapePredicate`. A LLM corrected my attempt and came up with the following as the most correct:
```py
def sort_by(
    self,
    sort_by: Callable[[T], K] | Axis |  Edge | Wire | SortBy | property = Axis.Z,
    reverse: bool = False,
) -> ShapeList[T]:

def group_by(
    self,
    group_by: Callable[[T], K] | Axis | Edge | Wire | SortBy | property = Axis.Z,
    reverse=False,
    tol_digits=6,
) -> GroupBy[T, K]:

def filter_by(
    self,
    filter_by: Callable[[T], bool] | Axis | Plane | GeomType | property,
    reverse: bool = False,
    tolerance: float = 1e-5,
) -> ShapeList[T]:
```
If you agree, would you mind expanding the scope of the PR to all three?

snoyer:
Alternatively, (assuming the CI passes) this PR could be merged as-is to only remove `ShapePredicate`, and changing `Shape` to `T` in `group_by` and `sort_by` could be a separate one

tobywf:
Thanks for taking a look! It's been a few months. I'm quite busy at the moment, but can take a look in a few weeks. Alternatively, feel free to modify the PR.

gumyr:
Thank you!

PR #1240, opened 2026-02-24
Update README.md

rexnfx:
the line edges need to be passed to the make_hull() or else you get a OCP.OCP.Standard.Standard_TypeMismatch: TopoDS::Edge error


gumyr:
Sorry, I fixed it before seeing your PR. Thanks for doing this though.

PR #1248, opened 2026-03-09
Add type annotations to Builder.__enter__()

DaWeidmann:
I recreated this PR to clean up the history...

The explicit type annotation helps additional tooling like language servers to correctly determine the type in context manager.

For me, this solves a problem with the ty language server, which was unable to determine the types BuildPart, BuildSketch and BuildLine context managers.


gumyr:
Thank you.

Issue #1258, opened 2026-03-23
`add` results inconsistent across Compound subtypes


jwagenet:
Adding an Edge (or Wire) to Curve produces a type output inconsistent with adding the corresponding same-`dim` shape to Sketch or Part. 

Additionally, one might expect `Compound.__add__` maintains Compound subtype if child Shapes the same `dim`.

**Current**
| Case | `Curve` | `Sketch`/`Part` |
| :--- | :---: | ---: |
|  `+ Shape` | `Edge` | `Compound` |
|  `+ list[Shape]` | `Edge` | `Compound` |
|  `+ Shape + Shape` | `ShapeList` | `Compound` |
|  `+ list[Shape, Shape]`  | `Curve` | `Compound` |

```py
e1 = Line((0,.5),(6.25/2,.5))
f1 = Rectangle(.1, .1)
s1 = Box(.1, .1, .1)
print(type(Curve() + e1)) # Edge
print(type(Curve() + [e1])) # Edge
print(type(Curve() + e1 + Pos(Y=1)*e1)) # ShapeList
print(type(Curve() + [e1,  Pos(Y=1)*e1])) # Curve <- The only case producing expected output
print(type(Sketch() + f1)) # Compound
print(type(Sketch() + [f1])) # Compound
print(type(Sketch() + f1 + Pos(Y=1)*f1)) # Compound
print(type(Sketch() + [f1, Pos(Y=1)*f1])) # Compound
print(type(Part() + s1)) # Compound
print(type(Part() + [s1])) # Compound
```

**Expected**
```py
print(type(Curve() + e1)) # Curve
...
print(type(Sketch() + f1)) # Sketch
...
print(type(Part() + s1)) # Part
...
```

jwagenet:
Initial issue impetus: https://discord.com/channels/964330484911972403/1074840524181217452/1485660122175897650
```py
m1 = Line((0,.5),(6.25/2,.5))
c = Curve() + m1 + Pos(Y=1)*m1 + Pos(Y=2)*m1 # Produces ShapeList instead of Curve due to first add
trimmer = Cylinder(6.25/2-.25, 1)
c_trim = trimmer & c # empty compound
```

Related discussion for non-touching Shape additions (not relevant if order of operations is preserved)
https://discord.com/channels/964330484911972403/1074840524181217452/1470949238484766894

Related discussion for 1D objects having special consideration for retaining properties (explicit conversion to Curve should come with this tradeoff)
[...]

PR #1259, opened 2026-03-24
Ensure add returns consistent results across Compounds

bernhard-42:
A fix attempt for https://github.com/gumyr/build123d/issues/1258 (add results inconsistent across Compound subtypes)

This has been analysed together with claude code:

**Solution approach:**

- one_d.py - `Mixin1D.__add__`

    Solves `Curve() + edges => Curve`

    `Mixin1D.__add__` (used by `Curve`) was designed for `Edge`/`Wire`. It would unwrap single-edge results and return bare `Edge` objects, losing the `Curve` type. The fix checks if `self` is a `Compound` type (`order == 4`) and ensures the result is always re-wrapped into the correct type (e.g., `Curve`).

- composite.py - `Compound.__add__`
    
    Solves `Part() + solids => Part` and `Sketch() + faces => Sketch`
    
    The general `__add__` path (for `Sketch`/`Part`) always created bare `Compound` objects regardless of the caller's type. The fix resolves the target class from `self._dim ({1: Curve, 2: Sketch, 3: Part}`, falling back to `Compound` for mixed dimensions) and uses that class for all wrapping. 

**Test impact:**

2 tests used 
- `isinstance(result, Edge)` which now is a `Curve` 
- `length` property on the Edge (now Curve). Curve is a `Mixin3D`, so it has `volume` but not `length`. `length` could have been added to `Compound` (leading to a `length` setter issue with `Box`). Hence `length` was added to `Curve` (and for completeness `area` to `Sketch`). btw. `volume` cannot easily be moved to `Part`.

An alternative solution would be to remove `length` and `area` again and fix the tests to use `sum(e.length for l.edges()`

### DIFF
--- a/src/build123d/topology/composite.py
+++ b/src/build123d/topology/composite.py
@@ -59,6 +59,7 @@ import warnings
 from collections.abc import Iterable, Iterator, Sequence
 from itertools import combinations
 from os import PathLike, fspath
+from typing import cast as tcast
 from typing_extensions import Self
 
 import OCP.TopAbs as ta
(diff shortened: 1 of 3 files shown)


bernhard-42:
Interesting, mypy locally returns "Success: no issues found in 37 source files"

Before spending more time on this, I wait whether the PR is relevant at all.

gumyr:
Given the current state of Issue #1258 this PR is no longer required. Thanks for attempting to solve the problem.

PR #1262, opened 2026-03-25
Add Axis and angle constructor overload in Rotation

### LINKED ISSUE #1253: Add Axis & angle to Rotation class constructor
To better align the `Rotation` class with the `rotate` method a `Rotation(Axis, angle)` constructor overload should be added.

DaWeidmann:
My extension to the parsing of the arguments in `Rotation.__init__()` does not feel very robust / elegant... If you have any suggestions on how to improve it, e.g. align it with `Location.__init__()`, I can rework it. If it is fine as is, I am happy as well :smile: 

From CONTRIBUTING.md
- [x] Run tests with: python -m pytest -n auto
- [x] Check added files' style with: pylint <path/to/file.py>
- [x] Check added files' type annotations with: mypy <path/to/file.py>
- [x] Run black formatter against files' changed: black <path/to/file.py>

Fixes #1253

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -2402,44 +2402,69 @@ class Rotation(Location):
         """Rotation from axis of rotation and angle."""
 
     def __init__(self, *args, **kwargs):
-        if not all(
-            key in ("X", "Y", "Z", "rotation", "ordering", "axis", "angle")
-            for key in kwargs
-        ):
-            raise TypeError("Invalid key for Rotation")
-        angles, rotations, orderings = [0, 0, 0], [], []
-        trsf = None
+        rotation = kwargs.pop("rotation", None)
+        x_angle = kwargs.pop("X", 0.0)
+        y_angle = kwargs.pop("Y", 0.0)
+        z_angle = kwargs.pop("Z", 0.0)
+        ordering = kwargs.pop("ordering", Intrinsic.XYZ)
+        axis = kwargs.pop("axis", None)
+        axis_angle = kwargs.pop("angle", None)
+
+        # If any unexpected kwargs remain
+        if kwargs:
+            raise TypeError(f"Unexpected keyword arguments: {', '.join(kwargs)}")
+
+        # Fill from positional args if not given via kwargs
         if args:
-            angles = list(filter(lambda item: isinstance(item, (int, float)), args))
-            vectors = list(filter(lambda item: isinstance(item, Vector), args))
-            tuples = list(filter(lambda item: isinstance(item, tuple), args))
-            axis_items = list(filter(lambda item: isinstance(item, Axis), args))
-            if tuples and not axis_items:
-                angles = list(*tuples)
-            if vectors:
-                angles = tuple(vectors[0])
-            if axis_items and angles:
-                angle = angles[0]
-                trsf = gp_Trsf()
-                trsf.SetRotation(axis_items[0].wrapped, angle)
-            if len(angles) < 3:
-                angles.extend([0.0] * (3 - len(angles)))
-            rotations = list(filter(lambda item: isinstance(item, Rotation), args))
-            orderings = list(
-                filter(lambda item: isinstance(item, (Extrinsic, Intrinsic)), args)
-            )
-        kwargs.setdefault("X", angles[0])
-        kwargs.setdefault("Y", angles[1])
-        kwargs.setdefault("Z", angles[2])
-        kwargs.setdefault("ordering", orderings[0] if orderings else Intrinsic.XYZ)
-        if rotations:
-            super().__init__(rotations[0])
-        elif trsf:
+            if rotation is None and isinstance(args[0], Rotation):
+                rotation = args[0]
+
+            elif axis is None and isinstance(args[0], Axis):
+                # Check for correct args
+                if len(args) < 2:
+                    raise TypeError(f"Too few arguments: Requires Axis and Angle")
+                elif not isinstance(args[1], (int, float)):
+                    raise TypeError(f"Angle must be a float not {args[1]}")
+
+                axis, axis_angle = args[0], args[1]
+
+                if axis_angle == 0.0:
+                    # No valid rotation: Fallback mode
+                    axis, axis_angle = None, None
+
+            else: # Euler angles
+                euler = list(filter(lambda item: isinstance(item, (int, float)), args))
+                vectors = list(filter(lambda item: isinstance(item, Vector), args))
+                tuples = list(filter(lambda item: isinstance(item, tuple), args))
+
+                if tuples:
+                    euler = list(*tuples)
+                if vectors:
+                    euler = tuple(vectors[0])
+
+                # Extract individual angles
+                x_angle = euler[0] if len(euler) > 0 else x_angle
+                y_angle = euler[1] if len(euler) > 1 else y_angle
+                z_angle = euler[2] if len(euler) > 2 else z_angle
+
+                ord_arg = next(
+                    filter(lambda item: isinstance(item, (Extrinsic, Intrinsic)), args),
+                    None,
+                )
+                if ord_arg:
+                    ordering = ord_arg
+
+        # Construct Rotation
+        if rotation:
+            super().__init__(rotation)
+
+        elif axis:
+            trsf = gp_Trsf()
+            trsf.SetRotation(axis.wrapped, axis_angle)
             super().__init__(trsf)
+
         else:
-            super().__init__(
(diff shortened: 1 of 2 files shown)


src/build123d/geometry.py
```diff
@@ -2396,18 +2396,36 @@ def __init__(
...
+        ):
             raise TypeError("Invalid key for Rotation")
         angles, rotations, orderings = [0, 0, 0], [], []
+        trsf = None
         if args:
             angles = list(filter(lambda item: isinstance(item, (int, float)), args))
             vectors = list(filter(lambda item: isinstance(item, Vector), args))
             tuples = list(filter(lambda item: isinstance(item, tuple), args))
-            if tuples:
+            axis = list(filter(lambda item: isinstance(item, Axis), args))
```
gumyr: Although mypy seems happy axis is defined as type Axis in the overload and type list here - how about using a different name like:
```py
axis_items = list(filter(lambda item: isinstance(item, Axis), args))
...
if axis_items and angles:
    angle = angles[0]
    trsf = gp_Trsf()
    trsf.SetRotation(axis_items[0].wrapped, angle)
```

gumyr (changes requested):
It took quite a few iterations to get to what I think are better constructors than this one (before your change) but I think we can leave it for now.

src/build123d/geometry.py
DaWeidmann: Done.

src/build123d/geometry.py
```diff
@@ -2396,18 +2396,36 @@ def __init__(
         """Subclass of Location used only for object rotation
         ordering is for order of rotations in Intrinsic or Extrinsic enums"""
 
+    @overload
+    def __init__(
+        self,
+        axis: Axis,
+        angle: float,
+    ):
+        """Subclass of Location used only for object rotation
```
gumyr, line 2405: This docstring doesn't really describe how the Axis and angle works.

src/build123d/geometry.py
```diff
@@ -2396,18 +2396,36 @@ def __init__(
...
             angles = list(filter(lambda item: isinstance(item, (int, float)), args))
             vectors = list(filter(lambda item: isinstance(item, Vector), args))
             tuples = list(filter(lambda item: isinstance(item, tuple), args))
-            if tuples:
+            axis_items = list(filter(lambda item: isinstance(item, Axis), args))
+            if tuples and not axis_items:
                 angles = list(*tuples)
             if vectors:
                 angles = tuple(vectors[0])
+            if axis_items and angles:
```
gumyr, line 2425: If the user provides only an axis or angle this is skipped with no warning. How about checking if they are both present if one of them is and providing the user an error message like "both an Axis and angle must be provided"?

src/build123d/geometry.py
```diff
@@ -2396,18 +2396,36 @@ def __init__(
...
+                angle = angles[0]
```
gumyr, line 2426: If the angle is zero it might be safer/faster to skip the whole thing. Not a big deal though.

src/build123d/geometry.py
```diff
@@ -2396,18 +2396,36 @@ def __init__(
...
+                trsf = gp_Trsf()
+                trsf.SetRotation(axis_items[0].wrapped, angle)
```
gumyr, line 2428: It seems like the design is very close to allowing multiple Axis/angle options. Do you think it would be useful to have the interface support multiple pairs? I don't know, seems like it makes the API too complex when multiple Rotations can be combined. What do you think?

gumyr (changes requested):
Thanks - this will be a nice enhancement.

src/build123d/geometry.py
DaWeidmann: With the latest changes, the docstrings of Rotation.__init__() should be more clear.

src/build123d/geometry.py
DaWeidmann: I will implement this and the next change. For me it seems the constructor is getting more complicated than it needs to be, so I will rewrite it to the style of the Location constructor (first, extract all kwargs, then handle the args and after that implement the initialization logic). This is why the PR is now back in Draft status.

src/build123d/geometry.py
DaWeidmann: For me, it feels more natural to chain rotations by multiplying them (which the API already supports I think?). Allowing it in the constructor would make it more complex without making the process easier in my opinion.
Do you have an example in mind, where a constructor supporting multiple Axis/angle pairs would be useful?

gumyr: > For me, it feels more natural to chain rotations by multiplying them (which the API already supports I think?). Allowing it in the constructor would make it more complex without making the process easier in my opinion. Do you have an example in mind, where a constructor supporting multiple Axis/angle pairs would be useful?

Agreed, let's stick with a single angle/axis pair.

src/build123d/geometry.py
```diff
@@ -2393,37 +2391,80 @@ def __init__(
...
+                if ord_arg:
+                    ordering = ord_arg
+
+        # Construct Rotation
+        if rotation:
+            super().__init__(rotation)
+
+        elif axis:
+            trsf = gp_Trsf()
+            trsf.SetRotation(axis.wrapped, axis_angle)
```
```suggestion
            trsf.SetRotation(axis.wrapped, radians(axis_angle))
```

DaWeidmann: Thank you. I noticed that in `Matrix.rotate()` (where I took the code from), the angle parameter is also not converted to radians. Is it assumed to be in radians in that case or should this be fixed as well?

DaWeidmann: I am talking about `geometry.py`, line 2575 specifically.

snoyer: All the user-facing angles seem to be in degrees, including `Vector.rotate(axis, angle)` https://github.com/gumyr/build123d/blob/6d73849d4aa59c3ea3c8d59a50a6d8bdd32a082d/src/build123d/geometry.py#L538-L545

but `Matrix.rotate(axis, angle)`'s doc string doesn't specify the unit https://github.com/gumyr/build123d/blob/6d73849d4aa59c3ea3c8d59a50a6d8bdd32a082d/src/build123d/geometry.py#L2523-L2524
and the method is tested to take in radians https://github.com/gumyr/build123d/blob/6d73849d4aa59c3ea3c8d59a50a6d8bdd32a082d/tests/test_direct_api/test_matrix.py#L118-L146

So that's a question for @gumyr

src/build123d/geometry.py
```diff
@@ -2354,37 +2352,89 @@ def __init__(
...
+
+                ord_arg = next(
+                    filter(lambda item: isinstance(item, (Extrinsic, Intrinsic)), args),
+                    None,
+                )
+                if ord_arg:
+                    ordering = ord_arg
+
+        # Construct Rotation
+        if rotation:
```
gumyr: If the user creates a `Rotation` with `Rotation(rotation=(10, 20, 30))` then `rotation` is set to `(10,20,30)` and passed to `Location` which will convert them to positions. I believe the goal here was that `rotation` is always of calls `Rotation`/`Location` as this point.

src/build123d/geometry.py
```diff
@@ -2354,37 +2352,89 @@ def __init__(
...
+            super().__init__(rotation)
+
+        elif axis:
+            trsf = gp_Trsf()
+            trsf.SetRotation(axis.wrapped, radians(axis_angle))
```
gumyr: If the user does `Rotation(axis=Axis.Z)` then `axis_angle` will be `None` at this point and fail. It would be good if the keyword path also checked for a valid float before this call.

gumyr (changes requested):
Thanks - this is looking good but I think there are a couple ways a user can mess up - see below.

src/build123d/geometry.py
gumyr: Matrix.rotate incorrectly uses radians and should be changed to use degrees like every other build123d API.

DaWeidmann: I opened a new issue to track this, see #1310

src/build123d/geometry.py
DaWeidmann: Ah, I misunderstood the use of this constructor. This is also why I removed the ordering from the declaration. I think my current changes make the distinction between those two cases clearer.

src/build123d/geometry.py
DaWeidmann: Yes, this was an oversight on my part. It should be fixed now.

gumyr:
Thank you!

PR #1263, opened 2026-03-28
Add constructor for BoundBox from TopoDS_Shape

### LINKED ISSUE #1010: BoundBox.from_topo_ds should be integrated into the constructor
The BoundBox class should be able to accept a Shape in its constructor and not need the `from_topo_ds` method.

DaWeidmann:
Resolves #1010.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1020,7 +1020,51 @@ class Axis(metaclass=AxisMeta):
 class BoundBox:
     """A BoundingBox for a Shape"""
 
+    @overload
     def __init__(self, bounding_box: Bnd_Box) -> None:
+        """Construct a bounding box from a Bnd_Box"""
+
+    @overload
+    def __init__(self, shape: TopoDS_Shape, tolerance: float | None = None, optimal: bool = True) -> None:
+        """Construct a bounding box from a TopoDS_Shape"""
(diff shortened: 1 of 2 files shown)


gumyr:
Thanks for working on this. Let me know when you're ready for a review.

DaWeidmann:
When running `mypy` locally, I do not get any errors. Also, I do not understand why the errors are in files I did not even touch...

DaWeidmann:
Apart from that, the PR is ready

gumyr:
Thank you!  Turns out that mypy was upgraded and the build servers get a new version every time so it saw new problems while you or I didn't. I was able to fix the typing problems, which were unrelated to this PR so we're back on track now.

gumyr:
@DaWeidmann Now that the constructor is in place and handles the simple case of a `TopoDS_Shape`, would you be willing to expand the constructor in a small way to take a `Shape`?  The reason behind all of this is that it's a bit weird that we have to create bounding boxes with `my_object.bounding_box()` while the standard Python method (that's used almost universally in build123d) is to have the constructor do this like `list(...)`, `tuple(...)` or `Shell(...)`. We should be able to do `bbox = BoundBox(my_object)`. If you don't have time I'll create a new issue for it.

PR #1265, opened 2026-03-31
Add NGSolve/Netgen interoperability API

### LINKED ISSUE #297: Integrate build123d & Netgen
The Netgen/NGSolve system, available at https://ngsolve.org/, enables CFD analysis of BREP parts and can generate objects using the OpenCascade CAD core. It appears possible to convert build123d objects into native Netgen objects for analysis, potentially unlocking a diverse range of applications. Netgen incorporates a basic version of build123d's Algebra system, suggesting a potential avenue for collaboration between these projects.

zalo:
## Summary

Adds a `to_ngsolve_mesh()` function that converts build123d shapes to NGSolve meshes for finite element analysis, closing the gap described in #297.

- **New module**: `src/build123d/ngsolve_interop.py` with `to_ngsolve_mesh()` function
- **BREP interchange**: Transfers geometry via temporary BREP file (direct shape passing not possible due to different pybind11 wrapper types)
- **Face label propagation**: Uses Netgen's `faces.Nearest(gp_Pnt).name` API (credit: @rmsc's [suggestion](https://github.com/gumyr/build123d/issues/297#issuecomment-4159945906)) to match build123d face labels to netgen faces before meshing. Handles non-convex faces (C-shapes, annular sectors) via rejection sampling in UV parameter space when `face.center()` falls outside the trimmed boundary.
- **Lazy imports**: `netgen` and `ngsolve` are imported lazily with helpful error messages — they are **not** added as hard dependencies
- **Python 3.10+**: No additional version requirements beyond build123d's own `requires-python >= 3.10`
- **Two labeling modes**: Pass a `face_labels` dict, or set `.label` on faces using build123d's selection API
- **Documentation**: Section added to `docs/import_export.rst`
- **Example**: `examples/ngsolve_poisson.py` (gracefully skips if ngsolve not installed)
- **Tests**: `tests/test_ngsolve_interop.py`

### Why Python 3.10+ with no OCP version restriction?

We tested every combination systematically:

| Python | cadquery-ocp package | OCP version | netgen-occt | Co-import | Full pipeline |
|--------|---------------------|-------------|-------------|-----------|---------------|
| 3.9 | cadquery-ocp 7.7.2 | OCCT 7.7.2 | 7.8.1 | **FAILS** (`gp_Pnt already registered`) | N/A |
| 3.10 | cadquery-ocp 7.7.2 | OCCT 7.7.2 | 7.8.1 | OK | OK |
| 3.10 | cadquery-ocp-novtk 7.7.2.2b2 | OCCT 7.7.2 | 7.8.1 | OK | OK |
| 3.10 | cadquery-ocp 7.8.1 | OCCT 7.8.1 | 7.8.1 | OK | OK |
| 3.10 | cadquery-ocp-novtk 7.9.3 | OCCT 7.9.3 | 7.8.1 | OK | OK |
| 3.11 | cadquery-ocp-novtk 7.9.3 | OCCT 7.9.3 | 7.8.1 | OK | OK |
| 3.12 | cadquery-ocp-novtk 7.9.3 | OCCT 7.9.3 | 7.8.1 | OK | OK |

The pybind11 type conflict is a **Python 3.9 runtime issue**, not a package version issue. The same `cadquery-ocp==7.7.2` that fails on 3.9 works fine on 3.10. Since build123d already requires Python >= 3.10, no additional version guard is needed.

### aarch64 Linux

`netgen-occt` only ships wheels for x86_64 Linux, macOS (x86_64 + arm64), and Windows. No aarch64 Linux wheels exist — this is a netgen upstream limitation. The lazy import handles this gracefully: build123d imports fine, and `to_ngsolve_mesh()` raises a clear `ImportError`.

### Usage

```python
import build123d as bd
from build123d import to_ngsolve_mesh
import ngsolve as ngs

part = bd.Cylinder(10, 20) - bd.Cylinder(5, 20)

# Option 1: face_labels dict
labels = {}
for f in part.faces():
    labels[f] = "ends" if abs(f.center().Z) > 1 else "walls"
mesh = to_ngsolve_mesh(part, face_labels=labels, maxh=2)

# Option 2: .label attribute (uses build123d's face selection API)
for f in part.faces():
    f.label = "ends" if abs(f.center().Z) > 1 else "walls"
mesh = to_ngsolve_mesh(part, maxh=2)
```

## Test plan

- [x] `from build123d import to_ngsolve_mesh` works without ngsolve installed
- [x] Calling without ngsolve raises helpful `ImportError` with install instructions
- [x] End-to-end example runs on Python 3.10, 3.11, and 3.12 (identical results)
- [x] Face labels propagate correctly on convex and non-convex geometries
- [x] Non-convex faces (C-shapes, U-channels) matched correctly via UV rejection sampling

### Test output (identical on 3.10, 3.11, 3.12)
```
Boundaries:  ('walls', 'ends', 'ends', 'walls')
DOFs:        3783 (3185 free)
Solution:    -16.2320 .. 50.0242
```

Closes #297

### DIFF
--- a/src/build123d/ngsolve_interop.py
+++ b/src/build123d/ngsolve_interop.py
@@ -1,0 +1,212 @@
+"""
+build123d NGSolve/Netgen Interoperability
+
+name: ngsolve_interop.py
+by:   build123d contributors
+date: March 30th, 2026
+
+desc:
+    Provides interoperability between build123d and NGSolve/Netgen for
+    finite element analysis. Transfers geometry via BREP interchange and
+    propagates face labels as mesh boundary condition names using geometric
+    center matching.
+
+    Requires Python 3.12+ where cadquery-ocp and netgen-occt both use
+    OCCT 7.8.1, allowing coexistence in the same process.
+
+    See: https://github.com/gumyr/build123d/issues/297
+
+license:
+
+    Copyright 2026 build123d contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+"""
+
+from __future__ import annotations
+
+import math
+import tempfile
+import warnings
+from typing import TYPE_CHECKING
+
+from build123d.exporters3d import export_brep
+from build123d.topology import Face, Shape
+
+if TYPE_CHECKING:
+    from netgen.occ import OCCGeometry
+    from ngsolve import Mesh as NGMesh
+
+__all__ = ["to_ngsolve_mesh"]
+
+
+def _lazy_import_netgen():
+    """Import netgen.occ, raising a helpful error if not installed."""
+    try:
+        from netgen.occ import OCCGeometry
+
+        return OCCGeometry
+    except ImportError:
+        raise ImportError(
+            "netgen is required for NGSolve interoperability but is not installed.\n"
+            "Install it with:  pip install netgen-occt netgen-occt-devel\n"
+            "Note: Python 3.12+ is required so that cadquery-ocp and netgen-occt\n"
+            "both use matching OCCT versions (7.8.1+).\n"
+            "See: https://ngsolve.org/  and  https://github.com/gumyr/build123d/issues/297"
+        ) from None
+
+
+def _lazy_import_ngsolve():
+    """Import ngsolve, raising a helpful error if not installed."""
+    try:
+        import ngsolve
+
+        return ngsolve
+    except ImportError:
+        raise ImportError(
+            "ngsolve is required for NGSolve interoperability but is not installed.\n"
+            "Install it with:  pip install ngsolve\n"
+            "Note: Python 3.12+ is required so that cadquery-ocp and netgen-occt\n"
+            "both use matching OCCT versions (7.8.1+).\n"
+            "See: https://ngsolve.org/  and  https://github.com/gumyr/build123d/issues/297"
+        ) from None
+
+
+def _match_face_labels(
+    b123d_faces: list[tuple[float, float, float, str]],
+    ng_geo: "OCCGeometry",
+) -> dict[int, str]:
+    """Match NGSolve geometry face indices to build123d face labels.
+
+    Uses nearest geometric center matching between build123d faces and
+    netgen geometry faces.
+
+    Args:
+        b123d_faces: list of (cx, cy, cz, label) tuples from build123d
+        ng_geo: netgen OCCGeometry object
+
+    Returns:
+        dict mapping netgen face index to label string
+    """
+    mapping = {}
+    for ng_idx, ng_face in enumerate(ng_geo.shape.faces):
+        nc = ng_face.center
+        best_label = min(
+            b123d_faces,
+            key=lambda fm: math.sqrt(
+                (nc.x - fm[0]) ** 2 + (nc.y - fm[1]) ** 2 + (nc.z - fm[2]) ** 2
+            ),
+        )[3]
+        mapping[ng_idx] = best_label
+    return mapping
+
+
+def to_ngsolve_mesh(
+    shape: Shape,
+    face_labels: dict[Face, str] | None = None,
+    maxh: float = 1.0,
+    **mesh_kwargs,
+) -> "NGMesh":
+    """Convert a build123d Shape to an NGSolve Mesh with named boundaries.
+
+    Transfers the geometry via BREP file interchange (direct shape passing is
+    not possible because build123d and netgen use different pybind11 wrappers
+    for OpenCASCADE types). Face labels are propagated by matching geometric
+    centers between the build123d and netgen representations.
+
+    Requires ``netgen`` and ``ngsolve`` packages to be installed. These are
+    **not** hard dependencies of build123d — an ``ImportError`` with
+    installation instructions is raised if they are missing.
+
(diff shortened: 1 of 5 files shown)


rmsc:
I honestly don't think this is the way to go.

zalo:
Is that because you'd like the API to [look different](https://github.com/gumyr/build123d/issues/297#issuecomment-4159945906)?  Or because you'd rather copy and paste the brep glue code every time you want to use ngsolve?

src/build123d/ngsolve_interop.py
```diff
@@ -50,14 +52,16 @@
 
 __all__ = ["to_ngsolve_mesh"]
 
+_MIN_PYTHON = (3, 12)
```
jdegenstein: This looks like nonsense to me, a quick check here https://pypi.org/project/ngsolve/#files implies that 3.10 through 3.14 are supported. I saw your earlier comment about 3.9 but that version of python was EOL in fall 2025. Please justify this minimum version.

jdegenstein:
The current stable release of build123d depends on cadquery-ocp==7.8.x which matches nicely with netgen-occt which is currently only available for ==7.8.x as well. The dev branch and upcoming imminent release of build123d==0.11.0 will depend on cadquery-ocp==7.9.x for which there is no matching netgen-occt. I saw the earlier discussion about runtime incompatibility with differing versions of ocp/occt. This seems to imply to me that this integration will be quite brittle and will potentially break immediately.

The other issue is that netgen-occt is not available for linux-aarch64 but that will be a supported platform for the next release build123d==0.11.0. This is IMHO a much more minor issue but worth mentioning.

Given the first point above, maybe it makes sense to keep this functionality in a separate project? Open to suggestions here.

rmsc:
Sarcasm aside, I think the point of any integration would be to use build123d's face selection API.

As is, the proposed API doesn't materially improve usability compared to just using NGSolve's API:

```python
# Proposed
labels = {}
for f in part.faces():
    labels[f] = "ends" if abs(f.center().Z) > 1 else "walls"

# Netgen
shape.faces.Max(Z).name = "ends"
shape.faces.Min(Z).name = "ends"
(...)
```

Also missing is the ability to set other properties such as `maxh`.

The example glue code I gave is just 9 LoC including the `tempfile` import and the utility function. It allows me to use build123d's face selection for setting the properties without having to build a dict of labels.

So while I really appreciate the effort, as I said, I don't think this is the way to go.

zalo:
@jdegenstein The reasoning given by Claude is that it was the specific OCCT versions that caused the trouble with importing different pybind11 OCCT wrappers into the same process, but that doesn’t make sense to me; I suspect pybind11 just got better at namespace isolation between the Python 3.9 and 3.12 releases, which prevents objects of the same name from conflicting.

The BREP format between versions is stable, as far as I can tell, and if it isn’t, we can use STEP files as the communication medium, which are definitely stable.

I only tried 3.9 and 3.12; I’ll try 3.10 and each one in between to verify minimum compatibility.


@rmsc I appreciate the elegance of your suggested API and I’d be happy to take suggestions there.  Is your preferred outcome that all b123d + FEA users find the magic 9-line temporary file shim (which apparently nobody found for like 3 years since the issue was opened) over having a simple 9-line pair of named functions documented in the API?

My intent with this API is to solve CAD -> FEA interop in a “once and for all” sense, but I also understand if lazily imported FEA integrations pollute the elegance of b123d’s CAD-only sphere of concerns.

zalo:
Just determined that the interop does indeed work back to 3.10, so no issues there on any version of netgen or b123d.  The issue in 3.9 was indeed just a pybind11 namespace conflict unique to only 3.9 (vanishing in 3.10 even with the same package versions).  So it should be robust wrt versioning.

I think I might need to get smarter about face selection, since I think the center selection will not work will for highly non-convex faces (eg crescent moon).

src/build123d/ngsolve_interop.py
zalo: You’re right; it works as early as 3.10 and every version in-between.

zalo:
PR Description updated with all the latest information.

I now have it fall back to rejection sampling within the UV Space of a face to guarantee finding a point that lies on the face.  This should make the face transfer more robust for non-convex faces.

gumyr:
Thank you for the considerable work investigating build123d ↔ NGSolve/Netgen interoperability. After review, I’m going to close this PR rather than merge it into build123d.

The core difficulty is not BREP transfer or producing a mesh. It is preserving the semantic relationship between a CAD model and its analysis mesh:

- build123d can select meaningful BREP faces—e.g. a mounting face, loaded face, or contact surface;
- an FEA solver applies conditions to the many mesh elements created from those faces;
- the mesher needs a reliable CAD-face → mesh-element classification to turn those selections into named boundary regions, physical groups, materials, mesh-size controls, etc.

There is no broadly supported interchange standard for that provenance. BREP and ordinary STEP exchange geometry, but do not reliably preserve a stable cross-tool identity for individual faces. STEP/XDE can carry some subshape metadata, such as names, colours, and layers, but receiving tools are not generally required to preserve or consume it as analysis regions. Once another CAD/meshing system imports and reconstructs the BREP, topology may be renumbered, split, merged, or otherwise changed.

The proposed Nearest(point) approach attempts to reconstruct the relationship after import, but it is necessarily geometric matching rather than provenance. It can be ambiguous and cannot provide the general guarantees needed for boundary conditions, contact, material regions, or local mesh controls. In addition, labels set on transient Face selection wrappers are not persistent metadata on the source shape, so the proposed .label workflow does not reliably survive a later shape.faces() traversal.

This is also a two-sided interoperability problem. There is little value in build123d generating a new face-metadata convention unless Netgen, Gmsh, and other meshing tools consume it; conversely, meshing tools have little reason to implement such a convention unless CAD systems produce it. Established CAD/FEA workbenches generally solve this within their own document models or dedicated adaptors, where they control both the CAD-face identity and the mesh physical groups.

A future solver-specific companion project could own both ends of such a contract and demonstrate a robust mapping strategy. For build123d itself, however, adding and maintaining a Netgen-specific integration would create a large support commitment without a reliable, general interchange mechanism.

Issue #312, opened 2023-09-06
Add a major/minor radius flag parameter to the Polygon shapes


gumyr:
Currently the radius provided is the major radius (that of the circumscribed circle) - the new flag would indicate that the radius is a minor radius (that of the inscribed circle) thus allowing either of the hexes shown in this example to be created without the user doing any mathematics.

![image](https://github.com/gumyr/build123d/assets/85638518/15024413-1220-4498-a944-5b48944b7c98)

jdegenstein:
The ratio of the inscribed radius r to the circumscribed radius R is `r/R = cos(pi/n)` where n is the number of sides of the regular polygon. For instance `r/R = 0.866...` for a hexagon and `r/R = 0.5` for a triangle.

gumyr:
Thanks @jdegenstein !

Issue #305, opened 2023-08-24
Unfold Sheet Metal Parts


gumyr:
Take a model designed in 3D representing a part made from sheet metal and unfold it back to a planar piece of sheet metal as a cutting pattern. Roughly it work works as follows:

The algorithm that I've come up with is:
1) Find all of the Faces that are continuous with each other - this separates the "edges" of the sheet metal from the surface of the sheet metal.  The attached screenshot shows some continuous Faces (pink) and a translucent non-continuous Face in the back.  Unfortunately, I found that the OCP BRep_Tool.Continuity_s tool's result vary depending on how the faces were constructed so I've got more work to do here.
2) Pick a non-planar Face identified in step 1, generate a new planar version with "the same" area (actually the area will be a little different based on the "k-factor" of the material - the physics of bending stuff). I'm going to assume that the bend is circular to make the math manageable. This new planar Face will contain the common Edge from the adjoined planar Face.
3) Sew the Faces together to create a new part with one less bend.
4) Repeat steps 2 and 3 until there are no more non-planar Faces.
5) Extrude the newly formed planar Faces back into a 3D sheet metal part. The extrusion amount is determined from the sheet metal "edges" found in step 1.

kcleung:
Some thoughts: for the unfolding function, perhaps we can have two variants:

* laser cut and CNC bend - where the SVG represents the following layers: untouched, cut, top engrave, bottom engrave, fold line
* photoetch - where the SVG have layers representing the following layers: unetched, fully-edged, top-edged and bottom-edged layers.

However, for the fold lines, how can we represent the angle and radius of each fold line in the SVG or DXF file?  This information is vital for CNC fold.

Furthermore, I understand from my experience in building from brass locomotive kits that in the case of photoetch, it is a common practice to convert fold lines into a half etch on the inner side of the fold, so that the user can easily fold the sheet correctly with minimal tools.

Can people with experience in laser cut, waterjet cut, CNC fold and photoetching share your thoughts?

jorgedanisc:
Hi! I can share some **practical fabrication feedback** (not implementation details) since I have direct access to a factory with a modern laser cutter, and we do a lot of HVAC ductwork.

- **DXF output is basically the default for laser cutting.** Most modern laser cutters (or the workflow around them) expect DXF as the handoff format.

- **An unfolded 3D part often becomes *multiple* separate 2D pieces (not one connected pattern).**  
[...]

PR #1270, opened 2026-04-07
Materials class

bernhard-42:
As discussed, here is a proposal for a `Material` class and shape integration (`Shape.material`, `Shape.mass`).

Ways to define material:
- `shape.material = "brass"` (this needs to be a string supported by py-materials)
    This will use the simple PBR definitions of py-materials
- `shape.material = brass` (can be used after `from pymat import brass`)
    This will also use the simple PBR definitions of py-materials
- `shape.material = Material(brass, pbr=PbrProperties.from_gpu("Brass Brushed")` 
    This adds sophisticated visual material properties to the physical material

It is similar to Color ("red" or Color(1.0, 0.0, 0.0))

- - -

**Open issues to be discussed:**
- Python 3.10 support
    py-materials currently only supports Python >= 3.11 (see https://github.com/MorePET/mat/pull/6)
    This PR temporarily sets the min Python to 3.11, but tests fail due to linting under 3.10
- Optional dependencies of threejs-materials 
    materialx and openexr are optional, since there is no binary for Python 3.14 on pypi. Under Linux and Macos, `pip install materialx openexr` will compile the package, but Windows machines without compiler will not be able to do so. So in order to not break Py 3.14 compatibility I made both packages optional (workaround: you can in a separate 3.13 environment download an bake materials into the cache, and in the 3.14 loading materials will read from cache without baking)
- I haven't created tests by now, thought to first discuss whether the approach makes sense
- - -

Example:
```python
from build123d import *
from ocp_vscode import *
from threejs_materials import PbrProperties

from ocp_vscode.utils import create_shader_ball

# pymat.Material
from pymat import brass

# threejs-material
brass_tm = PbrProperties.from_gpuopen("Brass").scale(4, 4)

# Import threejs-material from GLTF2 (blender export)
brass_gltf = PbrProperties.load_gltf("env/brass-deco/brass_cube.gltf")[
    "Ornamental Design Embossed Brass"
]

shader_ball_1 = Pos(-30, 0, 0) * create_shader_ball("shader_ball_1")
shader_ball_1.label = "shader_ball_pymat"
shader_ball_1.material = "brass"

shader_ball_2 = Pos(0, 0, 0) * create_shader_ball("shader_ball_2")
shader_ball_2.label = "shader_ball_tm"
shader_ball_2.material = Material(brass, pbr=brass_tm)

shader_ball_3 = Pos(30, 0, 0) * create_shader_ball("shader_ball_3")
shader_ball_3.label = "shader_ball_gltf"
shader_ball_3.material = Material(brass, pbr=brass_gltf)

show(
    shader_ball_1,
    shader_ball_2,
    shader_ball_3,
    materials=[brass_tm, brass_pymat, brass_gltf],
    studio_texture_mapping=StudioTextureMapping.PARAMETRIC,
    studio_environment=StudioEnvironment.SPOTLIT_SETUP,
)
```
Screenshot 2026-04-07 at 17 55 05


src/build123d/topology/composite.py
```diff
@@ -182,10 +183,21 @@ def _dim(self) -> int | None:
...
         return sum(i.volume for i in [*self.get_type(Solid), *self.get_type(Shell)])
 
+    @property
+    def mass(self) -> float:
+        """mass - the mass of this Compound in g"""
+        # requires material being set for the shapes
+        if isinstance(self.material, Material):
+            density_g_mm3 = self.material.mechanical.density / 1000
+            if density_g_mm3 == 0:
+                print("Shape's density is 0")
```
gumyr: This should be a warning

src/build123d/topology/shape_core.py
```diff
@@ -355,6 +360,51 @@ def color(self, value: ColorLike | None) -> None:
...
+    def material(self, value: str | pymat.Material | Material | None) -> None:
+        """Set the shape's material"""
+        if value == "" or value is None:
+            self._material = None
+            return
+
+        if isinstance(value, str):
+            self._material = Material(copy.deepcopy(pymat.registry.get(value)))
+            if self._material is None:
+                print(f"Unknown material {value}, using 'pla'")
```
gumyr: This should be a warning

src/build123d/topology/shape_core.py
```diff
@@ -355,6 +360,51 @@ def color(self, value: ColorLike | None) -> None:
...
+                self._material = Material(copy.deepcopy(pymat.registry.get("pla")))
+
+        elif isinstance(value, Material):
+            self._material = value
+
+        elif isinstance(value, pymat.Material):
+            self._material = Material(value)
+
+        else:
+            print(f"Unknown material type {type(value)}, using 'pla'")
```
gumyr: warning

src/build123d/topology/three_d.py
```diff
@@ -753,9 +754,20 @@ def __init__(
...
+        # hence density == 1, and OCCT mass == volume
         return Shape.compute_mass(self)
+    
+    @property
+    def mass(self) -> float:
+        """mass - the mass of this Solid in g"""
+        if isinstance(self.material, Material):
+            density_g_mm3 = self.material.mechanical.density / 1000
+            if density_g_mm3 == 0:
+                print("Shape's density is 0")
```
gumyr: warning

src/build123d/topology/two_d.py
```diff
@@ -2688,6 +2694,16 @@ def volume(self) -> float:
             return properties.Mass()
         return 0.0
 
+    @property
+    def mass(self) -> float:
+        """mass - the mass of this Shell if manifold in g, otherwise zero"""
+        if isinstance(self.material, Material):
+            density_g_mm3 = self.material.mechanical.density / 1000
+            if density_g_mm3 == 0:
+                print("Shell's density is 0")
```
gumyr: warning

src/build123d/geometry.py
```diff
@@ -1553,6 +1565,178 @@ def _rgb_from_str(name: str) -> tuple:
...
+
+        Returns:
+            Color: interpolated color
+        """
+        return Color(
+            self.pbr.interpolate_color()
+        )
+
+    @classmethod
+    def create(
```
gumyr, line 1732: Should this be _create? It's an internal function not intended for the user correct?

src/build123d/geometry.py
```diff
@@ -1553,6 +1565,178 @@ def _rgb_from_str(name: str) -> tuple:
...
+            mat = material
+        else:
+            raise TypeError(
+                f"A parameter of type {type(material)} is not supported for materials"
+            )
+
+        if density or color:
+            # deep copy to not change the original material
+            mat = copy_module.deepcopy(mat)
+            if color:
```
gumyr, line 1647: If color isn't defined here would something bad happen?

src/build123d/topology/three_d.py
```diff
@@ -753,9 +754,20 @@ def __init__(
...
+            return self.volume * density_g_mm3
+        raise RuntimeError("Shape has no material definition ")
```
gumyr: All the other cases where the material is defined correctly return 0/warning - should that be true here too?

gumyr (changes requested):
Thank you!

I'm okay dropping support for Python 3.10 as we'll still support the last 4 versions which was our goal.  @jdegenstein what do you think?

I just have minor issues, with some tests we should be able to merge.

src/build123d/topology/shape_core.py
```diff
@@ -355,6 +360,51 @@ def color(self, value: ColorLike | None) -> None:
...
+            node_material = parent_material
+        else:
+            node_material = self._material
+        self._material = node_material  # Set the node's material for next time
+        return node_material
+
+    @material.setter
+    def material(self, value: str | pymat.Material | Material | None) -> None:
+        """Set the shape's material"""
+        if value == "" or value is None:
```
jdegenstein, line 387: I think this is equivalent to `if not value:` which seems superior to me.

src/build123d/topology/composite.py
```diff
@@ -182,10 +183,21 @@ def _dim(self) -> int | None:
 
     @property
     def volume(self) -> float:
-        """volume - the volume of this Compound"""
+        """volume - the volume of this Compound in mm^3"""
```
jdegenstein: I have advocated for some intentional ambiguity in units internal to build123d. Essentially the policy is (and IMHO should be) that units only become "real" when you move data in or out of the library via e.g. a STEP export, which *defaults* to mm. This also matches the approach taken by OCCT itself.

I recommend this and related changes are reverted and that the mass methods are changed (see below).

src/build123d/topology/composite.py
```diff
@@ -182,10 +183,21 @@ def _dim(self) -> int | None:
...
         # when density == 1, mass == volume
         return sum(i.volume for i in [*self.get_type(Solid), *self.get_type(Shell)])
 
+    @property
+    def mass(self) -> float:
+        """mass - the mass of this Compound in g"""
```
jdegenstein: I agree that the mass property should (1) default to output grams and (2) by default assume `Shape` is in units of mm. However, I believe this should be configurable in keeping with API consistency.

New signature:
```py
def mass(self, length_unit: Unit = Unit.MM, mass_unit: MassUnit = MassUnit.G) -> float:
```
Note that `MassUnit` would need to be created in `build_enums.py` (and a related change to add a dict in `build_common.py` would be a nice-to-have as well). I have not cross-checked if this creates any conflicts with py-materials.

Please propagate this change to the other mass methods as appropriate.

pyproject.toml
```diff
@@ -52,6 +52,10 @@ dependencies = [
     
     # Install py-lib3mf ONLY on linux aarch64
     "py-lib3mf >= 2.4.1; sys_platform == 'linux' and platform_machine == 'aarch64'",
+
+    "pygltflib>=1.16.5, < 1.17.0",
+    "py-materials>=2.0.5, <2.1.0",
+    "threejs-materials>=1.0.0,<1.1.0",
```
jdegenstein: Please make the spacing consistent in this section.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ authors = [
 ]
 description = "A python CAD programming library"
 readme = "README.md"
-requires-python = ">= 3.10, < 3.15"
+requires-python = ">= 3.11, < 3.15"
```
jdegenstein, line 17: would like to see this expanded upstream in py-materials (saw your PR), but I also don't think it is necessarily a deal breaker to drop py310 early

jdegenstein (commented):
I prefer not to merge this before build123d==0.11.0 release (which I hope to finalize very soon) but this seems like a valuable enhancement.

src/build123d/topology/composite.py
bernhard-42: Done

src/build123d/topology/shape_core.py
bernhard-42: Done

src/build123d/topology/shape_core.py
bernhard-42: Done

src/build123d/topology/three_d.py
bernhard-42: Done

src/build123d/topology/two_d.py
bernhard-42: Done

src/build123d/geometry.py
bernhard-42: It is intended to be a public method. Reason:

```python
Material("brass")
``` 

uses the predefined py-materials marterial `brass` and converts `brass.PBRProperties` to a threejs-materials `PbrProperties` property `pbr` (which is the interface for ocp_vscode)

```text
material: Material('brass', ρ=8.5 g/cm³)
pbr: PbrProperties(name='Brass (Cu-Zn)', source='custom', license='')
  values:  PbrValues(color=[0.88, 0.78, 0.5], metalness=1.0, roughness=0.25, ior=1.5, emissive=[0, 0, 0])
  maps:    PbrMaps()
```

If you want to use a py-materials material and overwrite the density, then in line with the py-materials API one uses

```python
Material("brass", density=8.6)
```

leading to 

```text
material: Material('brass', ρ=8.6 g/cm³)
pbr: PbrProperties(name='Brass (Cu-Zn)', source='custom', license='')
  values:  PbrValues(color=[0.88, 0.78, 0.5], metalness=1.0, roughness=0.25, ior=1.5, emissive=[0, 0, 0])
  maps:    PbrMaps()
```

I also needed a way to create custom materials that don't exist in py-materials:

```python
plywood_pbr = PbrProperties.load_gltf("/Users/bernhard/Downloads/plywood/plywood_cube.gltf")["Light Natural Plywood"]
Material.create("plywood", density=0.6, pbr=plywood_pbr)
```

leading to 

```text
material: Material('plywood', ρ=0.6 g/cm³)
pbr: PbrProperties(name='Light Natural Plywood', source='gltf', license='')
  values:  PbrValues(color=[1.0, 1.0, 1.0], metalness=0.0, roughness=1.0, side=2)
  maps:    PbrMaps(color='data:...;base64,...', metalness='data:...;base64,...', roughness='data:...;base64,...', normal='data:...;base64,...')
  normalize_uvs: False
```

So using a py-materials material and creating a new one has the same method signature `(str, float)`, hence the `create` class method.

If there is a better name/pattern, please let me know

src/build123d/geometry.py
bernhard-42: Then the default `base_color` of py-materials will be used, which is always set with default `(0.8, 0.8, 0.8, 1.0)`.
The `density` and `color` args are just for quick overrides, e.g. red tinted glass.

src/build123d/topology/three_d.py
bernhard-42: I am not sure what to do. The volume always exists, but a mass only when density is available.
I thought it should raise an error when the user asks for the mass but hasn't given any density.
But I would follow your guidance here. What are you typically returning in such a case?

bernhard-42: Or should we "assume" `density = 1.0` and add a warning that "density is assumed as 1.0 g/mm3"

src/build123d/topology/shape_core.py
bernhard-42: Using `if not value` would lead to

```python
In [28]: x(True)
Unknown material type <class 'bool'>, using 'pla'

In [29]: x(False)
# sets self._material to None
```

As implemented, it correctly falls through to the "unknown material" branch

```python
In [31]: x(True)
Unknown material type <class 'bool'>, using 'pla'

In [32]: x(False)
Unknown material type <class 'bool'>, using 'pla'
```

I personally have had some bad experience with Python's falsey behavior and avoid it.

In this case I only want the two allowed "empty" constructs ("" and None) being treated as empty, everything else should warn the suer

pyproject.toml
bernhard-42: Done

pyproject.toml
bernhard-42: Not a lot I can do in upstream ...

src/build123d/topology/three_d.py
bernhard-42: Reminder: needs to be same in Shell.mass in two_d.py

src/build123d/topology/composite.py
bernhard-42: Done, reverted the changes where I added mm^3 in the comments

src/build123d/topology/composite.py
bernhard-42: With this change, mass can't be a property any more, leading to

```python
shape.volume
shape.mass()
```

Would this be OK?

src/build123d/topology/three_d.py
bernhard-42: Revisited this and changed for `Compound`, `Solid`, and `Shell` the `mass` property to have this fallback:

```python
        warnings.warn("Solid's density is missing, assuming 1.0 g/mm^3")
        return self.volume
 ```
 (replace "Solid" with "Shell" or "Compound" in the other classes)

File: src/build123d/topology/composite.py (hunk 11 above)
Comment (bernhard-42) [reply src/build123d/topology/composite.py]: I don't think that `length_unit` and `mass_unit` are shape properties, but session properties.
How about a global functions in geometry.py

```python
MASS_UNIT = Unit.G
LENGTH_UNIT = Unit.MM

def get_units():
    return {"mass_unit": MASS_UNIT, "length_unit": LENGTH_UNIT}

def set_units(length_unit: Unit = Unit.MM, mass_unit: MassUnit = MassUnit.G):
    global MASS_UNIT, LENGTH_UNIT
    MASS_UNIT = mass_unit
    LENGTH_UNIT = length_unit
```

with 

```python
class MassUnit(Enum):
    G = auto()
    KG = auto()
    OZ = auto()
    LB = auto()
```

(or should it be just 4 more constants in `Unit`?

But I don't oversee what this would mean for the different areas of build123d.

For this PR, I would prefer to assume length unit `mm` and mass unit `g` in the `mass` methods which is in line with the build123d pattern

```Python
h = 10 * IN
```

which returns `mm`.

Or do I miss something?

pyproject.toml
bernhard-42: Should we change 

build123d/.readthedocs.yaml
.github/workflows/coverage.yml
.github/workflows/lint.yml
.github/workflows/test.yml
.github/workflows/type.yml

to use Python 3.11 instead of 3.10?
Then the CI/CD process could run

bernhard-42:
@jdegenstein 
> I prefer not to merge this before build123d==0.11.0 release (which I hope to finalize very soon) but this seems like a valuable enhancement.

What are your concerns in merging it now?
It is a well encapsulated change and as long as users don't set the `material` property, the changes don't imapct their code.

However, with it being merged now I could update the Readme and examples in OCP VS Code.

Obviously, we need
- docs
- tests

which I am working on now.

jdegenstein:
>What are your concerns in merging it now?

(1) I would like to have one last release supporting python 3.10 and (2) this change is large enough that I think the likelihood of API / behavior changes is very likely and I prefer there is time for things to "settle" with real user testing.

.github/workflows/coverage.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup/
         with:
-          python-version: "3.10"
+          python-version: "3.11"
```
jdegenstein, line 12: per latest update of py-materials please revert this and all related changes

src/build123d/topology/three_d.py
gerchowl: Joining late — quick pushback on `warn + assume 1.0 g/mm³`: silent wrong numbers can be worse than loud errors.

Examples where the 1.0 assumption fails catastrophically in nuclear medicine / particle transport work:

- **Gamma shielding calc on a lead collimator**: density-less "mass" feeds μ/ρ attenuation at 511 keV as if ρ = 1.0 g/cm³. Actual lead is 11.34. Shielding is off by 11×, with a warning no one reads.
- **Tungsten PET septa**: same story, worse — assumed 1.0 vs actual 19.3 is a 19× error on the mass that determines the dose attenuation.
- **Phantom dosimetry**: D = E / m. If `m` is fake because density was assumed, the absorbed dose is fake too — and 1.0 g/cm³ happens to match soft tissue, so the bug is invisible for water phantoms and silently wrong for every bone-equivalent or lung material in the same script.

`warnings.warn()` is effectively silent in every environment this code actually runs in:

- **Jupyter notebooks**: shown once, scrolled off the screen after the next cell — and notebooks are the primary authoring environment for this kind of analysis work.
- **CI pipelines**: captured in the log stream, but nobody reads runtime warnings unless a test fails — and the calculation "succeeds", so nothing fails.
- **Batch scripts**: stderr often redirected, piped into a log file, or collapsed into a report.
- **Production services**: warnings sit in the log aggregator next to a million INFO lines, unread until someone's already investigating a separate incident.

The user who forgets the density is exactly the one who won't notice.

Two alternatives that stay non-raising:

1. **Return `None`**, type as `float | None`. `sum(s.mass for s in shapes)` hits `TypeError` at the aggregation point — loud, exactly where the mistake is.
2. **Return `float('nan')`**. Keeps float type, propagates through `sum()`/arithmetic so callers get nan out, `math.isnan(x)` is a clean explicit check.

I'd lean 1 unless float-type stability is load-bearing elsewhere.

pyproject.toml
gerchowl: sorry for that — this is no longer blocked on us. py-materials `2.1.0` shipped to PyPI earlier today with Python 3.10 support (via the `tomli` shim that @bernhard-42 himself contributed as [MorePET/mat#6](https://github.com/MorePET/mat/pull/6)). There's a `Tests (py3.10, all)` matrix cell in our CI verifying it on every PR, so it's not going to regress quietly.

src/build123d/topology/shape_core.py
```diff
@@ -1021,6 +1053,54 @@ def __sub__(self, other: None | Shape | Iterable[Shape]) -> Self | ShapeList[Sel
...
+        units = get_units()
+        mass_unit, length_unit = units["mass_unit"], units["length_unit"]
+
+        if isinstance(self.material, Material):
+            density_g_cm3 = self.material.mechanical.density  # g/cm^3
+        else:
+            warnings.warn("Shape's density is missing, assuming 1.0 g/cm^3")
+            density_g_cm3 = 1
+
+        if density_g_cm3 == 0:
```
jdegenstein, line 1094: Not clear to me how this can ever be reached.

src/build123d/exporters3d.py
```diff
@@ -287,6 +291,8 @@ def export_gltf(
     Returns:
         bool: write status
     """
+    file_str = fsdecode(file_path)
+    is_binary = file_str.endswith(".glb")
```
jdegenstein, line 295: may be good to use the case insensitive version `is_binary = file_str.lower().endswith(".glb")` instead

tests/test_direct_api/test_shape.py
```diff
@@ -82,8 +82,7 @@ def test_mirror(self):
         self.assertAlmostEqual(box_bb.max.Z, 0, 5)
 
     def test_compute_mass(self):
-        with self.assertRaises(NotImplementedError):
-            Shape.compute_mass(Vertex())
+        self.assertEqual(Vertex().compute_mass(), 0)
```
jdegenstein, line 85: seems like there is some mixing of assertEqual vs. assertAlmostEqual here -- might be good to make it consistent. I suspect the former should work fine in all cases where 0.0 is expected?

tests/test_materials.py
```diff
@@ -0,0 +1,380 @@
...
+
+# Independent unit conversions for verifying compute_mass
+# pymat density is g/cm^3.  We convert volume and density independently
+# to the target unit system, then multiply.
+#
+# (MM, G):  volume mm^3, density g/cm^3 → g/mm^3 (÷1000), mass = vol_mm3 * dens_g_mm3
+# (M, KG):  volume m^3, density kg/m^3 (×1000), mass = vol_m3 * dens_kg_m3
+# (IN, LB): volume in^3, density lb/in^3, mass = vol_in3 * dens_lb_in3
+
+G_PER_LB = 453.59237
```
jdegenstein, line 68: This can be imported from your new build_constants.py file I think.

tests/test_materials.py
```diff
@@ -0,0 +1,380 @@
...
+            6,
+        )
+        self.assertAlmostEqual(
+            self.sphere.mass,
+            expected_mass(SPHERE_VOLUME_MM3, WOOD_DENSITY_G_CM3, "in", "lb"),
+            6,
+        )
+
+    def test_volume_unchanged_by_units(self):
+        """Volume is always in mm^3 regardless of unit settings."""
```
jdegenstein, line 181: Consider revising to:
```py
"""Volume is always in arbitrary-units^3 regardless of unit settings."""
```

.readthedocs.yaml
```diff
@@ -7,7 +7,7 @@ formats:
 build:
   os: "ubuntu-22.04"
   tools:
-    python: "3.10"
+    python: "3.11"
```
jdegenstein, line 10: please also change back to 3.10 per updates to py-materials

src/build123d/topology/shape_core.py
gerchowl: Worth distinguishing three states:

1. **`density = None`** — unknown/unspecified. This is the default on a fresh `pymat.MechanicalProperties` ([properties.py#L29](https://github.com/MorePET/mat/blob/v2.1.0/src/pymat/properties.py#L29)). Currently crashes at [shape_core.py#L1100](https://github.com/bernhard-42/build123d/blob/0fec289ad7040652307d96e026fb059e704ae2a1/src/build123d/topology/shape_core.py#L1100) via `None / float → TypeError`.
2. **`density = 0.0`** — **vacuum**. py-materials already ships `pymat.vacuum` with `density_value = 0.0` ([gases.toml#L363-L368](https://github.com/MorePET/mat/blob/v2.1.0/src/pymat/data/gases.toml#L363-L368)), and vacuum regions are central to Monte Carlo transport, PET scanner, and accelerator-beamline simulations. The 0-check on this line protects the vacuum case, though it's defensively redundant — `volume × 0 == 0` works fine through the normal multiplication path too.
3. **`density > 0`** — normal matter.

So the check here isn't dead, but the hidden bug is case (1). Both reconcile cleanly if `.mass` returns `None` / `nan` when density is `None` (see [the fallback thread at #discussion_r3088537341](https://github.com/gumyr/build123d/pull/1270#discussion_r3088537341)) — vacuum still returns `0.0` correctly via `volume × 0`, and missing density becomes loud instead of silently crashing or silently assuming `1.0`.

jdegenstein: Great points, I was more commenting on case-1 as you said but I should have been more clear. Density of 0 IS physically realistic. This also implies that there need to be test cases in build123d for covering some of the extremes to ensure the behavior is as expected (e.g. with `pymat.vacuum` ). The test cases also need to use isequal vs almostequal checks as discussed below (but not exactly related).

gumyr:
It seems as though a lot of progress has been made on getting the material system worked out - thanks to all of you. Is it true that the current gating item is the v0.11.0 release? With the changes that have been made by the pymat team how much rework (if any) is required of this PR?

bernhard-42:
I have time tomorrow and will analyze the changes in py-materials and adapt the PR. I will also cover the open review comments.
An important issue, Py 3.10 support, is solved, so we can go back to build123d supporting 3.10.
More on the PR tomorrow

gerchowl:
## py-materials is in a very different place than when you last looked

**Architecture now**: `py-materials` owns the domain model — `Material`, properties, and the `.vis` namespace for visual/PBR concerns. `mat-vis` owns the visual side — hosted PBR mirror (ambientcg / polyhaven / …) + the `mat-vis-client` Python package for fetching textures + MaterialX files. Materials carry a `source` field so you always know where the appearance came from.

Two complementary search verbs:

- `pymat.search("stainless 316")` — **domain** library (fuzzy find across registered Materials incl. grades + vendors).
- `pymat.vis.search(category="metal")` — **visual** catalog (mat-vis / ambientcg).

Two axes, same verb, no namespace collision.

Output formats from a `Material`:

- **Three.js** `MeshPhysicalMaterial` dict — `pymat.vis.to_threejs(m)` or `m.vis.to_threejs()`
- **glTF 2.0 material node** — `pymat.vis.to_gltf(m)` or `m.vis.to_gltf(name=...)` (KHR_materials_ior + KHR_materials_transmission handled; base-color + normal textures inlined as base64 data URIs)
- **MaterialX** — `m.vis.mtlx.xml()` / `m.vis.mtlx.export(dir)` (self-contained package: `.mtlx` + PNG channels)

**py-materials 3.3.0 is on PyPI now** (min `mat-vis-client>=0.5.0`, Py 3.10+). Namespace discipline: `Material` owns domain, `Material.vis` owns everything visual — visual concerns never smear into `Material(...)` top-level kwargs.

**Heads-up on the mat-vis mirror**: the visual catalog is still being baked — thin but not broken. As of today `vis.search(category="metal")` returns 1 result, wood 3, stone 2, fabric/plastic 0. I just smoke-tested all 6 available materials end-to-end: every one fetches 4–5 texture channels (color, normal, roughness, displacement, plus metalness for metals and aoMap for stone) cleanly and serializes correctly to Three.js, glTF, and MaterialX. So **what's there works**; more bakes land over the coming days. Our e2e tests `pytest.skip` when a category is empty rather than failing, and I'd suggest the same defensive pattern on build123d-side integration tests until the mirror fills out.

## Quick example — and a sketch of how build123d could plug it in

```python
from pymat import Material, search, vis

# Find a Material by loose query — grades and variants surface too
housing_mat = search("stainless 316")[0]   # → Material: Stainless Steel 316L

# Or construct directly — visual concerns namespaced under vis={...}
m = Material(name="Brushed 316L",
             vis={"source": "ambientcg", "material_id": "Metal032", "tier": "1k"})

# Hand off (textures fetch + cache on first touch)
threejs_dict  = m.vis.to_threejs()    # for Three.js / ocp_vscode viewer
gltf_material = m.vis.to_gltf()       # drop-in glTF 2.0 material node
m.vis.mtlx.export("out/")             # full MaterialX package for DCC pipelines
```

From a shape-centric build123d API, that could plug in like:

```python
from build123d import Box, export_gltf
from pymat import Material, search

housing = Box(50, 50, 10)
housing.material(search("stainless 316")[0])
# or: housing.Material("ambientcg/Metal032")   # whatever shorthand your
#                                               Materials class ends up exposing

export_gltf(housing, "part.glb")
# → full PBR in the glb: baseColor + normal, metalness/roughness once
#   mat-vis#91 (packed texture) lands. Loads in Blender, ocp_vscode,
#   Omniverse, any glTF viewer.
```

Where and how the `to_gltf()` call lands inside `export_gltf` is entirely your design — we just own the material side of the hand-off. When `mat-vis` adds packed metalness+roughness textures, KTX2, or any new extension, it flows through to build123d automatically — no build123d release needed.

Ready to test / bind whenever you pick this back up. Happy to turn around review on the py-mat side quickly if anything on the API shape would be easier the other way around.

.github/workflows/coverage.yml
bernhard-42: Done

src/build123d/topology/shape_core.py
bernhard-42: If a user defines materials as 

```python
from build123d import *
import pymat

b = Box(1,2,3)
b.material = Material(pymat.Material("wood", density=0))
```

or uses a material that has (accidentially) density=0
Then `self.material.mechanical.density == 0`

Just thought it would be good to warn the user. Happy to drop it if you want

src/build123d/exporters3d.py
bernhard-42: Done

.readthedocs.yaml
bernhard-42: Done

tests/test_direct_api/test_shape.py
bernhard-42: Done
changed this and all others from AssertEqual to AssertAlmostEqual

tests/test_materials.py
bernhard-42: Done, moved it to build_constants.py

tests/test_materials.py
bernhard-42: Changed it 
And also changed `BOX_VOLUME_MM3` to `BOX_VOLUME` and `SPHERE_VOLUME_MM3` to `SPHERE_VOLUME` since the value of the volume is unit-less

src/build123d/topology/shape_core.py
bernhard-42: @gerchowl Changed to

```python
        density_g_cm3 = None
        if isinstance(self.material, Material):
            density_g_cm3 = self.material.mechanical.density  # g/cm^3
        
        if density_g_cm3 is None:  
            raise ValueError("Shape's density is missing")

        if density_g_cm3 == 0:
            warnings.warn("Shape's density is 0")
            return 0.0
```

leading to

```python
In [14]: b.material = Material(pymat.Material("wood", density=0))

In [15]: b.mass
/Users/bernhard/Development/CAD/build123d-fork/src/build123d/topology/shape_core.py:1096: UserWarning: Shape's density is 0
  warnings.warn("Shape's density is 0")
Out[15]: 0.0

In [16]: b.material = Material(pymat.Material("wood"))

In [17]: b.mass
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[17], line 1
----> 1 b.mass

File ~/Development/CAD/build123d-fork/src/build123d/topology/composite.py:198, in Compound.mass(self)
    195     if not s.material:
    196         # inherit and add material to the contained solid/shell
    197         s.material = self.material
--> 198     masses.append(s.mass)
    199 return sum(masses)

File ~/Development/CAD/build123d-fork/src/build123d/topology/three_d.py:764, in Solid.mass(self)
    761 @property
    762 def mass(self) -> float:
    763     """mass - the mass of this Solid"""
--> 764     return self.compute_mass()

File ~/Development/CAD/build123d-fork/src/build123d/topology/shape_core.py:1093, in Shape.compute_mass(self)
   1090     density_g_cm3 = self.material.mechanical.density  # g/cm^3
   1092 if density_g_cm3 is None:
-> 1093     raise ValueError("Shape's density is missing")
   1095 if density_g_cm3 == 0:
   1096     warnings.warn("Shape's density is 0")

ValueError: Shape's density is missing
```

bernhard-42:
Quick status update: 
- I have covered all comments (but not pushed yet)
- I am currently rewriting build123d's Material class based on the massive changes in py-materials.
- I have several issues open to clarify how to work with py-materials now. As soon as @gerchowl replies I can finish the rewrite and push the newest version

gerchowl:
Quick follow-up, @bernhard-42 — both py-mat issues you filed are shipped in **3.4.0** (just published to PyPI):

- **#88** (grades inherit parent `vis`) — fixed at the loader level, no more manual parent-chain walks.
- **#89** (non-fuzzy lookup) — shipped as `pymat["..."]` subscript instead of the `get()` I originally proposed (Python-idiom reasons per a falsify review — `dict.get()` returns `None`, so a raising `get` would've been a footgun). Matches key / name / grade with NFKC + case + whitespace normalization, raises with candidates on ambiguity.
- **#90** still open on mat-vis (typed errors [mat-vis#141](https://github.com/MorePET/mat-vis/issues/141), gpuopen metadata [mat-vis#142](https://github.com/MorePET/mat-vis/issues/142)) — those are data-pipeline items.

```python
from pymat import Material, search
import pymat

# Any of these work end-to-end in 3.4.0:
housing = pymat["Stainless Steel 304"]     # by name
bolt    = pymat["s316L"]                   # by key
crystal = pymat["304"]                     # by grade
housing.vis.source        # "ambientcg" — inherited from parent, was None in 3.3.x

# For build123d's string-based Material= assignment:
b1.material = pymat[user_input]           # KeyError on miss, with candidate list

# Fuzzy for when ambiguity is OK:
candidates = pymat.search("stainless 316")  # list[Material]
```

Also shipped: [`examples/build123d_integration.py`](https://github.com/MorePET/mat/blob/main/examples/build123d_integration.py) — cell-style (`# %%`) file mirroring how you write your own examples, wired into pytest via `runpy` so the full file runs top-to-bottom on every CI push. If any API drifts, the example goes red before you copy-paste it.

Ready to finish the rewrite whenever.

bernhard-42:
Thanks @gerchowl for the quick fixes. I will change my code now and test

bernhard-42:
Given all the changes in py-materials and mat-vis, I decided to create a new design for the API and rewrite the 'Material' class.

I still need to thoroughly test ocp_vsode before I can publish the version that works with this `Material` class, but wanted to already share the proposal.

## API design

The API design is based on the following principles:

- **Minimum footprint in build123d**
    
    ⇒ geometry.py: Visulation logic moved into a new file visualization.py

    ⇒ visualization.py: Leverage py-materials and threejs-materials for all visualization logic

- **Based on py-materials** (as far as possible, see below)

    - threejs-materials needs packages materialx and openexr to bake and cache MaterialX files. When a compiler is installed, installation under Python 3.14 will compile it. However, this is a restriction for many Windows users under Python 3.14 (Windows does not ship with a default compiler)
    - py-materials has built a library of pre-baked materials (mat-vis package), so no critical dependencies
    - However, UX in pymat currently not optimal (e.g. uuid based for gpuopen), see "Open issues". 
    
    ⇒ Currently there is a keyword `use_pymat` with default `True`. If `False`, threejs-materials will be used requiring installed packages materialx and openexr

- **Lazy and lean**

    - Materials can be heavy (several MB with textures). So the Material class only stores the characteristics of the material/visualization, not the textures
    - Very low memory overhead for all build123d shapes.

    ⇒ The property `pbr`, which is the interface to ocp_vscode, will resolve the materials and all their textures in order to send it to the viewer or to add materials into a glTF export (*.gltf, *.glb)

- **Abstract complexity away from the end user**

    - Both py-materials and threejs-materials are very young, so I expect changes and abstracted them away from the end user (as far as possible)

- **The API is built for 4 UX levels**, see below

    ⇒ Own `class VisProperties` with an optimized user experience for build123d


## Open issues

- mat-vis does **not support Python 3.10** and is a new dependency of py-materials (which runs under Python 3.10), see https://github.com/MorePET/mat-vis/issues/197
- User friendly access of MaterialX via material names visible on websites, see https://github.com/MorePET/mat-vis/issues/143 

CC: @gerchowl 

## The 4 UX levels

1. **Just give my physical properties, I don't care about rendering**

    ```python
    shape.material = aluminum
    shape.material = "aluminum"
    shape.mass
    ```

2. **I still don't want to render, but I want `shape.mass` to be correct for materials like wood (not in py-materials)**

    ```python
    shape.material = Material.create(
        "wood", 
        density=0.65
    )
    shape.mass
    ```

3. **I want to render, but I don't want to make a PhD about it**

    - The predefined finishes are fine for me

        ```python
        shape.material = "aluminum"
        shape.material.finishes  # show predefined finishes in py-materials (pymat)
        # ["smooth", "machined"]
        
        shape.material = Material(
            "aluminum", 
            finish="machined"
        )
        show(shape)
        ```
        
    - I want more: Red anodized aluminum, a bit shinier

        ```python
        shape.material = Material(
            "aluminum", 
            finish="machined", 
            color="red",     # override color -> tinting
            roughness=0.6    # instead of 1.0 -> shinier
        )
        show(shape)
        ```

    - Adapt the texture to my object size, e.g doubled in u and v direction

        ```python
        shape.material = Material(
            "aluminum", 
            finish="machined", 
            color="red",
            roughness=0.6,            
            texture_scale=(2,2)
        )
        show(shape)
        ```

4. **I want the nicest rendering possible:**

    - MaterialX based

        ```python
        vis = VisProperties.from_gpuopen(
            "Portoro Green Marble", 
            use_pymat=False  # pymat is hard to use for gpuopen (uuid based)
        )
        shape.material = Material.create(
            "Portoro Green Marble", 
            vis=vis
        )
        show(shape)

        vis_polished = vis.override(
            roughness=0.3, 
            texture_scale=(3,3)
        )
        shape_polished.material = Material.create(
            "Portoro Green Marble", 
            vis=vis_polished
        )
        show(shape_polished)
        ```

    - Blender baked material (supported only in threejs-materials)

        ```python
        materials = VisProperties.list_gltf_materials("brass-deco/brass_cube.gltf")
        # ["Ornamental Design Embossed Brass"]

        vis = VisProperties.from_gltf(
            "brass-deco/brass_cube.gltf", 
            name="Ornamental Design Embossed Brass"
        )
        b.material = Material(
            "brass",
            vis=vis,
        )
        ```

gerchowl:
Quick update from the py-mat side, @bernhard-42 — both the gating items you flagged in your design comment are now resolved upstream, and the safe-derive surface you sketched in `VisProperties.override(...)` lands as proper API in py-materials 3.7.0 + 3.8.0.

## Items closed

- **mat-vis 3.10 support** ([mat-vis#197](https://github.com/MorePET/mat-vis/issues/197)) — fixed; smoke-tested in a fresh `cpython-3.10.19` venv against the live `v2026.04.2` substrate, all features green.
- **gpuopen UUID UX** ([mat-vis#143](https://github.com/MorePET/mat-vis/issues/143)) — fixed in mat-vis-client 0.6.3 (which py-materials 3.5.0+ pulls). `Vis.material_id` accepts either UUID or human-readable name from `mat_vis.name`. So `use_pymat=False` is no longer needed for gpuopen access:

  ```python
  m = pymat["Portoro"] if "Portoro" in pymat else None  # if you have a TOML for it
  # or, direct construction:
  v = Vis(source="gpuopen", material_id="Portoro Green Marble", tier="1k")
  v.textures  # resolves the name → fetches; UnknownMaterialError on miss
  v.mtlx.xml()  # also works by name
  ```

  Pinned by 5 e2e tests in `test_e2e_vis.py::TestGpuopenNameLookup`.

## Safe-derive API (3.7.0 → 3.8.0)

The `VisProperties.override(roughness=0.3, texture_scale=(3,3))` pattern in your design now has direct support on the py-mat side:

```python
import pymat
from pymat.vis import VisDeltas, to_threejs

steel = pymat["Stainless Steel 304"]   # registry singleton — DON'T mutate

# Derive an independent variant
polished_vis = steel.vis.override(roughness=0.05, finish="polished")
shiny = steel.with_vis(polished_vis)

three = to_threejs(shiny)              # uses the polished values
assert steel.vis.material_id == "Metal012"  # registry untouched

# For helper functions that build delta dicts, type them:
def matte_red() -> VisDeltas:
    return {"roughness": 0.7, "base_color": (0.7, 0.1, 0.1, 1.0)}
m2 = steel.with_vis(steel.vis.override(**matte_red()))
```

- **`Vis.override(**deltas) -> Vis`** — deep-copied variant; `finishes` deep-copied too; identity changes invalidate the texture cache atomically; unknown kwargs raise `TypeError`. Typed via `Unpack[VisDeltas]` (PEP 692) so pyright/mypy/Pylance flag typos like `roughnes=` at edit time.
- **`Material.with_vis(vis) -> Material`** — registry-detached Material with the supplied Vis attached. Deep-copies the vis to defend against `m.with_vis(m.vis)` aliasing.
- **`Material.copy() -> Material`** — generic detach if you need to tweak more than just the vis.

## Item not implemented (deliberate)

**`texture_scale: (u, v)`** — three independent design reviews converged on declining this as a `Vis` field. UV transform is per-instance render state (the same Portoro marble wants `(1,1)` on a coaster and `(8,8)` on a countertop), not a material property. It belongs in build123d's `VisProperties` or as a per-call render hint to the adapters. If we add anything, it would be a `texture_transform=` kwarg on `to_gltf` / `to_threejs`. Happy to discuss if you want a sketch.

## Try it

```bash
pip install py-materials==3.8.0
```

3.7.0 added `Material.copy` / `with_vis` + 41 tests pinning the safe-derive contract. 3.8.0 adds the typed-kwarg layer + PEP 561 `py.typed` marker so your downstream type-checks pick it up.

Full CHANGELOG: <https://github.com/MorePET/mat/blob/main/CHANGELOG.md>

gerchowl:
PS: regarding the failing `tests (3.10, ubuntu-latest)` job — that's the mat-vis 3.10 break (`StrEnum` import in `mat_vis_client/schema.py`). Resolved upstream in mat-vis-client 0.6.3 (which py-materials 3.5+ pulls). A rebase / no-op push on `bernhard-42:materials` will pick up the fixed dep tree and clear the failure. (Ours didn't catch it earlier because we never import `mat_vis_client.adapters` directly; build123d's vis layer does at collect-time, so the failure mode only surfaces on your side. - thanks for the hit!)

bernhard-42:
Thanks @gerchowl for implementing the two changes.

Notes:
- `VisProperties.override`: Currently I apply the overrides to the exported `PbrProeprties` to have a single location for all incoming visualisation properties (mat-vis, threejs-materials for MaterialX and glTF). But I think, `override` is anyways a good method in py-materials/mat-vis
- `texture_scale`: agreed, not a generic material property

I worked with the new versions today and several issues surfaced:
- https://github.com/MorePET/mat-vis/issues/280
- https://github.com/MorePET/mat-vis/issues/281
- https://github.com/MorePET/mat-vis/issues/282
- https://github.com/MorePET/mat-vis/issues/283
- https://github.com/MorePET/mat-vis/issues/284
- https://github.com/MorePET/mat-vis/issues/285

Happy to discuss each of them in your repository, just flagging them here for visibility.

bernhard-42:
Note: The missing materials were actually a caching issue. I had to delete the old outdated cache of mat-vis, sorry

bernhard-42:
Now that I found the caching issue I could do more tests, leading to

- https://github.com/MorePET/mat-vis/issues/286
- https://github.com/MorePET/mat-vis/issues/287
- https://github.com/MorePET/mat-vis/issues/288

Also to be discussed in @gerchowl 's repo

bernhard-42:
Thanks for your quick fixes, @gerchowl, unfortunately, I had to reopen two issues as new issues, since they don't seem to be fixed (#312, #313).

But more importantly, I am not sure, the user experience of py-materials and mat-vis is already fit for purpose for build123d users. I have worked with py-materials and mat-vis now for several hours, and still struggle to get materials I find on the 4 supported webpages reliably downloaded. I have opened https://github.com/MorePET/mat-vis/issues/311 to explain what I mean.

Another very critical issue is https://github.com/MorePET/mat-vis/issues/285. I have spent days in building and fine tuning the threejs-materials baker so that:
- Materials look in three-cad-viewer like in the 3D previews of the material webpages
- Materials exported from build123d to glTF files look in different glTF viewers the same as in three-cad-viewer and the 3D previews of the material webpages.

As such
- the quality of the baked materials (https://github.com/MorePET/mat-vis/issues/285) 
- and the ease of use and reliability (https://github.com/MorePET/mat-vis/issues/311 )

are important topics to be resolved before I can recommend py-materials and mat-vis as foundation for the material system in build123d. 

**Note**, this is based on py-materials 3.10.0 and mat-vis 0.6.4 and the two libs are developing rapidly. I just want to emphasize the priority that I see for using them in build123d and I am more than happy to help getting the two dritical topics resolved

gerchowl:
Hi @bernhard-42 — landed my last comment without seeing yours; thanks for the careful pushback. You're right the UX bar isn't met yet from your vantage, and the four open items (mat-vis #285 baking, #311 usability, #312/#313 reopens of the closed-but-not-actually-fixed pair) are the bar.

Apologies for the rocky state. The build/CI/publish fabric needed more work than expected, which delayed the DX/UX investment where the actual bar sits. Pivoting there now.

Narrower readout of what did close in the recent batch: `Vis` is officially a public class with a stable contract (py-mat #98 / mat-vis #282), `Vis.tier=` validates at the assignment site, and the audit your usage triggered surfaced a small substrate-side cluster (mat-vis #298 + #302–#305) now filed upstream with concrete fixes. The two critical-path items (#285, #311) and the reopens (#312, #313) are work-in-progress upstream — engaging there.

The one open on the py-mat side (#99 — int color from `to_threejs`) is upstream too; routed as mat-vis #298 with a `color_format=Literal["hex","int","tuple"]` proposal defaulting to the Pythonic hex-string.

Thank you for the offer to help on the critical pair — open in the other direction too. Next stretch shifts to consumer DX/UX, closer to what your integration has been pushing on.

gerchowl:
Separately from the checkpoint above — opened a discussion on py-mat about the **contribution pathway for community-curated materials**: [MorePET/mat#218](https://github.com/MorePET/mat/issues/218). The question is which mode (ad-hoc per script / per-team TOML / upstream PR) build123d users would actually land at, and whether our current schema + licensing bar fits the long tail. No timeline; flagging here for visibility, happy to take it in py-mat's repo where it makes sense.

bernhard-42:
@gumyr fyi, I've merged `dev` today into `materials` to avoid that the `materials` branch and `dev` diverge too much.

But, the PR is not finalized, I will continue working with @gerchowl to make the material system stable

bernhard-42:
**Status update:** 
Back from vacation and checked the latest versions of `py-materials` and `mat-vis-client` on pypi. They are still 3.10 and 0.6.4. When I look into the repos, @gerchowl has changed quite some important stuff, hence waiting for the new releases on pypi.

gumyr:
I've found that this will be a common pattern:
```py
        self.color = Color(0xC0C0C0)
        self.material = metals.aluminum()
```
In the ocp_vscode Tree view if the color isn't set then the default yellow is used while in the Studio view the fancy material properties are used. I checked and `metals.aluminum().color` is `None` so that can't be used as a Tree view default. I realize that aluminum doesn't have an inherent color so this is probably the best that can be done.

gumyr:
I've found `auto_set_color` which looks like it resolves this issue. However, it's unclear to me how this works with a part library like bd_warehouse - is this set at the module level?

bernhard-42:
> I've found `auto_set_color` which looks like it resolves this issue. However, it's unclear to me how this works with a part library like bd_warehouse - is this set at the module level?

Yes, it is a global variable in shape_core.py and set at module level (`AUTO_COLOR = False`), controled by `auto_set_color`.

Not everyone might want to have colors auto set, hence I thought this is a good way to do so.

Are global variables OK from your side for use cases like this?

bernhard-42:
And as such it works wth bd_warehouse (just tested)

gumyr:
For viewing a global variable is probably ok as it isn't going to cause multi-threaded issues. Would it be possible for `metals.aluminum().color` to return this PBR color? If so a library could use this and get consistent results independent of what the user does with `auto_set_color`?

bernhard-42:
Currently you have to use

```python
In [20]: metals.aluminum().pbr.interpolate_color()
Out[20]: (0.9621, 0.9654, 0.9658, 1.0)
```

gumyr:
This is excellent:
```py
        self.material = metals.stainless()
        self.color = self.material.pbr.interpolate_color()
```

src/build123d/topology/shape_core.py
```diff
@@ -171,6 +175,21 @@
 CalcFn = Callable[[TopoDS_Shape, GProp_GProps], None]
 CompositeFactory = Callable[[Iterable["Shape"]], "Shape"]
 
+AUTO_COLOR = False
+
+
+def auto_set_color(flag: bool):
```
gumyr: This seems like more complex than what we need. How about having the material setter auto assign `color` to `material.pbr.interpolate_color()` but allowing the user to manually set color (as today) to override the value? So:
```py
shape.material = brass_material  # color auto-set from PBR
shape.color = "red"  # user overrides with explicit color
```
There would be no need for a global value or this `auto_set_color` function.

src/build123d/topology/shape_core.py
```diff
@@ -773,9 +830,9 @@ def combined_center(
         """
         objects = list(objects)
         if center_of == CenterOf.MASS:
-            total_mass = sum(Shape.compute_mass(o) for o in objects)
+            total_mass = sum(o.compute_volume() for o in objects)
             weighted_centers = [
```
gumyr, line 821: Shouldn't this use the density of the objects to truly "weight" the centre calculation?

src/build123d/exporters3d.py
```diff
@@ -259,7 +261,6 @@ def export_gltf(
     to_export: Shape,
     file_path: PathLike | str | bytes,
```
gumyr, line 262: Didn't we support BytesIO in non binary format - hence the need for the `binary` parameter. I don't think we should change this in this PR as it isn't related to materials.

src/build123d/topology/shape_core.py
```diff
@@ -366,6 +387,42 @@ def color(self, value: ColorLike | None) -> None:
...
+                    node_material = current_node._material
+                    break
+                current_node = current_node.parent
+        else:
+            node_material = self._material
+        self._material = node_material  # Set the node's material for next time
+        return node_material
+
+    @material.setter
+    def material(self, value: FinishedMaterial | None) -> None:
```
gumyr: Users are going to ask for a `MaterialLike` interface here, i.e. accept a str like "brass" and convert that into `material.brass()`. There may be a circular import type problem is this translation is provided by bd_materials.

src/build123d/build_enums.py
```diff
@@ -417,12 +417,15 @@ def __repr__(self):
 class Unit(Enum):
```
gumyr: I think we can use `from typing_extensions import StrEnum` to enable `class Unit(StrEnum)` with Python3.10.

src/build123d/geometry.py
```diff
@@ -118,6 +118,32 @@ def _canonical_quaternion_key(
     return _rounded_key(components, digits)
 
 
+MASS_UNIT = Unit.G
+LENGTH_UNIT = Unit.MM
+
+
+def get_units():
```
gumyr: These functions should be added to the formal docs via:
```py
.. autofunction:: build123d.geometry.get_units
.. autofunction:: build123d.geometry.set_units
```

These only apply to mass calculations, correct? For example if a user sets the units to meters does the constant `MM` change?

The material tutorials should clearly show the defaults units (via get_units) and how to set them to something else with set_units.

gumyr (changes requested):
This is quite a huge step forward for the project - thank you!

There are a few things that I think need to be resolved before merging this PR. The whole units thing is a can of worms.

src/build123d/topology/shape_core.py
bernhard-42: Should `MaterialLike` be similar to `ColorLike`

```python
MaterialLike: TypeAlias = (
    str  # name, ex: "brass"
    | FinishedMaterial
)
```
?

For this to work I need to enhance bd_materials to build a registry of all materials of all categories.
It assumes that there is no same material name in two different categories. This won't happen for the metals, but could be more likely for the abbreviations in plastics (I mean same abbreviation in another category meaning a different material)

src/build123d/topology/shape_core.py
bernhard-42: The reason why I added `auto_set_color` is that some materials/finishes have a white/white'ish interpolation color (chrome, pla, ...) and are nearly invisible with white background. For these materials one needs to set the color every time. If that is OK for you, I will remove the function and the global variable.

src/build123d/topology/shape_core.py
bernhard-42: Agreed, not sure why I changed this.

src/build123d/exporters3d.py
bernhard-42: ok, I roll it back

src/build123d/build_enums.py
bernhard-42: I look into it. 
Luckily Pythin 3.10 goes EOL in a few months (on Oct 31st 2026)

src/build123d/geometry.py
bernhard-42: `get_units` is only used in `Shape.compute_mass`, nowhere else.
So the constants are not changed (as they shouldn't as constants)

I must admit I struggle with the "unit-less" approach of build123d, while the default "somehow is g/mm".
For Jern build123d is unit-less (which is correct), for me I always work in g/mm and I have no idea what other users think/believe.

Like `auto_set_color`, `get_unit` and `set_unit` add complexity and were only needed because we want `.mass` to be a property (hence one cannot provide the unit of choice as a parameter).

Easiest solution: `def mass(mass_unit:Unit=Unit.G, length_unit:Unit=Unit.MM)` and live with `shape.mass()`
No `get_unit`, `set_unit`. And the defaults or set args of `mass` define in which units the mass should be provided. No side (possible) effects

But this is your call, I am happy to follow your guidance here.

bernhard-42:
> The whole units thing is a can of worms.

100% agreed, if we remove the requirement that `Shape.mass` needs to be a property, `set_units` and `get_units` can be removed, see above

src/build123d/geometry.py
jdegenstein: I am fine with `def mass(mass_unit:Unit=Unit.G, length_unit:Unit=Unit.MM)` -- it is simple enough and most importantly predictable

src/build123d/build_enums.py
jdegenstein: > I think we can use from typing_extensions import StrEnum to enable class Unit(StrEnum) with Python3.10.

This is not correct, it will raise an `ImportError`. The better solution is this polyfill:
```py
import sys

if sys.version_info >= (3, 11):
    from enum import StrEnum
else:
    from enum import Enum

    class StrEnum(str, Enum):
        """Polyfill for Python < 3.11 StrEnum behavior."""
        def __str__(self) -> str:
            return self.value
```

tests/test_materials.py
```diff
@@ -0,0 +1,403 @@
...
+#
+
+# Read densities straight from bd_materials (kg/m^3) so the tests track the
+# library's own data rather than hard-coded numbers.
+BRASS = metals.brass()
+WALNUT = wood.walnut()
+BRASS_DENSITY = BRASS.material.density  # kg/m^3
+WALNUT_DENSITY = WALNUT.material.density  # kg/m^3
+
+# Geometry: build123d models in mm; OCCT volume is unit-agnostic.
```
jdegenstein: > build123d models in mm 

Please consider rewording this

tests/test_materials.py
```diff
@@ -0,0 +1,403 @@
...
+    def test_metal_mechanical_ranges(self):
+        mat = self.box.material.material
+        self.assertIsInstance(mat.tensile_strength, Range)
+        self.assertLessEqual(mat.tensile_strength.min, mat.tensile_strength.max)
+        self.assertIsInstance(mat.modulus_of_elasticity, Range)
+        self.assertLessEqual(
+            mat.modulus_of_elasticity.min, mat.modulus_of_elasticity.max
+        )
+
+    def test_metal_thermal(self):
```
jdegenstein: recommend to rename to e.g. `test_metal_melt_temperature` we may in the future want thermal conductivity in bd_materials

pyproject.toml
```diff
@@ -54,6 +54,9 @@ dependencies = [
     
     # Install py-lib3mf ONLY on linux aarch64
     "py-lib3mf >= 2.4.1; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "pygltflib >= 1.16.5, < 1.17.0",
```
jdegenstein: From what I can tell I only see this used in the tests. If that is true maybe it is better for the option [development] category? Also, can this need not be fulfilled by OCCT/cadquery-ocp built-in GLTF support?

jdegenstein (commented):
Added a few more suggestions and question on pygltflib

gumyr:
I’m fine with mass as a method and not a property. If we’re going to fix this stuff now is the time.

src/build123d/topology/shape_core.py
bernhard-42: Removed `auto_set_color`.

`b.material = plastics.pla(color="cyan")` now sets color via `interpolate_color`.

Followed e.g. by `b.color = None` the original orange is applied in CAD view, followed by `b.color="red"` applies this color independent of the material.

src/build123d/topology/shape_core.py
bernhard-42: Now I remember: In the past it was called `compute_mass`, but it actually was the volume due to lack of defined density. 

To be backward compatible, I changed it to `compute_volume`

Plus, If we use `compute_mass` all shapes need to have a density (i.e. a material attached), else `combined_center` raises an issue.

@gumyr Do you prefer this to work independent of the density/material or should density be a prerequiste to using `combined_center`

src/build123d/exporters3d.py
bernhard-42: We do not support BytesIO with `export_gltf:

```python
def export_gltf(
    to_export: Shape,
    file_path: PathLike | str | bytes,
    unit: Unit = Unit.MM,
    binary: bool = False,
    linear_deflection: float = 0.001,
    angular_deflection: float = 0.1,
) -> bool:
```

compared to

```Python
def export_brep(
    to_export: Shape,
    file_path: PathLike | str | bytes | BytesIO | BinaryIO,
) -> bool:
```

I nevertheless rolled it back

src/build123d/build_enums.py
bernhard-42: `Unit` now is a `StrEnum` using this polyfill

src/build123d/geometry.py
bernhard-42: Removed `get_unit` and `set_units` and change `mass` accordingly in every class that defines it

tests/test_materials.py
bernhard-42: Done

pyproject.toml
bernhard-42: removed pygltflib dependency and converted tests to raw json and OCP based reader

tests/test_materials.py
bernhard-42: Renamed

src/build123d/topology/shape_core.py
bernhard-42: bd_materials now has functions `resolve(material_name)` and `material_names()`

```python
In [1]: from build123d import *

In [2]: b = Box(1,2,3)

In [3]: b.material="aluminum"

In [4]: b.material
Out[4]: FinishedMaterial(Alu_G6061_T6)

In [5]: b.material="aluminumm"
---------------------------------------------------------------------------
...
ValueError: unknown material 'aluminumm' -- did you mean aluminum? (bd_materials.material_names() lists all 155)
```

docs/tutorial_materials.rst
```diff
@@ -0,0 +1,307 @@
...
+
+    Note that density is a typical value for the material (to allow mass calculation in build123d), but all other properties are ranges for the typical values
+
+    These units are used for the properties:
+
+    .. code-block:: python
+
+        from bd_materials.core import PROPERTY_UNITS as pu
+        
+        pprint(pu)
```
jdegenstein, line 107: Please add an import for pprint

docs/tutorial_materials.rst
```diff
@@ -0,0 +1,307 @@
...
+            # hinge_outer.mass()=  136.993 g
+
+    -  The same model read as *in³*, mass reported in *lb*:
+
+        .. code-block:: python
+
+            print(f"{hinge_outer.mass(Unit.LB, Unit.IN)=:9.3f} lb")
+            # hinge_outer.mass(Unit.LB, Unit.IN)= 4949.191 lb
+
+    The volume number is identical in both cases — only its interpretation changes. One inch is 25.4 mm, so reading the model in inches makes it 25.4³ ≈ 16000 times heavier.
```
jdegenstein: revise to `≈ 16000 times more massive.`

docs/tutorial_materials.rst
```diff
@@ -0,0 +1,307 @@
...
+**Volume and mass calculation**
+
+- ``volume`` is a property and returns a bare number — it carries no unit. The number is the volume in whatever length unit the model was drawn in, cubed.
+
+    .. code-block:: python
+
+        print(f"{hinge_outer.volume=:9.3f}")
+        # hinge_outer.volume=16116.838
+
+- ``mass()`` multiplies that volume by ``<shape>.material.material.density``. Since ``volume`` is unitless, ``mass(mass_unit, length_unit)`` takes both the unit the volume should be read as and the unit the result is reported in. Both are optional and default to *gram* and *millimeter*. A shape without a material raises a ``ValueError``.
```
jdegenstein: Consider revising to:
```
- ``mass()`` multiplies the volume by ``<shape>.material.material.density``, which is always measured in kg/m³. Because ``volume`` is unitless, ``mass(mass_unit, length_unit)`` takes the output mass unit and the length unit used to interpret the volume. Both are optional and default to *gram* and *millimeter*, respectively. A shape without a material raises a ``ValueError``.
```

jdegenstein (commented):
added a few extra suggestions to tutorial_material

gumyr:
Huge Thank You!

PR #1272, opened 2026-04-09
Add UV texture coordinates for OBJ and glTF exports

zalo:
## Summary

Ports [CascadeStudio's UV mapping pipeline](https://github.com/zalo/CascadeStudio) to build123d, enabling texture mapping, UV baking, and displacement-map workflows like [BumpMesh](https://bumpmesh.com/) (a service for adding displacement heightmap tessellations to 3D-printable meshes) that can take advantage of mesh UVs.

- **`Shape.tessellate_with_uvs()`** — extracts per-vertex normals and UV coordinates from OCCT's surface parameterization, with optional atlas packing (shelf-based bin packing preserving physical aspect ratios)
- **`export_obj()`** — new Wavefront OBJ exporter with `vt`/`vn` records and atlas-packed UVs
- **`include_uvs`** parameter on both `export_obj()` and `export_gltf()` (default `True`) to optionally disable UV output
- **`SetForcedUVExport`** on `RWGltf_CafWriter` enables `TEXCOORD_0` accessors in glTF output

### UV Baking

The atlas-packed UVs are designed for **UV baking** workflows: since all faces are packed into a single non-overlapping `[0,1]` texture atlas, you can bake lighting, ambient occlusion, or displacement maps directly onto the UV layout. This is the same approach used by game engines and tools like Blender's bake system.

### Performance

Uses numpy bulk transforms on `MapNodeArray`/`MapUVNodeArray` for the vertex/normal extraction hot path. The overhead vs plain `tessellate()` is ~1.9x, dominated by per-element OCP Python/C++ binding calls (not Python math):

| Shape | Verts | tessellate | + UVs + atlas | Overhead |
|-------|-------|-----------|---------------|----------|
| Box | 24 | 3.6ms | 4.2ms | 1.2x |
| Cylinder | 506 | 7.1ms | 14.2ms | 2.0x |
| Sphere | 4,066 | 55ms | 105ms | 1.9x |
| Torus | 16,129 | 210ms | 392ms | 1.9x |

### Visualization

Debug visualizations showing atlas layout, 3D checkerboard texture mapping (with proper barycentric interpolation), per-face UV parameterization, and UV distortion histograms:

image

image

image

image

## Test plan

- [x] `TestTessellateWithUVs` — 6 tests covering box/cylinder/sphere, atlas and no-atlas UV ranges, triangle index validity
- [x] `TestExportObj` — 4 tests covering OBJ output structure, index validity, no-atlas mode, and `include_uvs=False`
- [x] `include_uvs=False` produces valid OBJ without `vt`/`vn` lines and plain `f a b c` indices
- [x] Verify glTF `include_uvs=True` produces `TEXCOORD_0` accessors via `SetForcedUVExport`
- [x] Verify glTF `include_uvs=False` strips `TEXCOORD_0` from output


src/build123d/exporters3d.py
```diff
@@ -452,3 +477,74 @@ def export_stl(
 
     writer.ASCIIMode = ascii_format
     return writer.Write(to_export.wrapped, fsdecode(file_path))
+
+
+def export_obj(
+    to_export: Shape,
+    file_path: PathLike | str | bytes,
```
jdegenstein: Can you add `| BytesIO` support to this exporter? I want all exporters to support this (when possible, currently not possible for GLTF/GLB). Further details here https://github.com/gumyr/build123d/issues/1024 with linked PR for reference. The primary reason is that this makes it easier to support this format in pyodide/WASM environments.

FYI OCCT does seem to have OBJ export support but likely does not have BytesIO support so I am OK with your approach here.

zalo: Sounds good.

If you don’t mind the dependency, three.js has an excellent GLTF / GLB Exporter, which could be used in the contexts where WASM is used.

zalo:
All set @jdegenstein 

If you’d like to test it on your own; I tend to enjoy using Windows’ first party 3D Viewer for checking out UVs mapped to meshes.

src/build123d/topology/shape_core.py
```diff
@@ -2138,6 +2139,256 @@ def tessellate(
...
+                    u, v = all_uvs[j]
+                    all_uvs[j] = (
+                        (u * bw + bx) * inv_scale,
+                        (v * bh + by) * inv_scale,
+                    )
+
+        return all_vertices, all_triangles, all_normals, all_uvs
+
+    @staticmethod
+    def _arc_length(
```
gumyr: We should use the OCCT functionality for this as shown here:
```py
from math import pi
from build123d import *
from OCP.GeomAdaptor import GeomAdaptor_Curve
from OCP.GCPnts import GCPnts_AbscissaPoint
from OCP.BRep import BRep_Tool


def _arc_length(
    adaptor: GeomAdaptor_Curve,
    param_min: float,
    param_max: float,
    tolerance: float = 1e-7,
) -> float:
    return GCPnts_AbscissaPoint.Length_s(
        adaptor,
        param_min,
        param_max,
        tolerance,
    )


arc = CenterArc((0, 0), 1, 0, 180)
comp_curve = arc.geom_adaptor()
param_min, param_max = BRep_Tool.Range_s(arc.wrapped)
print(param_min, param_max, arc.length)  # 0.0 3.141592653589793 3.141592653589793
partial_length = _arc_length(arc.geom_adaptor(), pi / 4, 3 * pi / 4)
print(partial_length, partial_length / pi)  # 1.5707963267948966 0.5
```

src/build123d/topology/shape_core.py
```diff
@@ -2138,6 +2139,256 @@ def tessellate(
...
+                    surface = BRep_Tool.Surface_s(face.wrapped)
+                    u_center = (u_min + u_max) * 0.5
+                    v_center = (v_min + v_max) * 0.5
+                    viso = surface.VIso(v_center)
+                    uiso = surface.UIso(u_center)
+                    u_adaptor = GeomAdaptor_Curve(viso)
+                    v_adaptor = GeomAdaptor_Curve(uiso)
+                    arc_w = self._arc_length(u_adaptor, u_min, u_max)
+                    arc_h = self._arc_length(v_adaptor, v_min, v_max)
+                except Exception:
```
gumyr: Best not to use a bare exception here.

src/build123d/topology/shape_core.py
```diff
@@ -2138,6 +2139,256 @@ def tessellate(
...
+                    n1, n2 = n2, n1
+                all_triangles.append(
+                    (n1 + offset - 1, n2 + offset - 1, n3 + offset - 1)
+                )
+
+            offset += nb_nodes
+
+        # Atlas packing: scale per-face UVs to world-space proportions and pack
+        if atlas_packing and face_uv_ranges:
+            padding = 2.0
```
gumyr: Why is `2` used here? Is there some reason that 2 is always the right amount of padding?

src/build123d/topology/shape_core.py
```diff
@@ -2138,6 +2139,256 @@ def tessellate(
...
+            boxes = [
+                {"w": r["w"] + padding, "h": r["h"] + padding, "index": i}
+                for i, r in enumerate(face_uv_ranges)
+            ]
+            pack_w, pack_h = self._potpack(boxes)
+            scale = max(pack_w, pack_h)
+            if scale < 1e-15:
```
gumyr: 1e-15 is many orders of magnitude below the current TOLERANCE limit of 1e-6 so this is essentially 0.

src/build123d/topology/shape_core.py
```diff
@@ -2138,6 +2139,256 @@ def tessellate(
...
+            adaptor.D0(s, pnt)
+            cur = (pnt.X(), pnt.Y(), pnt.Z())
+            if prev is not None:
+                dx, dy, dz = cur[0] - prev[0], cur[1] - prev[1], cur[2] - prev[2]
+                length += (dx * dx + dy * dy + dz * dz) ** 0.5
+            prev = cur
+        return length
+
+    @staticmethod
+    def _potpack(boxes: list[dict]) -> tuple[float, float]:
```
gumyr: build123d already has a packing algorithm that seems similar to this: src/build123d/src/build123d/pack.py  With a little bit of work it could be made into a utility that shape_core could use to avoid having duplicate code. Is this worth doing?

src/build123d/topology/shape_core.py
```diff
@@ -2138,6 +2139,256 @@ def tessellate(
...
+                normalized to [0, 1].
+
+        Returns:
+            A 4-tuple of (vertices, triangles, normals, uvs) where:
+            - vertices: list of Vector positions
+            - triangles: list of (i0, i1, i2) index triples
+            - normals: list of Vector per-vertex normals
+            - uvs: list of (u, v) texture coordinates per vertex
+        """
+        import numpy as np
```
gumyr: geometry.py (which is a dependent of shape_core) already imports numpy so this should be moved to the top of the module with the other imports as late loading doesn't apply.

gumyr (dismissed):
Thank you - this is quite a significant bit of added functionality. The docs will need to be updated at some point to show this new export capability but that isn't necessarily required in this PR.

gumyr:
Thank you! This is a large addition to the functionality of build123d - go make all the bumpy things :)

PR #1275, opened 2026-04-14
Add error handling for font file loading in register_font

karugaru:
When loading fonts, if corrupted fonts are present, even valid fonts will cause the operation to fail.

When corrupted fonts are detected, log a warning and skip them instead.


gumyr:
Although there were some issues with the code the problem and PR was taken into account with this commit: 0e7f401545ea423d305f6af76c529cb45f88ffbf where credit was given "Based on the original proposal by @karugaru in https://github.com/gumyr/build123d/pull/1275." Thanks for the PR.

karugaru:
Thank you!

PR #1276, opened 2026-04-15
feat: accept pymat.Material on Compound/Solid.material (draft, refs MorePET/mat#3)

gerchowl:
## Draft — exploratory

Opened as a **draft** in response to the feature request in [MorePET/mat#3](https://github.com/MorePET/mat/issues/3), where a user proposed that build123d shapes carry \`pymat.Material\` instances on \`.material\` for unified physics + PBR rendering.

**Not ready to merge** until:
1. The py-materials side ([MorePET/mat#30](https://github.com/MorePET/mat/pull/30)) lands and ships as py-materials 2.2.0 on PyPI
2. The \`materials\` extra's git+https pin is switched to a PyPI version pin
3. The API-collision resolution (type widen — see below) is reviewed
4. @bernhard-42's \`threejs-materials\` side is confirmed as the canonical PBR loader

## What this does

Evolves the existing \`Compound.material\` / \`Solid.material\` attribute (currently a free-form \`str\` tag for external-tool export) to **also** accept \`pymat.Material\` instances. Type-widened, not renamed — existing str usage stays backward compatible.

\`\`\`python
from build123d import Box
from pymat import Material
from pymat.pbr import PbrProperties  # requires build123d[materials]

box = Box(50, 50, 10)
box.material = Material(
    name="Stainless Steel 304",
    density=8.0,
    formula="Fe",
    pbr_source=PbrProperties.from_gpuopen("Stainless Steel Brushed"),
)

# One object, two consumers:
box.material.density         # → 8.0    (physics: mass calc)
box.material.molar_mass      # → 55.85  (physics: radiation transport)
box.material.to_three_js_material_dict()  # → Three.js render dict (PBR)
\`\`\`

## Why type widen instead of replace

\`Compound.material\` and \`Solid.material\` currently exist as \`str = ""\` — documented as "tag for external tools". Replacing the type outright would break STEP/STL export code that writes those tags. Widening to \`str | pymat.Material | None\` preserves every existing consumer and lets downstream code \`isinstance\`-check to distinguish.

If you prefer a different resolution (rename the legacy attribute, keep them separate, deprecation path), happy to rework — the goal here is to surface the collision concretely so the design decision is grounded in real code.

## Changes

- \`src/build123d/topology/composite.py\`: \`Compound.__init__\` material type widened + docstring. TYPE_CHECKING-only import of \`pymat.Material\`.
- \`src/build123d/topology/three_d.py\`: \`Solid.__init__\` matches. Same TYPE_CHECKING-only import.
- \`pyproject.toml\`: new \`[materials]\` optional extra pulling \`py-materials[pbr]\` (git+https to the exploration branch for now — will switch to PyPI pin before merge).
- \`examples/pbr_material_pymat.py\`: end-to-end demo — creates a machined flange part, attaches a stainless steel Material, prints physics values, geometry-derived mass, and the Three.js PBR dict. \`--visual\` flag calls \`ocp_vscode.show()\` for live rendering.

## Runtime dep posture

**Zero runtime dependency on py-materials** in the default install. The \`PymatMaterial\` import is \`TYPE_CHECKING\`-only. Users without \`[materials]\` installed keep using \`shape.material = "some-tag"\` as they do today. Only users who want the unified physics + PBR story pay the install cost:

\`\`\`bash
pip install build123d[materials]           # + py-materials[pbr]
pip install build123d[materials,ocp_vscode]  # + live rendering
\`\`\`

## Verified end-to-end

\`\`\`text
$ python examples/pbr_material_pymat.py
=== Physics properties (via py-materials) ===
  name:         Stainless Steel 304
  density:      8.0 g/cm³
  formula:      Fe
  molar mass:   55.85 g/mol
  part volume:  21.99 cm³
  part mass:    175.92 g

=== PBR rendering (Three.js MeshPhysicalMaterial dict) ===
{ "color": [0.75, 0.75, 0.77], "metalness": 1.0, "roughness": 0.35 }

=== Backend summary ===
  rich PBR available:    False
  rich backend in use:   False
\`\`\`

(Rich backend path is verified manually with \`pip install py-materials[pbr]\` installed — downloads Stainless Steel Brushed MaterialX textures from matlib.gpuopen.com on first run and caches them.)

## Refs

- [MorePET/mat#3](https://github.com/MorePET/mat/issues/3) — the feature request, design discussion, ADR-style analysis
- [MorePET/mat#30](https://github.com/MorePET/mat/pull/30) — py-materials side (matching draft PR)
- [bernhard-42/threejs-materials](https://github.com/bernhard-42/threejs-materials) — the PBR loader this integration ultimately feeds

cc @bernhard-42 (threejs-materials + ocp_vscode)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,12 +47,11 @@ dependencies = [
     "scipy",
     "scikit-learn >= 1.5, < 2",
     "webcolors ~= 24.8.0",
-
     # Install standard official lib3mf on everything EXCEPT linux aarch64
     "lib3mf >= 2.4.1; sys_platform != 'linux' or platform_machine != 'aarch64'",
-    
     # Install py-lib3mf ONLY on linux aarch64
     "py-lib3mf >= 2.4.1; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "ocp-vscode>=3.3.4",
 ]
 
 [project.urls]
@@ -72,12 +71,13 @@ ocp_vscode = [
 # extra installed, those attributes can carry a `pymat.Material` instance
 # in addition to the legacy str tag.
 #
-# EXPLORATION: this pin points at the pymat feature branch (MorePET/mat#30)
-# where the pbr Protocol lives. Before proposing this change upstream to
-# gumyr/build123d, switch to a PyPI version pin — `py-materials[pbr]>=2.2.0`
-# once the pbr extra ships in a PyPI release.
+# EXPLORATION: pinned to the exact commit on the pymat feature branch
+# (MorePET/mat#30) where the pbr Protocol lives. SHA pin is deliberate —
+# before proposing this change upstream to gumyr/build123d, switch to a
+# PyPI version pin (`py-materials[pbr]>=2.2.0`) once the pbr extra ships
+# in a PyPI release.
 materials = [
-    "py-materials[pbr] @ git+https://github.com/MorePET/mat.git@feature/3-pbr-protocol-integration",
+    "py-materials[pbr] @ git+https://github.com/MorePET/mat.git@8c7729c",
 ]
 
 # development dependencies
@@ -115,6 +115,17 @@ all = [
     "build123d[stubs]",
     "build123d[materials]",
 ]
+
+[tool.uv.sources]
+# EXPLORATION: pin ocp_vscode to the fork commit carrying the
+# pymat.pbr_source adapter fix, pending the upstream PR at
+# bernhard-42/vscode-ocp-cad-viewer#228.
+# Revert to the PyPI `ocp_vscode` dep once the fix is released.
+#
+# NOTE: this section is uv-specific. If you use `pip` or `poetry`,
+# you'll need to install ocp_vscode from the fork manually:
+#   pip install "ocp-vscode @ git+https://github.com/gerchowl/vscode-ocp-cad-viewer.git@94cecc7"
+ocp-vscode = { git = "https://github.com/gerchowl/vscode-ocp-cad-viewer.git", rev = "94cecc7" }
 
 [tool.setuptools.packages.find]
 where = ["src"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,19 @@ ocp_vscode = [
     "ocp_vscode",
 ]
 
+# enable the optional py-materials (physics + PBR rendering) integration
+# See docstrings on `Compound.material` and `Solid.material` — with this
+# extra installed, those attributes can carry a `pymat.Material` instance
+# in addition to the legacy str tag.
+#
+# EXPLORATION: this pin points at the pymat feature branch (MorePET/mat#30)
+# where the pbr Protocol lives. Before proposing this change upstream to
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -47,12 +47,11 @@ dependencies = [
     "scipy",
     "scikit-learn >= 1.5, < 2",
     "webcolors ~= 24.8.0",
-
     # Install standard official lib3mf on everything EXCEPT linux aarch64
     "lib3mf >= 2.4.1; sys_platform != 'linux' or platform_machine != 'aarch64'",
-    
     # Install py-lib3mf ONLY on linux aarch64
     "py-lib3mf >= 2.4.1; sys_platform == 'linux' and platform_machine == 'aarch64'",
+    "ocp-vscode>=3.3.4",
```
jdegenstein: I am not sure I support adding ocp-vscode as a required dep of build123d. Historically we have avoided this as not all users of build123d use this viewer. Changing this approach is a significant deviation. This also has some negative side-effects for wasm / pyodide because pyperclip (dep of ocp_vscode) does not work for that platform, which will require extra workarounds with mock packages for those users.

jdegenstein (changes requested):
Added one comment so far.

bernhard-42:
We anyhow need to discuss this in the light of https://github.com/gumyr/build123d/pull/1270 
Maybe this PR has some things that could be changed in 1270.

We already started discussion on 1270 with @jdegenstein and @gumyr Roger and according to this discussion 1270 adds some more changes like improved mass/volume calculation

@jdegenstein  I agree, ocp vscode should not be a required dependency of build123d

gerchowl:
@jdegenstein — fair call, apologies. The `ocp-vscode>=3.3.4` entry in main `dependencies` was a leftover from a mid-session edit that got bundled into this PR by accident. Reverted in b7e0ffe5 — ocp-vscode is back to only appearing in the existing `[ocp_vscode]` optional extra, matching upstream. The example at `examples/pbr_material_pymat.py` already documents `uv sync --extra materials --extra ocp_vscode` as the install path for anyone wanting to reproduce it, so the optional-extra approach works for the exploratory flow.

@bernhard-42 — thanks for pointing at #1270. I completely missed that PR when I opened this one, and for that matter I missed your detailed comments on [MorePET/mat#3](https://github.com/MorePET/mat/issues/3) before posting the PoC summary there. Apologies — I should have refreshed both threads before acting. Leaving a longer substantive reply on MorePET/mat#3 shortly; it's the right place for the design discussion.

TL;DR: this PR was intended as a "minimal-effort wiring" exploration of what an integration could look like, not as a proposal that competes with gumyr/build123d#1270. After reading gumyr/build123d#1270 in depth, your design is cleaner on the decoupling story (build123d's `Material` wrapper keeps py-materials and threejs-materials independent; ours couples py-materials to a rendering optional-extra). Happy to close this one and MorePET/mat#30 / bernhard-42/vscode-ocp-cad-viewer#228 once the collaboration direction is settled on MorePET/mat#3, and redirect effort to helping land gumyr/build123d#1270.

gumyr:
Thank you for all of the effort that went into integrating pymat and build123d. However, we've decided to go with a less complex, build123d specific material library: bd_materials.

PR #1281, opened 2026-04-18
fix `Plane.__mul__` typing and improve testing

snoyer:
I just wanted to fix `Plane * Location -> Plane` but had to touch up the impl and testing.
The implementation didn't exactly agree line up with the declared type so I went with what I assumed it meant. Let me know if I messed it up.

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -915,12 +915,6 @@ class LocationList:
         self.location_index += 1
         return result
 
-    def __mul__(self, shape: Shape) -> list[Shape]:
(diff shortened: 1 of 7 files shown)


snoyer:
Added some more refactoring using `__rmul__` on the `Shape` side so that we can still do `Location * Shape` and `Plane * Shape` without `Location` or `Plane` having to know about `Shape`.

snoyer:
failing test is about fonts, so unrelated to these changes

gumyr:
Thank you!

PR #1282, opened 2026-04-20
make `Plane` moveable like `Shape`

snoyer:
this PR reworks `Plane` multiplication so that it behaves similarly to `Shapes`'s, as per recent discussion on Discord.

The intentional breaking change here is that to apply a location to a plane and get a moved plane, one now has to do `location * plane` (same as `location * shape` to get a moved shape) instead of `plane * location` (which now results in a location)

with this PR:
image

without this PR:
image

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -3050,12 +3050,11 @@ class Plane(metaclass=PlaneMeta):
             return self.moved(other)
         try:
             others = list(other)
-            if any(not isinstance(other, Location | Plane) for other in others):
-                raise ValueError("Planes can only be multiplied by locations or planes")
-            return [self.moved(loc) for loc in others]
+            if all(isinstance(other, Location | Plane) for other in others):
+                return [self.moved(loc) for loc in others]
         except TypeError:  # not iterable
             pass
-        raise TypeError()
+        raise TypeError("Planes can only be multiplied by locations or planes")
 
     def __and__(self: Plane, other: Axis | Location | Plane | VectorLike | Shape):
         """intersect plane with other &"""
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -3047,12 +3047,12 @@ class Plane(metaclass=PlaneMeta):
         self, other: Location | Plane | Iterable[Location | Plane]
     ) -> Plane | list[Plane]:
         if isinstance(other, Location | Plane):
-            return self.move(other)
+            return self.moved(other)
         try:
             others = list(other)
             if any(not isinstance(other, Location | Plane) for other in others):
                 raise ValueError("Planes can only be multiplied by locations or planes")
-            return [self.move(loc) for loc in others]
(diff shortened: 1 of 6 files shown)


src/build123d/geometry.py
```diff
@@ -3151,18 +3177,19 @@ def rotated(
 
         return Plane(self._origin, new_x_dir, new_z_dir)
 
-    def move(self, loc: Location) -> Plane:
-        """Change the position & orientation of self by applying a relative location
+    # TODO should be called `moved` as it makes a new Plane?
+    def move(self, loc: Location | Plane) -> Plane:
+        """Change the position & orientation of a copy of self by applying a relative location
```
snoyer, line 3182: This `move` method returns a new object, should it be renamed `moved` to match `Shape.moved`? https://github.com/gumyr/build123d/blob/d72084f4a51a2693b586fea619d8b38f959a1554/src/build123d/topology/shape_core.py#L1591

gumyr: There is an incomplete PR to fix the Plane.move/Plane.moved issue: https://github.com/gumyr/build123d/pull/444 so yes it should be fixed.

src/build123d/geometry.py
```diff
@@ -3017,19 +3017,45 @@ def __neg__(self) -> Plane:
...
+        if isinstance(other, Location | Plane):
+            return self.move(other)
+        try:
+            others = list(other)
+            if any(not isinstance(other, Location | Plane) for other in others):
+                raise ValueError("Planes can only be multiplied by locations or planes")
+            return [self.move(loc) for loc in others]
+        except TypeError:  # not iterable
+            pass
+        raise TypeError()
```
gumyr: How about something like:
```py
raise TypeError(
    f"Plane.__rmul__() requires Location or Plane, not {type(other).__name__}"
)
```

gumyr (changes requested):

src/build123d/geometry.py
snoyer: done

src/build123d/geometry.py
snoyer: I'll set the message to match the one from `Shape` https://github.com/gumyr/build123d/blob/d72084f4a51a2693b586fea619d8b38f959a1554/src/build123d/topology/shape_core.py#L1020 for now

I have just noticed that Shapes raises `ValueError` where it should probably raise `TypeError` so that would need to be uniformized.

Do you want me to change to `TypeError` in this PR or a later one? if I should do it now do you want to have the `not {type(...)}` part added to both?

gumyr: Uniformization in this PR would be great. Adding the extra type info might help users figure out want went wrong.

snoyer: went with `A cannot be multiplied by C, D, ...` to keep it short, what it _can_ be multiplied by (Plane, Location) would be in the docstrings, types, ...

PR #1283, opened 2026-04-21
change `.wrapped` attrs to read-only props

snoyer:
Change the public `.wrapped` attribute to a private `._wrapped` attribute and add a `.wrapped` read-only property to the following classes: `Axis`, `Location`, `OrientedBoundBox`.

This improves the type checking and removes the need for handling and testing of a bunch of unnecessary `.wrapped is None`.

I'm planning on also cleaning up the `Plane` class but I'll do that in a _separate_ PR _after_ #1282 is merged

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -637,24 +637,23 @@ class Axis(metaclass=AxisMeta):
     _dim = 1
 
     @overload
-    def __init__(self, gp_ax1: gp_Ax1):
+    def __init__(self, gp_ax1: gp_Ax1) -> None:
         """Axis: point and direction"""
 
     @overload
-    def __init__(self, location: Location):
+    def __init__(self, location: Location) -> None:
         """Axis from location"""
 
     @overload
-    def __init__(self, origin: VectorLike, direction: VectorLike):
+    def __init__(self, origin: VectorLike, direction: VectorLike) -> None:
         """Axis: point and direction"""
 
     @overload
-    def __init__(self, edge: Edge):
+    def __init__(self, edge: Edge) -> None:
         """Axis: start of Edge"""
 
-    def __init__(
-        self, *args, **kwargs
-    ):  # pylint: disable=too-many-branches, too-many-locals
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        # pylint: disable=too-many-branches, too-many-locals
 
         gp_ax1 = kwargs.pop("gp_ax1", None)
         origin = kwargs.pop("origin", None)
@@ -714,7 +713,11 @@ class Axis(metaclass=AxisMeta):
         elif not isinstance(gp_ax1, gp_Ax1):
             raise ValueError(f"Invalid Axis parameter: {gp_ax1}")
 
-        self.wrapped: gp_Ax1 = gp_ax1  # type: ignore[annotation-unchecked]
+        self._wrapped: gp_Ax1 = gp_ax1
+
+    @property
+    def wrapped(self):
+        return self._wrapped
 
     @property
     def position(self) -> Vector:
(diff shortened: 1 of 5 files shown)


gumyr:
In general I like this change but the loss of the error messages might be problematic . What if we centralize the error checks and error message generation in geometry.py and update all the wrapped properties to use it - something like this:

```py
# geometry.py

import inspect
from typing import Any, NoReturn

_WRAPPED_OPERATION_NAMES = {
    "__add__": "combine",
    "__sub__": "cut",
    "__and__": "intersect",
    "__or__": "fuse",
    "__rmul__": "move",
    "__mul__": "multiply",
    "locate": "locate",
    "located": "locate",
    "move": "move",
    "moved": "move",
    "relocate": "relocate",
    "bounding_box": "calculate the bounding box of",
    "oriented_bounding_box": "calculate the oriented bounding box of",
    "center": "calculate the center of",
    "distance_to": "calculate distance from",
    "distance_to_with_closest_points": "calculate distance from",
    "intersect": "intersect",
    "faces": "get faces from",
    "edges": "get edges from",
    "vertices": "get vertices from",
    "wires": "get wires from",
    "solids": "get solids from",
    "shells": "get shells from",
}

def _wrapped_caller_operation(obj: object) -> str:
    """Best-effort name of the user-level method that accessed wrapped.

    This is only used while raising an error, so stack inspection cost is not on
    the normal execution path.
    """
    frame = inspect.currentframe()
    if frame is None:
        return "operate on"

    # _wrapped_caller_operation -> empty_wrapped_error -> wrapped
    frame = frame.f_back
    frame = frame.f_back if frame is not None else None
    frame = frame.f_back if frame is not None else None

    while frame is not None:
        if frame.f_locals.get("self") is obj and frame.f_code.co_name != "wrapped":
            method_name = frame.f_code.co_name
            return _WRAPPED_OPERATION_NAMES.get(
                method_name, method_name.replace("_", " ")
            )
        frame = frame.f_back

    return "operate on"

def empty_wrapped_error(obj: object) -> ValueError:
    """Create a user-facing error for objects whose internal wrapped value is empty."""
    operation = _wrapped_caller_operation(obj)
    return ValueError(
        f"Cannot {operation} an empty {type(obj).__name__}. "
        "The object has no underlying geometry."
    )

def raise_empty_wrapped(obj: object) -> NoReturn:
    """Raise a user-facing error for objects whose internal wrapped value is empty."""
    raise empty_wrapped_error(obj)
```
Then `Shape.wrapped` can become:
```py
from build123d.geometry import raise_empty_wrapped

@property
def wrapped(self):
    if self._wrapped is None:
        raise_empty_wrapped(self)
    return self._wrapped
```
And geometry classes can use the same function:
```py
@property
def wrapped(self) -> TopLoc_Location:
    if getattr(self, "_wrapped", None) is None:
        raise_empty_wrapped(self)
    return self._wrapped
```
Example error messages:

- Cannot move an empty Edge. The object has no underlying geometry.
- Cannot calculate the bounding box of an empty Compound. The object has no underlying geometry.
- Cannot operate on an empty Location. The object has no underlying geometry.

What do you think? Maybe this PR should be merged as is and this or something like it added as a separate change?

snoyer:
The goal wish this PR is that:
- the constructors ensure they do not leave `._wrapped == None`
- `.wrapped` cannot be set to `None` from the outside because it's a read only property

So unless I've missed something all of theses checks are not needed.

Are there still cases where `.wrapped` is `None`?

gumyr:
For the geometric objects I don't think there is a way for them to be None. For the Shape objects yes they could be. Since the check error message code could be common I didn't think there was much overhead in checking the geometric objects as well.

gumyr:
Thinking about this some more. The enhanced error messages should definitely not be part of this PR. There is the open review of the use of None - maybe that will essentially make None impossible across the board which makes the checking and error reporting obsolete.

snoyer:
I agree, I'd personally rather not add introspective stuff unless absolutely necessary and I'm hoping this whole `.wrapped is None` thing is avoidable by design (like, what does it even _mean_ to have a valid object with `.wrapped` being `None`?)

PR #1285, opened 2026-04-21
drop stateful `__iter__` implementations

snoyer:
Refactor `__iter__` implementations by returning an iterator to the underlying structures instead of managing the state internally in order to simplify the code.

(Unless there's an unstated and untested reason to do it the stateful way, in which case let me know)


gumyr:
Thank you!

PR #1288, opened 2026-04-22
make `Plane` use properties and not attributes

snoyer:
Refactor `Plane` class to use property backed by the wrapped `gp_Pln` class instead of having it's own attributes that can be modified or otherwise go out of sync.
Also ensure that the underlying `gp_Pln` is always right-handed as per the specification of the `Plane` class.


gumyr:
Thank you!

PR #1291, opened 2026-04-24
use `@deprecated` for methods and classes

snoyer:
Simplifies the code and allow typecheckers/IDE to see deprecated classes and methods before the script runs

image

### DIFF
--- a/src/build123d/topology/shape_core.py
+++ b/src/build123d/topology/shape_core.py
@@ -131,7 +131,7 @@ from OCP.TopTools import (
     TopTools_ListOfShape,
     TopTools_SequenceOfShape,
 )
-from typing_extensions import Self
+from typing_extensions import Self, deprecated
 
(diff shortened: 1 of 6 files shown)


src/build123d/topology/shape_core.py
```diff
@@ -452,14 +452,12 @@ def is_null(self) -> bool:
         return self._wrapped is None or self.wrapped.IsNull()
 
     @property
+    @deprecated(
+        "The ``is_planar_face`` property is deprecated and will be removed in a future version."
+        " Use ``Face.is_planar`` instead"
```
jdegenstein: Can you change these two to use single-quote instead of double back-ticks? I believe the current behavior is somehow motivated by RST compatibility but I don't see how that is needed here in a deprecated decorator / former warning.

jdegenstein:
Added one small thing to change -- once completed this looks good to me.

gumyr:
Thank you!

PR #1297, opened 2026-05-03
export `Shape`

snoyer:
Fixes an import warning for my fellow type-checking enthusiasts
Screenshot From 2026-05-03 07-42-52


gumyr:
Users shouldn't need to use `Shape` that's why it wasn't put in the __init__ .py in the first place. Why do you need it?

snoyer:
Alright, I thought it was some oversight but that makes sense, my bad.
I only use it to type some boilerplate code so never mind me :)

PR #1302, opened 2026-05-04
Fix issue 1294

### LINKED ISSUE #1294: `ConstrainedArcs` solutions parameters offset
Putting together an implementation of `full_round()` using `ConstrainedArcs()` I stumbled upon this. It appears that, if the input edge is a circle and the solution circle is inside it, that solution circle has its parameters shifted by `pi`.
This fails as it is now:
```python
with BuildLine() as l:
    ln1 = JernArc((0, 0), (1, 0), 20, 90)
    ln2 = JernArc((0, 10), (1, 0), 10, 90)
    ln3 = Line(ln1 @ 1, ln2 @ 1)
    ln4 = ConstrainedArcs(ln1, ln2, ln3, sagitta=Sagitta.BOTH)
```
The changes to `constrained_lines.py` in the attached patch seem to address this problem.

[constrained_lines.py.3tan.patch](https://github.com/user-attachments/files/27247810/constrained_lines.py.3tan.patch)

javimixet:
Fixes [#1294](https://github.com/gumyr/build123d/issues/1294).
When generating constrained tangent arcs, the tangency parameter on the solution circle must be shifted by `pi` if the generated circle is fully enclosed within one of the target tangent circular edges. 

* Added `_enclosed_circ_param_offset` to automatically detect when a solution circle is enclosed by a circular tangency target and apply the necessary `pi` offset to its parameters.
* Applied the offset correction to the `_make_2tan_on_arcs` and `_make_3tan_arcs` solver functions.


javimixet:
### Summary of Changes

*   **Corrected Arc Trimming for Enclosed Circles**: 
    *   Introduced `_enclosed_circ_param_offset` in `src/build123d/topology/constrained_lines.py` to adjust solution parameters when a circle is enclosed within a tangent circular edge.
    *   Applied this logic to `_make_2tan_on_arcs` and `_make_3tan_arcs` to ensure correct arc segments are selected.

*   **Enhanced Arc Selection (`sagitta` logic)**:
    *   Updated `ConstrainedArcs` and `Edge.make_constrained_arcs` to make the `sagitta` parameter optional (`None` by default).
    *   When `sagitta` is `None` in `_make_3tan_arcs`, the system now automatically selects the specific arc segment that touches all three input curves.

*   **API & Documentation Updates**:
    *   Updated type hints and docstrings in `src/build123d/objects_curve.py` and `src/build123d/topology/one_d.py` to reflect the new optional `sagitta` behavior.

*   **Validation**:
    *   Added new test cases in `tests/test_direct_api/test_constrained_arcs.py` (`test_make_constrained_arcs_3tan` and `test_constrained_arcs_2tan_center_on`) to verify these fixes.

*   **Additional info**:
    * Problem: The OCCT class `Geom2dGcc_Circ2d3Tan` (used to find a circle tangent to three 2D curves) fails when any input is a spline. 

    * Root Cause: In the OCCT source, the iterative solver's core function object `Geom2dGcc_FunctionTanCuCuCu` has its `Values()` and `Derivatives()` methods completely disabled via `#if 0` blocks. While the analytical solver for lines/circles works, any "complex" curve triggers the iterative path, which currently operates on uninitialized memory for the Jacobian and function residuals, leading to divergence or crashes.

```python
from build123d import *  # type: ignore
from ocp_vscode import colors, set_colormap, show_all

set_colormap(colors.ColorMap().tab20())

# test 3tan
c1 = CenterArc((0, 20), 20, -90, 90)
c2 = CenterArc((0, 20), 10, -90, 90)
ln1 = Line(c1 @ 1, c2 @ 1)

# arc1 and arc2 failed in `constrained_lines.py:533 out_topos.append(arcs[sagitta.value])`
# with `IndexError: list index out of range`
arc1 = ConstrainedArcs(c1, c2, ln1)
arc1_alt = ConstrainedArcs(c1, c2, ln1, sagitta=Sagitta.SHORT)

# test 2tan on
ln2 = Line((10, 10), ln1 @ 1)

# arc3 returned a wrongly trimmed arc.
arc3 = ConstrainedArcs(tangency_one=c1, tangency_two=ln1, center_on=ln2)


show_all()
```
image

gumyr (changes requested):
I'm unsure about this PR. The change to `sagitta` is changing how the API works. The original intention was that if there were multiple valid solutions the `sagitta` parameter would select between them. If I understand the PR correctly, `sagitta` is now part of the solution finding process. If the pi adjustment was applied earlier, wouldn't the `sagitta` selection be fine as is?

javimixet:
> I'm unsure about this PR. The change to `sagitta` is changing how the API works. The original intention was that if there were multiple valid solutions the `sagitta` parameter would select between them. If I understand the PR correctly, `sagitta` is now part of the solution finding process. If the pi adjustment was applied earlier, wouldn't the `sagitta` selection be fine as is?

The pi adjustment and the automatic arc selection address different issues.

The pi adjustment ensures that the tangency points are computed correctly on the solution circle. However, once those points are known, there are still two possible arcs between the first two tangency points.

_make_3tan_arcs is a special case because it has a third tangency point. That third point can be used to determine which of the two candidate arcs is the geometrically correct one.

The purpose of changing the default to sagitta=None is not to make sagitta part of the solution-finding process. Circle solutions are still found and filtered exactly as before. Rather, it allows _make_3tan_arcs to use its additional geometric information to automatically select the correct arc instead of defaulting to Sagitta.SHORT.

If the user explicitly provides Sagitta.SHORT, Sagitta.LONG, or Sagitta.BOTH, the existing sagitta behavior is preserved and the automatic selection is bypassed.

The change is also limited to _make_3tan_arcs. For the other ConstrainedArcs constructors, the current behavior is preserved by internally mapping:
```py
if sagitta is None:
    sagitta = Sagitta.SHORT
```
so their default behavior remains unchanged.

Therefore, the only behavioral change is the default behavior of _make_3tan_arcs, where the third tangency point makes automatic selection possible preventing the default Sagitta.SHORT selection from discarding the geometrically correct arc.

That said, this sagitta-related change is not required to fix #1294. The π adjustment alone solves that issue. The sagitta=None default is a way to automate the arc selection independently of sagitta, but it can be removed if changing the API in this way is not desired.

javimixet:
Consider this example, used for [`test_make_constrained_arcs_3tan` ](https://github.com/gumyr/build123d/pull/1302/changes#diff-b0b0d3217235aaf2038ba755da0242b5942428323c600ec12a592060db01131bR529):
```python
c1 = CenterArc((0, 20), 20, -90, 90)
c2 = CenterArc((0, 20), 10, -90, 90)
l1 = Line(c1 @ 1, c2 @ 1)
c3 = RadiusArc(c1 @ 1, c2 @ 1, 10)
tan3_l1_s = Edge.make_constrained_arcs(c1, c2, l1, sagitta=Sagitta.SHORT)
tan3_c3_s = Edge.make_constrained_arcs(c1, c2, c3, sagitta=Sagitta.SHORT)

tan3_l1_b = Edge.make_constrained_arcs(c1, c2, l1, sagitta=Sagitta.BOTH)
print([e.length for e in tan3_l1_b.edges()])  # [15.707963267948966, 15.707963267948964]

tan3_c3_b = Edge.make_constrained_arcs(c1, c2, c3, sagitta=Sagitta.BOTH)
print([e.length for e in tan3_c3_b.edges()])  # [15.707963267948966, 15.707963267948966]
```

image

Here the two resulting arcs (two for tan3_l1_b and two for tan3_c3_b) should be geometrically identical in length, `Sagitta.SHORT` resorts to numerical error to select the returned arc, leading to random results (in tan3_l1_s the bottom arc is selected while in tan3_c3_s it's the top one). In  [`test_make_constrained_arcs_3tan` ](https://github.com/gumyr/build123d/pull/1302/changes#diff-b0b0d3217235aaf2038ba755da0242b5942428323c600ec12a592060db01131bR529) sagitta is not needed to return the correct result in both tan3_l1_s and tan3_c3_s.

Maybe we don't need sagitta at all for `_make_3tan_arcs` (?).

gumyr:
`sagitta` for the 3 tangent case is still valid. It's possible that the user wants the part of the arc that doesn't touch the third object. `sagitta` allows the user more control and should be consistent across the whole API. Is it possible to have the pi fix without the changes to the how `sagitta` works?

gumyr:
Thank you! This was a tricky problem, I really appreciate the effort you've put into fixing it.

PR #1308, opened 2026-05-08
Avoid duplicate boolean clean passes

sammekekko:
Avoid running a second clean pass after boolean operations that already clean their result.

The shared boolean helper already applies ShapeUpgrade_UnifySameDomain when SkipClean.clean is enabled. This removes the caller-level clean from direct Shape and Compound additions. Builder context integration now skips the caller-level clean for add, subtract, and intersect paths that call the shared boolean helper, while retaining cleanup for non-boolean integrations that do not go through that helper.

To keep the scope tight, this does not change logging behavior and does not enable OCCT glue.

* Remove redundant clean work in direct Shape and Compound additions.
* Skip duplicate clean work when builder context integration uses add, subtract, or intersect.
* Keep builder cleanup for non-boolean context integration paths.
* Add regression tests for empty-object batch additions with overlapping boxes, mixed curved solids, and rotated solids.

Validation:

* `uv run --extra development pytest` (`1988 passed, 2 skipped`)
* `uv run --extra development pytest tests/test_benchmarks.py::test_ppp_0106 tests/test_algebra.py tests/test_build_common.py tests/test_build_part.py tests/test_direct_api/test_compound.py tests/test_direct_api/test_shape.py` (`292 passed`)
* `uv run --extra development pytest tests/test_benchmarks.py::test_ppp_0106 tests/test_examples.py -k ppp0106` (`1 passed`)
* `uv run --extra development pylint --rcfile=.pylintrc --fail-under=9.5 src/build123d`

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -451,6 +451,7 @@ class Builder(ABC, Generic[ShapeT]):
                     mode,
                 )
                 combined: Shape | list[Shape] | None
+                needs_clean = clean
                 if mode == Mode.ADD:
                     if self._obj is None:
                         if len(typed[self._shape]) == 1:
@@ -459,16 +460,20 @@ class Builder(ABC, Generic[ShapeT]):
                             combined = (
                                 typed[self._shape].pop().fuse(*typed[self._shape])
                             )
+                            needs_clean = False
                     else:
                         combined = self._obj.fuse(*typed[self._shape])
+                        needs_clean = False
                 elif mode == Mode.SUBTRACT:
                     if self._obj is None:
                         raise RuntimeError("Nothing to subtract from")
                     combined = self._obj.cut(*typed[self._shape])
+                    needs_clean = False
                 elif mode == Mode.INTERSECT:
                     if self._obj is None:
                         raise RuntimeError("Nothing to intersect with")
                     combined = self._obj.intersect(Compound(typed[self._shape]))
+                    needs_clean = False
                 elif mode == Mode.REPLACE:
                     combined = self._sub_class(list(typed[self._shape]))
 
@@ -487,7 +492,7 @@ class Builder(ABC, Generic[ShapeT]):
                 #     else combined
                 # )
 
-                if self._obj is not None and clean:
+                if self._obj is not None and needs_clean:
                     self._obj = self._obj.clean()
 
                 logger.info(
--- a/src/build123d/topology/composite.py
+++ b/src/build123d/topology/composite.py
@@ -103,7 +103,6 @@ from .one_d import Edge, Wire, Mixin1D
 from .shape_core import (
     Shape,
     ShapeList,
-    SkipClean,
     Joint,
     downcast,
     shapetype,
@@ -491,7 +490,6 @@ class Compound(Mixin3D[TopoDS_Compound]):
             s for s in self.get_top_level_shapes() + summands if s is not None
         )
 
-        # Only fuse the parts if necessary
         if len(summands) <= 1:
             result: Shape = Compound(summands[0:1])
         else:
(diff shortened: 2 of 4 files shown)


gumyr:
The logging changes - `if logger.isEnabledFor(logging.INFO):` - might make processing a little faster but outside of the very special case of fusing many, many boxes (6 planar sided objects) the extra checks would slow processing down. What is the application that requires merging large numbers of boxes? Couldn't this be done more efficiently in the 2D domain?

sammekekko:
Fair point. I’ll trim this PR down so it’s focused on the batch fuse behavior and remove the logging/context micro-optimizations.

The case I was looking at was generated geometry where a lot of simple box-like solids get combined, for example grid/fixture/block-style models. But I agree that when the model is really a set of 2D footprints with a common height, doing the union in 2D and extruding afterward is probably the better approach.

I’ll update the PR so the scope is clearer and less speculative.

sammekekko:
I tightened the PR further and force-pushed the reduced version.

This no longer changes logging, location context handling, edge parallelization, or OCCT glue. I also dropped the box-specific optimized path after checking it more closely, since glue is not the right general answer for overlapping boxes.

The PR now only avoids a duplicate clean pass after boolean operations that already clean inside the shared boolean helper. The existing boolean path is unchanged, and I added regression tests for overlapping box batches, mixed curved additions, and rotated additions. Full test suite is green, and the updated PR description has the reduced benchmark numbers.

src/build123d/topology/composite.py
```diff
@@ -492,15 +492,17 @@ def __add__(self, other: None | Shape | Iterable[Shape]) -> Compound | Wire:
...
             result: Shape = Compound(summands[0:1])
         else:
             fuse_op = BRepAlgoAPI_Fuse()
             fuse_op.SetFuzzyValue(TOLERANCE)
             self.copy_attributes_to(summands[0], ["wrapped", "_NodeMixin__children"])
             result = self._bool_op(summands[:1], summands[1:], fuse_op)
+            clean_already_done = SkipClean.clean
 
-        if SkipClean.clean:
+        if SkipClean.clean and not clean_already_done:
```
gumyr: It seems as though this clean operation should be removed entirely. Either a single object is in the Compound (should that be cleaned?) or the fuse will have already cleaned the result.

src/build123d/topology/shape_core.py
```diff
@@ -980,15 +980,22 @@ def __add__(self, other: None | Shape | Iterable[Shape]) -> Self | Compound:
...
                 sum_shape = summands[0]
             else:
                 sum_shape = summands[0].fuse(*summands[1:])
+                clean_already_done = SkipClean.clean
         else:
             sum_shape = self.fuse(*summands)
+            clean_already_done = SkipClean.clean
 
-        if SkipClean.clean and not isinstance(sum_shape, list):
+        if (
```
gumyr: Same comment, it seems that this clean is redundant. Either a single object is present or the results were already cleaned.

src/build123d/build_common.py
```diff
@@ -487,7 +493,7 @@ def _add_to_context(
                 #     else combined
                 # )
 
-                if self._obj is not None and clean:
+                if self._obj is not None and clean and not clean_already_done:
```
gumyr: Same comment here, I think clean is redundant here. I'm guessing clean was added to _bool_op after this code was written and the duplicate clean never identified/removed..

gumyr (changes requested):
Thanks for identifying this, what appears to be, redundant call to clean.

sammekekko:
I pushed an update for these review comments.

The direct `Shape` and `Compound` addition paths no longer do a caller-level clean after the boolean operation. For builder context integration, add, subtract, and intersect now also skip the caller-level clean after `_bool_op()` has already handled cleanup.

I kept cleanup for the non-boolean builder integration paths because removing it entirely broke `ttt-ppp0106.py`: `make_face()` adds a single face without going through `_bool_op()`, and the following fillet depends on that cleanup. So the builder path now only cleans when the integration did not run through the shared boolean helper.

Validation is updated in the PR body. Full test suite and pylint both pass locally.

gumyr:
Thank you!

PR #1312, opened 2026-05-14
Change Matrix.rotate() to use degrees

### LINKED ISSUE #1310: Matrix.rotate() should be changed to use degrees
Matrix.rotate incorrectly uses radians and should be changed to use degrees like every other build123d API.

_Originally posted by @gumyr in https://github.com/gumyr/build123d/pull/1262#discussion_r3235116729_

DaWeidmann:
Fixes #1310


gumyr:
Thank you!

PR #1313, opened 2026-05-14
Allow BoundBox._init_() to take a Shape

### LINKED ISSUE #1293: Add Shape to BoundBox class constructor
@DaWeidmann Now that the constructor is in place and handles the simple case of a `TopoDS_Shape`, would you be willing to expand the constructor in a small way to take a `Shape`?  The reason behind all of this is that it's a bit weird that we have to create bounding boxes with `my_object.bounding_box()` while the standard Python method (that's used almost universally in build123d) is to have the constructor do this like `list(...)`, `tuple(...)` or `Shell(...)`. We should be able to do `bbox = BoundBox(my_object)`. If you don't have time I'll create a new issue for it.

_Originally posted by @gumyr in https://github.com/gumyr/build123d/issues/1263#issuecomment-4183715281_

DaWeidmann:
Fixes #1293

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -1034,7 +1034,7 @@ class BoundBox:
         tolerance: float | None = None,
         optimal: bool = True,
     ) -> None:
-        """Construct a bounding box from a TopoDS_Shape"""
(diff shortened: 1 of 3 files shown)


gumyr:
Thank you!

PR #1321, opened 2026-05-28
Type Hint Improvements to + and - Operators

DavidMStraub:
Hi!

I use build123d in teaching and make heavy use of type hints/checks.

One thing I noticed is that boolean operations with algebraic operators often lead to a surprisingly wide type, which requires manual type narrowing.

This PR adds some more specific type hints via overloads that make this more ergonomic.

AI generated summary:

---

## What Changed

Added `@overload` signatures to `Shape.__add__`, `Shape.__sub__`, and `Mixin1D.__add__`
so that static type checkers (pyright, mypy) can narrow the return type based on the
argument passed.

## The Problem Before

```python
box: Solid = Box(1, 1, 1)
cyl: Solid = Cylinder(0.5, 2)

result = box + cyl
# pyright/mypy: Shape | Compound  ← too wide, loses Solid identity
```

Every `+` or `-` call returned the base `Shape | Compound`, forcing callers to add manual
casts or live with false "possibly None" / "no attribute X" errors downstream.

The same applied to 1D shapes:

```python
w1: Wire = Wire.make_circle(1)
w2: Wire = Wire.make_circle(2)

result = w1 + w2
# before: Shape | Compound  ← should be Edge | Wire | Curve
result.edges()  # ← checker: "edges" may not exist on Shape
```

### 1. Accurate `reveal_type` / hover types in your IDE

Verified with **pyright** (0 errors, 0 warnings) and **mypy**:

| Expression | Before | After |
|---|---|---|
| `Solid(...) + Solid(...)` | `Shape \| Compound` | `Solid \| Compound` ✅ |
| `Solid(...) - Solid(...)` | `Shape \| Compound` | `Solid \| Compound` ✅ |
| `Wire(...) + Wire(...)` | `Shape \| Compound` | `Edge \| Wire \| Curve` ✅ |
| `solid + None` | `Shape \| Compound` | `Solid` ✅ |
| `wire + None` | `Shape \| Compound` | `Wire` ✅ |

### 2. No spurious "attribute not found" errors

```python
result = box - cyl          # result: Solid | Compound
result.shell_thickness(…)   # ✅ valid on both Solid and Compound — no error
```

### 3. The `+ None` / `- None` identity case is typed correctly

Passing `None` returns the object unchanged. The overloads express this as `-> Self`,
so the return type is the same concrete subclass as the receiver:

```python
def maybe_add(s: Solid, extra: Solid | None) -> Solid:
    return s + extra   # ✅ None → Solid; Solid → Solid | Compound
```

Previously, both cases returned `Shape | Compound`, requiring an explicit `cast`.

### 4. Zero runtime impact

Only the static signatures changed. All original runtime logic, including the
`fuse()` / `cut()` fallback paths, is completely untouched.

## Files Changed

| File | Change |
|---|---|
| `src/build123d/topology/shape_core.py` | `@overload` on `Shape.__add__` and `Shape.__sub__` |
| `src/build123d/topology/one_d.py` | `@overload` on `Mixin1D.__add__` (`→ Edge \| Wire \| Curve`); removed LSP-violating `Wire.__add__` / `Edge.__add__` subclass overrides |

### DIFF
--- a/src/build123d/topology/one_d.py
+++ b/src/build123d/topology/one_d.py
@@ -195,7 +195,7 @@ from OCP.TopTools import (
 )
 from scipy.optimize import minimize_scalar
 from scipy.spatial import ConvexHull
-from typing_extensions import deprecated
+from typing_extensions import Self, deprecated
 
 from build123d.build_enums import (
     AngularDirection,
@@ -609,7 +609,11 @@ class Mixin1D(Shape[TOPODS]):
 
     # ---- Instance Methods ----
 
-    def __add__(self, other: None | Shape | Iterable[Shape]) -> Edge | Wire | Shape:
+    @overload
+    def __add__(self, other: None) -> Self: ...
+    @overload
+    def __add__(self, other: Shape | Iterable[Shape]) -> Edge | Wire | Curve: ...
+    def __add__(self, other):
         """fuse shape to wire/edge operator +"""
 
         # Convert `other` to list of base topods objects and filter out None values
--- a/src/build123d/topology/shape_core.py
+++ b/src/build123d/topology/shape_core.py
@@ -958,7 +958,11 @@ class Shape(NodeMixin, Generic[TOPODS]):
             raise RuntimeError("Composite factory is not registered")
         return factory(shape_list)
 
-    def __add__(self, other: None | Shape | Iterable[Shape]) -> Self | Compound:
+    @overload
+    def __add__(self, other: None) -> Self: ...
+    @overload
+    def __add__(self, other: Shape | Iterable[Shape]) -> Self | Compound: ...
+    def __add__(self, other):
         """fuse shape to self operator +"""
         # Convert `other` to list of base objects and filter out None values
         if other is None:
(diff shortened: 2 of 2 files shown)


gumyr:
In general improvements in typing are welcome; however, I need to look at this closely as the "Problem" statement is incorrect, here is the class hierarchy for Box:
```py
>>> Box.mro()
[<class 'build123d.objects_part.Box'>, <class 'build123d.objects_part.BasePartObject'>, <class 'build123d.topology.composite.Part'>, <class 'build123d.topology.composite.Compound'>, <class 'build123d.topology.three_d.Mixin3D'>, <class 'build123d.topology.shape_core.Shape'>, <class 'anytree.node.nodemixin.NodeMixin'>, <class 'typing.Generic'>, <class 'object'>]
```
... it's not simply a `Solid`. Other than when combined with `None` I don't think `Self` is appropriate either, a `Box` combined with anything is not going to be class `Box` but some variant of `Part`/`Compound`. 

Note that typing in build123d isn't all that easy as there are many operations that could change the type of an object. A `Solid` could be broken into multiple `Solids` with combinations of other `Solid(s)` so it's impossible for a static typing system to know exactly what the tightest type should be so the more general `Part` / `Compound` types are used instead.

DavidMStraub:
You're right - the AI-generated description is wrong, sorry about that.

The improvement only concerns users of the algebraic/functional API who directly work with `Solid`s or `Wire`s, not with `Compound` subclasses like `Box`.

This should be the correct mapping table:


| Expression | Before | After |
|---|---|---|
| `s: Solid; s + None` | `Solid \| Compound` | `Solid` ✅ |
| `s: Solid; s - None` | `Solid \| Compound` | `Solid` ✅ |
| `w: Wire; w + None` | `Edge \| Wire \| Shape` | `Wire` ✅ |
| `w: Wire; w + w` | `Edge \| Wire \| Shape` | `Edge \| Wire \| Curve` ✅ |

gumyr:
Thank you!

PR #1322, opened 2026-06-01
feat: allow Axis construction with end_point

### LINKED ISSUE #1214: Consider adding new kwarg for Axis: `Axis((0,0,0), end_point=(1,2,3))`
`Axis((0,0,0), end_point=(1,2,3))` the idea is that one could specify a origin and a kwarg-only end_point instead and Axis would internally compute the direction. Currently possible to work around this via `Axis(Line(*points))`.

```py
class Axis
...
    @overload
    def __init__(self, origin: VectorLike, *, end_point: VectorLike): # extra asterisk enforces kwarg-only for end_point
        """Axis: point and end point"""
```

pranjalbhatia710:
## Summary
- Add `Axis(origin, end_point=...)` construction for point-to-point axis definitions
- Keep `end_point` keyword-only so existing `Axis(origin, direction)` behavior is unchanged
- Add tests for positional/keyword origin usage, normalized direction, conflicting direction/end_point input, and zero-length point pairs

Closes #1214

## Tests
- `uv run python -m pytest tests/test_direct_api/test_axis.py -q`
- `uv run black --check src/build123d/geometry.py tests/test_direct_api/test_axis.py`
- `git diff --check`

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -626,11 +626,12 @@ class AxisMeta(type):
 class Axis(metaclass=AxisMeta):
     """Axis
 
-    Axis defined by point and direction
+    Axis defined by point and direction or by two points
 
     Args:
         origin (VectorLike): start point
         direction (VectorLike): direction
+        end_point (VectorLike): point used with origin to define direction
         edge (Edge): origin & direction defined by start of edge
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #1326, opened 2026-06-02
fix(drafting): DimensionLine crash fix + ExtensionLine side parameter

pzfreo:
## Summary

Two improvements to the existing dimension line classes in `drafting.py`:

**1. `DimensionLine` — crash fix**

`DimensionLine` raised `ValueError("Can't get geom adaptor of empty wire")` when `pad_around_text` made `shaft_length` go negative even though the label nominally fit within the path. This happened with single-arrow dimensions at borderline path lengths (a real 1-unit window with default `Draft` values).

Fix: compute `shaft_length = max(0.0, ...)` unconditionally and add `and shaft_length > 0` to the existing condition, routing the edge case to the external-arrow branch instead of crashing.

**2. `ExtensionLine` — `side` parameter**

Adds an optional `side` parameter accepting `"above"`, `"below"`, `"left"`, `"right"`, or an explicit `VectorLike`. When provided, the sign of `offset` is computed automatically from the path orientation so callers don't need to manually determine the correct sign.

```python
# Before — caller must know the sign convention
ExtensionLine(border=edge, offset=6, draft=d)   # positive = above for this edge?

# After
ExtensionLine(border=edge, offset=6, draft=d, side="above")
```

The sign is derived from the midpoint tangent (better than start tangent for arcs and multi-segment wires). Input validation covers: unknown string keys, zero/invalid VectorLike, edges with no XY component, and ambiguous (parallel) directions — all with clear `ValueError` messages. Fully backward-compatible; existing code using raw `offset` is unchanged.

## Test plan

- [x] `test_side_above_places_dimension_above` — horizontal edge, bbox above
- [x] `test_side_below_places_dimension_below` — horizontal edge, bbox below
- [x] `test_side_above_and_below_are_mirrored` — symmetry check
- [x] `test_side_right_places_dimension_right_of_vertical_edge`
- [x] `test_side_left_places_dimension_left_of_vertical_edge`
- [x] `test_side_as_vectorlike` — VectorLike input matches string form, placement asserted
- [x] `test_side_invalid_string_raises` — bad string → `ValueError` with key in message
- [x] `test_side_zero_vectorlike_raises` — zero vector → `ValueError`
- [x] `test_side_parallel_to_edge_raises` — ambiguous direction → `ValueError`
- [x] `test_side_z_axis_edge_raises` — Z-axis border → `ValueError`
- [x] `test_label_wider_than_path_does_not_crash` — crash fix regression test
- All 40 drafting tests pass

### DIFF
--- a/src/build123d/drafting.py
+++ b/src/build123d/drafting.py
@@ -152,6 +152,14 @@ PathDescriptor: TypeAlias = Wire | Edge | list[PointLike]
 """General type for a path in 3D space"""
 
 
+_SIDE_VECTORS = {
+    "above": Vector(0, 1, 0),
+    "below": Vector(0, -1, 0),
+    "right": Vector(1, 0, 0),
+    "left": Vector(-1, 0, 0),
+}
+
+
 @dataclass
 class Draft:
     """Draft
@@ -400,8 +408,8 @@ class DimensionLine(BaseSketchObject):
         # Calculate the arrow shaft length for up to three types
         if arrows.count(True) == 0:
             raise ValueError("No output - no arrows selected")
-        if label_length + arrows.count(True) * draft.arrow_length < path_length:
-            shaft_length = (path_length - label_length) / 2 - draft.pad_around_text
+        shaft_length = max(0.0, (path_length - label_length) / 2 - draft.pad_around_text)
+        if label_length + arrows.count(True) * draft.arrow_length < path_length and shaft_length > 0:
             shaft_pair = [
                 path_obj.trim(0.0, shaft_length / path_length),
                 path_obj.trim(1.0 - shaft_length / path_length, 1.0),
@@ -489,7 +497,8 @@ class ExtensionLine(BaseSketchObject):
         border (PathDescriptor): a very general type of input defining the object to
             be dimensioned. Typically this value would be extracted from the part but is
             not restricted to this use.
-        offset (float): a distance to displace the dimension line from the edge of the object
+        offset (float): a distance to displace the dimension line from the edge of the object.
+            Ignored when ``side`` is provided.
         draft (Draft): instance of Draft dataclass
         label (str, optional): a text string which will replace the length (or arc length)
             that would otherwise be extracted from the provided path. Providing a label is
@@ -507,6 +516,11 @@ class ExtensionLine(BaseSketchObject):
         measurement_direction (VectorLike, optional): Vector line which to project the dimension
             against. Offset start point is the position of the start of border.
             Defaults to None.
+        side (str | VectorLike, optional): world-space direction for the dimension line —
+            ``"above"``, ``"below"``, ``"left"``, ``"right"``, or an explicit direction
+            vector. When provided, the sign of ``offset`` is computed automatically so
+            callers don't need to know the path orientation. For curved or multi-segment
+            borders the direction is inferred from the midpoint tangent. Defaults to None.
         mode (Mode, optional): combination mode. Defaults to Mode.ADD.
 
     """
@@ -522,6 +536,7 @@ class ExtensionLine(BaseSketchObject):
         tolerance: float | tuple[float, float] | None = None,
         label_angle: bool = False,
         measurement_direction: VectorLike | None = None,
+        side: str | VectorLike | None = None,
         mode: Mode = Mode.ADD,
     ):
         # pylint: disable=too-many-locals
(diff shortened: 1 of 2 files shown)


slobberingant:
This argument is duplicating the function of measurement_line with pre-defined sides rather than vectors.
I don't think ExtensionLine should have both of these arguments.
I would either remove  measurement_line in favor of sides (limit ability to project dimensions only to 4 sides) or add enums to measurement_line for 'above', 'below' while retaining vector processing ability etc.

gumyr:
There are a couple problematic things with this proposal:

1. `side` is a literal unlike the rest of the API, enums are expected for this type of thing 
2. more importantly, `side` is confusing or contradictory when used with `measurement_direction` as pointed out by @slobberingant 

I've explored quite a few alternatives and the best I can come up with is to allow `offset` to be either a `float` (preserving today's behaviour) or a `Vector` which allows the user to directly control the placement without the uncertainty of the current `offset` `Edge` direction behaviour. If the user provides both a `Vector` `offset` and `measurement_direction` they must be perpendicular.

Here are the options I considered for inspiration:

  1. LEFT / RIGHT / TOP / BOTTOM
      - Simple for horizontal/vertical drawings.
      - Too limited for arbitrary measurement_direction.
      - Fails or becomes unclear for diagonal/rotated dimensions.
  2. Use sign of measurement_direction
      - Formal rule could be: reverse measurement_direction, reverse placement side.
      - Too implicit and surprising.
      - Users expect measurement_direction to control what is measured, not where the dimension is placed.
  3. placement_direction
      - General vector-based control.
      - Problem: non-orthogonal inputs like (1, 0.1) are ambiguous because only the component perpendicular to the dimension matters.
      - Requires either projection semantics or validation.
  4. placement_point
      - Similar to interactive CAD “click where the dimension goes.”
      - Rejected because users would expect the dimension line to pass through that point, not merely use it to infer a side.
  5. placement_side = POSITIVE / NEGATIVE
      - General if defined relative to a computed placement axis.
      - But if measurement_direction is optional, the reference axis may be unavailable.
      - Also similar to the current signed offset, so it does not solve enough.
  6. placement_axis
      - Explicit perpendicular axis for placement.
      - More precise than placement_direction.
      - Requires user to supply an axis perpendicular to the dimension/measurement direction.
      - Still adds a new parameter and validation complexity.
  7. offset_direction
      - Defines the direction in which scalar offset is applied.
      - Clearer than placement_direction.
      - Works even when measurement_direction is absent.
      - Adds a new parameter.
  8. Overload offset: float | VectorLike
      - float: current behavior, backward compatible.
      - VectorLike: explicit displacement, providing both direction and magnitude.
      - No extra enum or parameter.
      - Avoids ambiguity because the vector is the actual offset, not a side hint.
      - Most general and least invasive option. Current favorite.

Any other thoughts?

pzfreo:
Thanks both — this is helpful, and I agree the `side` literal was the wrong shape for this API. The `measurement_direction` conflict @slobberingant raised is real, and overloading `offset: float | VectorLike` (option 8) is cleaner than anything I'd considered. Happy to rework the PR around it.

To make sure I implement what you have in mind:

1. **`float` offset** — unchanged from today (signed scalar along the existing derived direction).
2. **`VectorLike` offset** — taken as the literal displacement of the dimension line: direction *and* magnitude come straight from the vector, so there's no sign-convention guesswork.
3. **With `measurement_direction`** — if both are supplied, validate that the vector offset is perpendicular to it and raise a `ValueError` otherwise (rather than silently projecting). When `measurement_direction` is absent, the vector alone fully determines placement.

One small thing to confirm: for the perpendicularity check, do you want an exact test or a tolerance (e.g. angle within some epsilon to absorb floating-point error from computed directions)? I'd lean toward a small tolerance.

Separately — this PR also carries an unrelated `DimensionLine` crash fix (negative `shaft_length` when `pad_around_text` exceeds a borderline path, routing to the external-arrow branch instead of raising "Can't get geom adaptor of empty wire"). That part's independent of the placement discussion — want me to keep it here, or split it into its own PR so it can land while we settle the `offset` design?

slobberingant:
Overloading the offset seems like a neat solution.
So these two approaches are identical. Is this what we want?

```
l1 = Line((0,0), (10,10)
ExtensionLine(l1, offset=5, measurement_direction=(1, 0, 0))
ExtensionLine(l1, offset=5 * (1, 0, 0))
```

gumyr:
@pzfreo It would be easier to manage if this PR was split into two: 1) the placement of the extension line, and 2) error handling when text doesn't fit.  The perpendicular check should have some tolerance to it as were really only need a rough direction. Thanks for taking this on.

@slobberingant the `measurement_direction` is the direction of the dimension line (i.e. aligned with the arrows) while the `offset` is perpendicular to that indicating which side of the object the dimension should be placed. Here is an example with the `measurement_direction` of `(0,1)` (the vertical one) and `(1,0)` (the horizontal one). image


So the example would look more like:
```py
l1 = Line((0,0), (10,10)
ExtensionLine(l1, offset=5, measurement_direction=(1, 0))
ExtensionLine(l1, offset=(0, 5))
```
or
```py
l1 = Line((0,0), (10,10)
ExtensionLine(l1, offset=5, measurement_direction=(1, 0))
ExtensionLine(l1, offset=(0, -5))
```
In the `offset=(5,0)` case an error should be generated as one can't place a horizontal dimension line to the left of `l1` (the perpendicular check would catch this). 

By the way, Python controls this and it isn't really what we want:
```py
>>> 5 * (1,0,0)
(1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0)
```

pzfreo:
Splitting this into two focused PRs as suggested:

- **#1336** — `DimensionLine` crash fix (the uncontroversial half).
- **#1338** — `ExtensionLine` placement, reworked to the `offset: float | VectorLike` design discussed above (drops the `side=` literal; perpendicularity validated, inferred when `measurement_direction` is absent).

Closing this in favour of those two. Thanks for the review and the design steer.

PR #1327, opened 2026-06-03
Add implicit fields: marching cubes + signed-distance fields (build123d.implicit)

GibbinsJosh:
## Summary

Adds `build123d.implicit` — a bidirectional bridge between implicit/scalar fields and B-rep solids:

```
implicit field  --implicit_solid()-->  Solid          (VTK marching cubes)
Solid/STEP/STL  --mesh_to_sdf()---->    signed-distance field
```

Both directions share the signed-distance convention (`f < 0` inside), so a `mesh_to_sdf()` field plugs straight back into `implicit_solid()` — combine a B-rep part with analytic fields, then re-mesh. This enables modelling that is awkward or impossible with B-rep booleans alone: TPMS infill, lattice structures, smooth (metaball) blends, functionally graded structures, and robust uniform shelling.

## Public API (exported from `build123d`)

| Function | Purpose |
|---|---|
| `implicit_solid(func, bounds, resolution, *, level=0, close_boundary=True)` | field → watertight `Solid` |
| `implicit_mesh(...)` | field → raw `(vertices, faces)` |
| `implicit_stl(..., file_path)` | field → binary STL (fast path, no OCC sewing) |
| `mesh_to_sdf(source, *, tolerance=…)` | `Shape`/STEP/STL → `SignedDistanceField` callable |
| `SignedDistanceField` | the callable returned by `mesh_to_sdf` (carries `.bounds`) |

## How it works

- **Marching cubes** via VTK (already a build123d dependency — see `vtk_tools.py`). The sampling grid is padded with an "outside" border so any surface reaching the bounds is capped → guaranteed watertight output.
- **`mesh_to_sdf`** reconstructs the field from a triangulation with the exact point-to-triangle distance (Ericson closest-feature) for magnitude and the generalized winding number (Jacobson et al.) for sign — orientation-independent and robust for watertight tessellations. Pure NumPy, no new dependencies.

## Examples (`examples/`)

Each targets something B-rep handles poorly:

- **smooth_blend.py** — smooth-minimum (organic fillets / metaballs); renders a hard union beside a smooth union of the same spheres.
- **graded_lattice.py** — functionally graded lattice whose strut radius varies with height; one expression, no per-strut modelling.
- **conformal_shell.py** — uniform `abs(f) - t` shell of a curved "peanut" with a gyroid infill that follows the wall, shown as a cutaway.
- **gyroid_infill.py** — gyroid TPMS infill clipped to a part via `mesh_to_sdf`.
- **lattice_infill.py** — periodic strut lattice clipped to a sphere.

## Tests

`tests/test_implicit.py` — 17 cases covering both directions: sphere volume accuracy, valid `Solid`, level offset, `close_boundary` → watertight, empty-field `ValueError`; STL written and re-readable; `SignedDistanceField` type/shape/bounds, inside-outside sign, exact face distances, `Solid → field → Solid` round-trip, metric offset (erode/dilate), STEP + STL file inputs, unsupported-source `ValueError`. The five examples are exercised by the existing `tests/test_examples.py` harness.

## Verification (run natively against the dev source)

- `pylint --rcfile=.pylintrc src/build123d/implicit.py` → **10.00/10**
- `mypy --config-file mypy.ini src/build123d/implicit.py` → **clean**
- `black --check` → clean
- `pytest tests/test_implicit.py` → **17 passed**; all five examples pass `tests/test_examples.py`

## Known limitation

`mesh_to_sdf` evaluation is `O(n_points × n_triangles)` — cheap for low-poly inputs or sparse queries, but meshing the field of a densely tessellated part over a fine grid is slow. This is documented in the docstring. A spatial-acceleration follow-up (KD-tree distance + angle-weighted-pseudonormal sign) would make it `O(n log m)`; happy to do it in a follow-up if there's interest.

## Notes

- No new runtime dependencies (VTK and NumPy are already required).
- New file `src/build123d/implicit.py`; `__init__.py` gains one import and five `__all__` entries.

### DIFF
--- a/examples/conformal_shell.py
+++ b/examples/conformal_shell.py
@@ -1,0 +1,94 @@
+"""
+Robust uniform shelling + conformal infill, shown as a cutaway
+
+name: conformal_shell.py
+by:   Joshua Gibbins (GibbinsJosh)
+date: June 3, 2026
+
+desc:
+    Why implicits? Two things the B-rep kernel struggles with become trivial:
+
+      1. SHELLING / OFFSET. Hollowing a curved solid to a uniform wall is a
+         classic place B-rep `offset`/`shell` self-intersects -- especially in
+         concave regions like the neck of this "peanut". As a field, a uniform
+         shell of ANY geometry is just `abs(f) - thickness`, and it cannot fail.
+      2. CONFORMAL INFILL. The cavity is filled with a gyroid that follows the
+         curved wall exactly, because the infill is clipped by the same field.
+
+    The peanut here is the union of two overlapping spheres, built directly as a
+    signed-distance field. The identical shell/infill algebra works on a B-rep
+    part too -- just replace `peanut` with `mesh_to_sdf(my_part)` (see
+    gyroid_infill.py) to hollow and lattice an imported STEP/STL.
+
+    The model is cut in half so the uniform wall and the infill are visible.
+
+    Field algebra (signed-distance convention: f < 0 is solid):
+        C     = peanut field
+        shell = max(C, -t - C)            # solid wall band, thickness t
+        infill= max(|gyroid| - w, C + t)  # gyroid in the cavity (deeper than t)
+        solid = min(shell, infill)        # wall + infill
+        cut   = max(solid, y)             # keep y <= 0 to reveal the section
+
+license: Apache-2.0 (see build123d/src/build123d/implicit.py)
+"""
+
+# [Code]
+import numpy as np
+
+from build123d import implicit_stl, import_stl
+
+try:
+    from ocp_vscode import Render, show
+except ImportError:  # viewer is optional
+    Render = show = None
+
+# ---- Parameters ------------------------------------------------------------
+wall = 1.6  # shell wall thickness (mm)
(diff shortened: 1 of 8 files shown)


gumyr:
Thank you for the PR and for exploring this direction. Implicit/SDF modeling is interesting, but I don’t think this is a good fit for build123d’s core architecture.

build123d is fundamentally built around B-rep geometry and OpenCascade topology. A conversion from B-rep to SDF is effectively a lossy representation change: even if an SDF-derived object can later be wrapped as a shape, the original analytic surfaces, edges, vertices, and topology are no longer preserved in the way users expect from build123d/OCCT objects.

There also isn’t a common SDF exchange format comparable to STEP for B-reps, so even a narrower `export_sdf` feature would have limited general interoperability. In addition, this PR introduces/depends on VTK, and build123d is moving away from VTK as a dependency.

For these reasons, I’m going to close this PR. I think this work would be better suited as a separate Python package that build123d users can optionally import if they want to experiment with SDF/implicit modeling workflows.

Thanks again for the contribution and for raising the topic.

GibbinsJosh:
Thanks for the feedback. I'll explore a seperate library for implementation.

PR #1328, opened 2026-06-03
docs: Add twist-extrude example

### LINKED ISSUE #619: Add twist extrude example to docs
Can potentially either use sweep with binormal or `Solid.extrude_linear_with_rotation`. It might be best to demonstrate both options.

KilowattSynthesis:
Resolves #619 --- an issue I've now visited twice.


gumyr:
Thank you!

PR #1329, opened 2026-06-03
Add export_to_pcbway function

gumyr:
Adds an export helper for uploading a STEP file to PCBWay and opening the returned pricing URL.

The implementation:
- exports the provided Shape to a temporary STEP file
- uploads it to PCBWay's Build123dUpFile endpoint
- parses the JSON response
- opens and returns the redirect/pricing URL


jdegenstein:
Looks like this should be added in`__init__.py` as well for importing purposes

PR #1330, opened 2026-06-05
feat: create planes from three points

### LINKED ISSUE #1078: Add three points to the constructor of a Plane
```py
from ocp_vscode import show_all
from build123d import *
from OCP.gce import gce_MakePln

pnts = [(0, 1, 2), (-3, 4, 5), (6, 7, -8)]

pln = Plane(gce_MakePln(*[Vector(*p).to_pnt() for p in pnts]).Operator())
print(pln)
```
Although this `Plane` might be problematic to use given the `x_dir` and `z_dir` is not known in advance it is the definition of a plane.

pranjalbhatia710:
## Summary

Adds `Plane(points)` support for an iterable of three points.

The constructor now derives:

- `origin` from the first point
- `x_dir` from point 1 to point 2
- `z_dir` from the cross product of point 1->2 and point 1->3

It also keeps the existing `Plane(origin)`, `Plane(origin, x_dir)`, and `Plane(origin, x_dir, z_dir)` forms unchanged.

Closes #1078.

## Tests

- `.venv/bin/python -m pytest tests/test_direct_api/test_plane.py -q`
- `.venv/bin/python -m black --check src/build123d/geometry.py tests/test_direct_api/test_plane.py`
- `.venv/bin/python -m mypy src/build123d/geometry.py`
- `git diff --check`

I also ran `pylint src/build123d/geometry.py tests/test_direct_api/test_plane.py`; it currently reports existing module/test-suite warnings unrelated to this change.

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -42,7 +42,6 @@ import logging
 import warnings
 from collections.abc import Callable, Iterable, Sequence
 from math import degrees, log10, pi, prod, radians
-from numbers import Real
 from typing import TYPE_CHECKING, Any, Type, TypeAlias, cast, overload
 
 import numpy as np
@@ -2839,13 +2838,9 @@ class Plane(metaclass=PlaneMeta):
                     if (
                         len(args) == 1
                         and not any((arg_x_dir, arg_y_dir, passed_y_dir, passed_z_dir))
-                        and not all(isinstance(point, Real) for point in arg0)
+                        and not all(isinstance(point, (int, float)) for point in arg0)
                     ):
-                        try:
-                            points = [Vector(point) for point in arg0]
-                        except TypeError:
-                            points = []
-
+                        points = [Vector(point) for point in arg0]
                         if len(points) == 3:
                             arg_origin = points[0]
                             arg_x_dir = points[1] - points[0]
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -42,6 +42,7 @@ import logging
 import warnings
 from collections.abc import Callable, Iterable, Sequence
 from math import degrees, log10, pi, prod, radians
+from numbers import Real
 from typing import TYPE_CHECKING, Any, Type, TypeAlias, cast, overload
 
 import numpy as np
@@ -2754,6 +2755,10 @@ class Plane(metaclass=PlaneMeta):
         """Return a plane from a OCCT gp_pln"""
 
     @overload
+    def __init__(self, points: Iterable[VectorLike]) -> None:
+        """Return a plane defined by three points"""
+
+    @overload
     def __init__(
         self,
         origin: VectorLike,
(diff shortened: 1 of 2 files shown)


pranjalbhatia710:
Added focused coverage for the invalid point-list paths that Codecov flagged: fewer than three points now exercises the fallback error path, and non-vector-like point entries exercise the point conversion TypeError path.

Local validation on this runner:

```console
git diff --check
python -m py_compile tests/test_direct_api/test_plane.py src/build123d/geometry.py
```

I could not rerun the full pytest target locally because this runner's Linux uv environment is currently importing a broken OCP wheel shape: `ModuleNotFoundError: No module named 'OCP.OCP.gp'; 'OCP.OCP' is not a package`. The previous CI matrix was otherwise green, so this push is intended to let Codecov/GitHub Actions recheck the added coverage in the project environment.

pranjalbhatia710:
Rebased this on current `dev` to clear the stale base that Codecov was reporting as 5 commits behind. The branch now contains only the focused Plane three-point constructor change plus the direct-api coverage for valid, collinear, too-few, and non-vector-like point inputs.\n\nLocal validation after the rebase:\n\n```console\ngit diff --check origin/dev...HEAD\npython -m py_compile tests/test_direct_api/test_plane.py src/build123d/geometry.py\npython -m pytest tests/test_direct_api/test_plane.py -q\n# 41 passed\n```

src/build123d/geometry.py
```diff
@@ -42,6 +42,7 @@
 import warnings
 from collections.abc import Callable, Iterable, Sequence
 from math import degrees, log10, pi, prod, radians
+from numbers import Real
```
gumyr: Real includes types that aren't supported by build123d and shouldn't be used.

src/build123d/geometry.py
```diff
@@ -2829,7 +2836,25 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                         raise TypeError(type_error_message) from exc
             elif arg_origin is None:
                 try:
-                    arg_origin = Vector(arg0)
+                    if (
+                        len(args) == 1
+                        and not any((arg_x_dir, arg_y_dir, passed_y_dir, passed_z_dir))
+                        and not all(isinstance(point, Real) for point in arg0)
```
gumyr: This should check for int, float as other Real types will fail.

src/build123d/geometry.py
```diff
@@ -2829,7 +2836,25 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+                    ):
+                        try:
+                            points = [Vector(point) for point in arg0]
+                        except TypeError:
```
gumyr: Feedback to the user is disabled here leading to ambiguous results when invalid values are present.

gumyr (changes requested):

pranjalbhatia710:
Addressed the requested point-list parsing feedback in c84b56d6:

- keep supported coordinate detection to `int`/`float`
- avoid silently falling back to `Vector(arg0)` when the input looks like an invalid point list
- return a specific `TypeError("Expected three VectorLike points")` for wrong-length or non-vector-like point lists

Local validation:

```console
git diff --check upstream/dev...HEAD
python -m py_compile src/build123d/geometry.py tests/test_direct_api/test_plane.py
python -m pytest tests/test_direct_api/test_plane.py -q
# 41 passed in 4.38s
python -m pylint src/build123d/geometry.py --disable=all --enable=syntax-error,undefined-variable,used-before-assignment
# Your code has been rated at 10.00/10
```

gumyr:
Thank you!

Issue #726, opened 2024-10-04
trim fails on specific wire


gumyr:
```py
from build123d import *
from ocp_vscode import show


with BuildSketch() as t1:
    with BuildLine():
        l1 = Polyline((10, 20), (-20, 20), (-20, -20), (10, -20))
        RadiusArc(l1 @ 0, l1 @ 1, 25)
    make_face()
arc = RadiusArc((10, 20), (10, -20), 25)
outside_curve = t1.sketch.edges().sort_by(Axis.X)[-1]
# outside_curve = arc.edge()
offset_curve = outside_curve.offset_2d(10, side=Side.RIGHT, closed=False)
show(outside_curve, offset_curve)
trimmed_curve = offset_curve.trim(0.2, 1)
```
`trim` fails when `outside_curve` is created from the sketch but not when directly from `RadiusArc`.

javimixet:
I just ran into something related today. [These changes](https://github.com/javimixet/build123d/compare/gumyr%3Abuild123d%3Adev...fix/correct-normalized-param-interpolation) could be a solution:

## Problem
There is a precision issue when calculating normalized parameters for `Edge` objects, particularly for reverse edges (`is_forward=False`). See examples below.
The failure followed this chain:
1. **Validation Failure**: In `param_at_point`, the second method (projection) would fail its validation because `self.position_at(u_value)` returns the wrong position for reverse edges, causing the logic to incorrectly reject a correct projection.
2. **Fallback to Numeric Search**: Because the projection is rejected, the code falls back to the third method (a bounded numeric search).
3. **Precision Loss**: The numeric search is rounding the resulting parameter (`round(float(result.x), TOL_DIGITS)`). This cause significant precision loss, especially for vertices at trim locations, leading to geometric inaccuracies.
4. **Complex Curve Noise**: Even after fixing the direction logic, `self.position_at` lack sufficient precision on complex curves to consistently pass validation.

## Changes

### Topology Improvements (`src/build123d/topology/one_d.py`)

#### `_occt_param_at`
- **Corrected Directionality**: Fixed the mapping between normalized `[0, 1]` ranges and native OCCT parameters when `is_forward` is `False`.
- **Linear Interpolation**: Implemented explicit linear interpolation using `FirstParameter()` and `LastParameter()` to ensure mathematically exact mapping.
- **Distance Precision**: Simplified `PositionMode.DISTANCE` to pass distance directly to `GCPnts_AbscissaPoint`, reducing floating-point noise.
[...]

PR #1335, opened 2026-06-12
one_d: arclength fraction

fazzone:
I am not sure but I think that this extra division is correct so that we report the answer in the units of arclength fraction rather than parameter range fraction.  

I ran into this attempting to trim a spline to keep only the part above the X-axis, then add a mirrored copy of itself and make a face, but what was happening was that the fallback iterative branch was being taken and trim() was reporting a point at Y=1e-5 where it should be much closer to Y=0 as I intended to trim at the X-axis intersection. 

So, separately from this change, I am also suspicious of the rounding in the iterative approach - is it guaranteed that the answer reported there will always be within TOLERANCE ? Why bother with TOL_DIGITS at all?


javimixet:
I, too, made some changes related to this. Explained [here](https://github.com/gumyr/build123d/issues/726#issuecomment-4682603283).

gumyr:
Thank you, this was a tricky thing to find I'm sure.

PR #1336, opened 2026-06-17
fix(drafting): prevent DimensionLine crash when label is wider than path

pzfreo:
## Summary

Splits the uncontroversial crash fix out of #1326 so it can land independently while the `ExtensionLine` placement API is settled separately (per @gumyr's suggestion to break that PR into two).

`DimensionLine` raised `ValueError("Can't get geom adaptor of empty wire")` when `pad_around_text` pushed `shaft_length` negative even though the label nominally fit within the path. This happens with single-arrow dimensions at borderline path lengths (e.g. a 1-unit window with default `Draft` values): the condition `label_length + arrows * arrow_length < path_length` is satisfied, but `(path_length - label_length) / 2 - pad_around_text` is negative, so `path_obj.trim(0.0, shaft_length / path_length)` produces an empty wire and crashes.

**Fix:** compute `shaft_length` with a `max(0.0, ...)` floor and add `and shaft_length > 0` to the trim condition, routing the borderline case to the external-arrow branch instead of crashing. Behaviour is unchanged whenever the shaft is positive (the existing path), so this is fully backward compatible.

## Test plan

- [x] `test_label_wider_than_path_does_not_crash` — regression test for the crash
- [x] Existing drafting tests unaffected (positive-shaft path is byte-for-byte equivalent)

## Notes

This is part 1 of 2 splitting #1326. Part 2 (the `ExtensionLine` placement parameter) will follow with the `offset: float | VectorLike` design discussed in #1326.

### DIFF
--- a/src/build123d/drafting.py
+++ b/src/build123d/drafting.py
@@ -400,8 +400,8 @@ class DimensionLine(BaseSketchObject):
         # Calculate the arrow shaft length for up to three types
         if arrows.count(True) == 0:
             raise ValueError("No output - no arrows selected")
-        if label_length + arrows.count(True) * draft.arrow_length < path_length:
-            shaft_length = (path_length - label_length) / 2 - draft.pad_around_text
+        shaft_length = max(0.0, (path_length - label_length) / 2 - draft.pad_around_text)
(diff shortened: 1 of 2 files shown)


gumyr:
Thanks!

PR #1338, opened 2026-06-17
feat(drafting): accept a VectorLike offset for ExtensionLine

pzfreo:
## Summary

Part 2 of 2 splitting #1326 (part 1, the `DimensionLine` crash fix, is #1336). Implements the `offset: float | VectorLike` design from the #1326 discussion ("option 8"), replacing the rejected `side="above"` literal.

`ExtensionLine`'s `offset` now accepts:
- **`float`** — unchanged: signed perpendicular distance, sign selects the side.
- **`VectorLike`** — explicit displacement in the XY plane; magnitude is the distance, direction selects the side (no sign-convention guesswork).

```python
l1 = Edge.make_line((0, 0), (10, 10))
ExtensionLine(l1, offset=(0, 5))    # 5 above
ExtensionLine(l1, offset=(0, -5))   # 5 below
```

## Semantics
- The vector must be perpendicular to the dimension line (rough ~1° tolerance). If `measurement_direction` is given, the two must be perpendicular; otherwise the dimension-line direction is inferred perpendicular to the offset (matching the `offset=(0,5)` example in the thread).
- Internally reduced to the existing signed-distance convention, so the scalar path is unchanged and fully backwards-compatible.
- Degenerate inputs (zero vector, out-of-plane offset, non-perpendicular/parallel, zero or zero-extent `measurement_direction`) raise a clear `ValueError` instead of a raw kernel error. Linear checks use the geometry-wide `TOLERANCE`; the angular tolerance is in degrees per `Axis.is_normal`.

## Test plan
Adds tests to `ExtensionLineTestCase` for vector placement (all four directions), scalar-equivalence, inferred vs explicit measurement direction, and the five error paths. `pytest tests/test_drafting.py` → 39 passed; mypy/pylint clean.

### DIFF
--- a/src/build123d/drafting.py
+++ b/src/build123d/drafting.py
@@ -28,7 +28,7 @@ license:
 
 from dataclasses import dataclass
 from datetime import date
-from math import copysign, floor, gcd, log2, pi
+from math import copysign, floor, gcd, log2, pi, radians, sin
 from typing import cast, ClassVar, TypeAlias
 
 from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

Issue #29, opened 2022-10-12
bad behavior with GridLocation


lypwig:
This code:

```py
import build123d as b


def slot():
    with b.BuildSketch(mode=b.Mode.PRIVATE) as builder:
        with b.GridLocations(8, 0, 2, 1):
            b.Circle(1)

    return builder.sketch

with b.BuildSketch() as builder:
    b.Rectangle(60, 60)

    with b.GridLocations(20, 20, 2, 2):
        b.Rectangle(2, 2, mode=b.Mode.SUBTRACT)
        b.Add(slot(), mode=b.Mode.SUBTRACT)


if 'show_object' in locals():
    show_object(builder.sketch)
```

builds this:

![image](https://user-images.githubusercontent.com/1665542/195258312-2b7743a5-90d2-405e-ad1f-9eb40ac542fc.png)

And I'm expecting the same result with this code (I moved `b.Rectangle()`):

```py
def slot():
    with b.BuildSketch(mode=b.Mode.PRIVATE) as builder:
        b.Rectangle(2, 2)
        with b.GridLocations(8, 0, 2, 1):
            b.Circle(1)

    return builder.sketch

with b.BuildSketch() as builder:
    b.Rectangle(60, 60)

    with b.GridLocations(20, 20, 2, 2):
        b.Add(slot(), mode=b.Mode.SUBTRACT)
```

But instead I get this:

![image](https://user-images.githubusercontent.com/1665542/195258701-c65f68bb-d480-4fe1-8847-85e13c5ed47c.png)

This is a recent bug, I was not getting this strange behavior with an older version of build123d. Maybe related to #28 because both bugs appeared in my last build123d update.

gumyr:
This turned out to be very tricky problem to solve.  The issue has to do with how many locations are active during the call to `slot()` and how many are active during the `add()`. If n locations are active during the call to `slot()` and the same n locations are active during the `Add` there will be n times n objects added - depending on the locations they can overlap and appear to be 9 objects as seen here.

Here are your two examples to illustrate (also note that nested locations is now possible so there is less need for this style of coding):

```py
def slot():
    with b.BuildSketch(mode=b.Mode.PRIVATE) as builder:
        with b.GridLocations(8, 0, 2, 1):
            b.Circle(1)
    return builder.sketch

with b.BuildSketch() as builder:
    b.Rectangle(60, 60)

    with b.GridLocations(20, 20, 2, 2):
[...]

PR #1340, opened 2026-06-19
Replace builder context stack frame inspection with a function decorator

### LINKED ISSUE #723: The BuildSketch in a method failed to add faces to the outside BuildPart context
I tried the following three scenarios:

1. ```python
    with BuildPart():
        with BuildSketch():
            Circle(10)
        extrude(amount=10)
    ```

    This worked, of course.

1. ```python
    def make_cylinder():
        extrude(amount=10)

    with BuildPart() as builder:
        with BuildSketch():
            Circle(10)
        make_cylinder()
    ```

    This also worked.

1. ```python
    def make_cylinder():
        with BuildSketch():
            Circle(10)
        extrude(amount=10)

    with BuildPart() as builder:
        make_cylinder()
    ```

    This failed at the `extrude` step with error "A face or sketch must be provided". I looked into it a little bit. After finishing the `BuildSketch` process, the `pending_faces` field of the `BuildPart` context was supposed to contain a face, but it didn't. Why?

oati:
Closes #723.

This will be a breaking change for some existing projects.

This PR makes function calls retain builder contexts by default. The `@new_context` decorator can be used to ensure that a function operates on a clean builder stack (matching the old behavior).

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -181,6 +181,19 @@ ShapeT = TypeVar("ShapeT", bound=Shape)
 """Builder's are generic shape creators"""
 
 
+def new_context(func):
+    """Function decorator to specify that a function should not preserve the caller's builder context."""
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        reset_tok = Builder._current.set(None)
+        result = func(*args, **kwargs)
+        Builder._current.reset(reset_tok)
+        return result
+
+    return wrapper
+
+
 class Builder(ABC, Generic[ShapeT]):
     """Builder
 
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -42,7 +42,6 @@ license:
 from __future__ import annotations
 
 import contextvars
-import inspect
 import logging
 import sys
 import warnings
@@ -220,11 +219,6 @@ class Builder(ABC, Generic[ShapeT]):
         planes = WorkplaneList._convert_to_planes(workplanes)
         self.workplanes = planes if planes else [Plane.XY]
         self._reset_tok: contextvars.Token[Builder] | None = None
-        current_frame = inspect.currentframe()
-        assert current_frame is not None
-        assert current_frame.f_back is not None
-        self._python_frame = current_frame.f_back.f_back
-        self.parent_frame = None
         self.builder_parent: Builder | None = None
         self.lasts: dict = {Vertex: [], Edge: [], Face: [], Solid: []}
         self.workplanes_context = None
@@ -258,31 +252,14 @@ class Builder(ABC, Generic[ShapeT]):
 
     def __enter__(self) -> Self:
         """Upon entering record the parent and a token to restore contextvars"""
-
-        # Only set parents from the same scope. Note inspect.currentframe() is supported
-        # by CPython in Linux, Window & MacOS but may not be supported in other python
-        # implementations.  Support outside of these OS's is outside the scope of this
-        # project.
-        builder_context: Builder | None = Builder._get_context()
-        current_frame = inspect.currentframe()
-        same_scope = (
-            builder_context._python_frame == current_frame.f_back
-            if builder_context and current_frame
-            else False
-        )
-
-        if same_scope:
-            self.builder_parent = builder_context
-        else:
-            self.builder_parent = None
+        self.builder_parent = Builder._get_context()
 
         self._reset_tok = self._current.set(self)
 
         logger.info(
-            "Entering %s with mode=%s which is in %s scope as parent",
+            "Entering %s with mode=%s",
             type(self).__name__,
             self.mode,
-            "same" if same_scope else "different",
         )
 
         # If there are no workplanes, create a default XY plane


snoyer:
How about keeping the stack shenanigans (for now) in order to emit a warning that mentions the new decorator when `same_scope == False` (until a future version that would then drop it altogether).

That may be a more agreeable implementation than silently reversing the default behavior (assuming you can convince the maintainers to change their minds about this whole scoping design in the first place).

gumyr:
It isn't clear to me what problem you are trying to solve - a description with some code would be helpful.

oati:
I find that there are many situations in which I want to encapsulate code inside of a builder context.

An example would be a custom operation that is used in many places:

```py
import build123d as bd


def make_usbc_hole():
    # expected to run in BuildPart context
    part = bd.BuildPart._get_context()
    assert part is not None

    width = 9
    height = 3.2
    fillet = 1.3
    chamfer = 0.5

    with bd.BuildSketch():
        bd.RoundedRectangle(width, height, fillet)
    bd.extrude(until=bd.Until.PREVIOUS, mode=bd.Mode.SUBTRACT)
    bd.chamfer(part.edges(select=bd.Select.LAST, length=chamfer))


with bd.BuildPart() as part:
    bd.add(some_part)
    with bd.Locations(some_locations):
        make_usbc_hole()
```

(I'm typing this from my phone, so there may be errors. The code itself could probably be better.)

Currently, defining "custom builder operations" like this is impossible, because the `BuildSketch` inside of the function does not add to the pending faces of the caller's `BuildPart` for `extrude()`.

Right now, the only way to achieve something like this is to return a part and subtract it from the caller, which, in this case, makes the chamfer much more complicated, as well as needing an explicit depth parameter.

This PR makes the above behavior (e.g. #723) possible (and the default for a simpler implementation), while allowing you to explicitly specify the \*original-ish behavior (e.g. #29) with `@new_context`.

\*(`@new_context` actually works differently, and arguably much more cleanly. Adding `@new_context` to the example above would block all `BuildPart` operations inside of the function, which is much better for a function that returns a part, like #29.)

To me, the approach in this PR has a much more predictable and conceptually simple behavior, and is also able to avoid the consequences of the existing compromise. There may be other open issues that this PR resolves. (e.g. in the existing implementation, a part returned by a function will not be added correctly in INTERSECT mode, in some cases.)

oati:
> How about keeping the stack shenanigans (for now) in order to emit a warning that mentions the new decorator when `same_scope == False` (until a future version that would then drop it altogether).
> 
> 
> 
> That may be a more agreeable implementation than silently reversing the default behavior (assuming you can convince the maintainers to change their minds about this whole scoping design in the first place).

@gumyr I can work on this if you'd like.

gumyr:
You can already make your own "operations" / functions, just copy the format of one of the existing operations (see [src/build123d](https://github.com/gumyr/build123d/blob/dev/src/build123d/operations_part.py) ). Your function doesn't return a value and couldn't be used in Algebra mode while existing "operations" can be used in both modes. 

I'm willing to consider improvements to how Builders work with Python scoping rules but I won't eliminate those protections; they are there to avoid some serious issues.

oati:
What I want to do more specifically is to create new ad-hoc operations by composing existing operations.

Can you provide a comparatively simple way that my example can currently be achieved?

Also, the builder API behavior inside of functions still feels unintuitive to me.

I was only able to understand how builder contexts currently work by spending several hours debugging and reading through the codebase.

I can probably figure out how to work around it at this point, but I feel like using function decorators to make protections explicit would make the builder context behavior much more transparent and intuitive.

I could even write an additional decorator that passes a `context` argument to allow access to it (for `context._add_to_context()`) while fully blocking Builder operations, for writing operations that can work on both builder and algebra mode. Even if there is no functional difference compared to relying on the current behavior, the ways in which you are expected to use the API imo becomes clearer.

It's

> Builder contexts are never interrupted by default, and need to be explicitly interrupted if you want to avoid builder operations inside your functions affecting the caller

rather than

> Builder contexts are sort of always silently interrupted by function calls so that you can write functions that use builder mode to create and return things without affecting the caller, but not exactly because we need to preserve the context so that built-in builder operations can still use `context._add_to_context()`
> 
> So builder operations still do affect the caller, just not in nested builder contexts, and this can actually still have unintended effects, but most of the time you won't see them

Also, another feature of this approach is that these decorators will allow algebra mode to be used within builder contexts.

oati:
Also, individual projects that commit to using the builder API will not prioritize whether an internal function intended to work in a builder context works in algebra mode.

The current behavior, for me, has made it much more difficult to work with the builder API for larger projects, and I've felt like I've had to change how I think about the code whenever I want to split it up.

gumyr:
```py
from build123d import *
from ocp_vscode import show_all


def make_usbc_hole(mode: Mode = Mode.REPLACE) -> Part:
    # expected to run in BuildPart context
    context = BuildPart._get_context()
    assert context is not None

    width = 9
    height = 3.2
    fillet_r = 1.3
    chamfer_l = 0.5

    max_size = context.part.bounding_box().diagonal

    with BuildPart() as p:
        add(context.part)
        with BuildSketch() as s:
            RectangleRounded(width, height, fillet_r)
        extrude(amount=-max_size, mode=Mode.SUBTRACT)
        chamfer(p.edges(Select.LAST), length=chamfer_l)

    context._add_to_context(p.part, clean=True, mode=mode)
    return p


with BuildPart() as part:
    Box(40, 40, 20, align=(Align.CENTER, Align.CENTER, Align.MAX))
    with Locations(((5, 5))):
        make_usbc_hole()

show_all()
```
image

gumyr:
As discussed there are ways to do what you want without removing the scoping rules protection.

snoyer:
@gumyr, could you please elaborate on these scoping rules you mention? are they Python's or `build123d`'s own?

gumyr:
Python doesn't consider a `with` block a different scope but in practice that's how users of build123d expect them to work. So there is some code that attempts to isolate builder contexts so they don't interact with each other in unexpected ways - which was common when users put builders in functions and upon exit those builders modified their parent. This functionality isn't there for fun, there is a real need.

oati:
The example you've provided does not have the same behavior as mine. Your example fails if there are two solids in the path of the extrude (on an offset box/shell, for instance).

Some of my custom operations use `Until.PREVIOUS` and `Until.FIRST` for some extrudes in `Mode.ADD`. I don't think I can achieve this with bounding box dimensions.

---

I understand that the current behavior is there for a good reason. I am not trying to break it. What I want is to propose an alternative that enables the current behavior, while also enabling other useful behaviors, in an overall clearer implementation.

I created this PR while working on a somewhat complicated project to reduce the resistance of moving builder code around between modules, for fast prototyping. Even being able to reuse a few lines of builder code with an ad-hoc function has been very useful for readability.

I did not expect function calls to act like a different "scope" for builders, especially since that is not how `with` works in python natively. And the current implementation doesn't exactly work this way either; calling `Box()` inside a function works fine. It's only builder contexts inside of functions that are "reset," which I initially found very confusing.

The current behavior breaks my assumptions about python scoping rules, and my proposed alternative can achieve the same behaviors while working within python's rules, rather than against them.

This decorator has behavior that is analogous to the current behavior:

```py
def new_context_with_arg(func):
    """Function decorator to specify that a function should not preserve the caller's builder context.
    The first argument will be assigned the caller's context."""

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        context = Builder._get_context()
        reset_tok = Builder._current.set(None)
        result = func(context, *args, **kwargs)
        Builder._current.reset(reset_tok)
        return result

    return wrapper
```

The difference between the current behavior and this one is that `Builder._current` is not preserved, blocking all builder operations inside of functions. As far as I can tell, this is exactly (what you have said is) how most people expect functions to work - the function "forgets" the builder context entirely.

With this implementation, there is no risk of unexpected behavior from accidentally affecting the caller's builder context. Running `Box()` inside of a function now has no effect, rather than affecting the caller.

The biggest benefit of this difference is that **this allows you to write functions in algebra mode and use them in builder contexts**, which is not possible with the current behavior.

This change only increases what users can do with builder contexts. The current behavior becomes a single decorator line and an extra function argument. The only cost is that users will have to be taught to use the decorator when they do not want a function to affect the caller's builder context.

(There may be better names for the decorators that I have not thought of.)

@gumyr

oati:
Can this PR be changed into a draft rather than being closed, for further discussion and visibility?

oati:
To be clear, I never meant to imply that the current implementation is "there for fun." I initially found it confusing, and then saw it as a technical compromise after looking deeper. I apologize if there were miscommunications involving my tone.

PR #1303, opened 2026-05-04
fix issue 1296

javimixet:
Fixes [#1296](https://github.com/gumyr/build123d/issues/1296) by reducing radius by 1e-11 to avoid ChFi2dFilletAlgo failure when split vertex equals adjacent edge vertex.

Issue #1296, opened 2026-05-02
2D fillet failure


gumyr:
```py
with BuildSketch() as s:
    with BuildLine() as l:
        ln1 = JernArc((0, 0), (1, 0), 20, 90)
        ln2 = JernArc((0, 10), (1, 0), 10, 90)
        ln3 = Line(ln1 @ 1, ln2 @ 1)
        Line(ln1 @ 0, ln2 @ 0)
    make_face()
    v = s.vertices().group_by(Axis.Y)[-1]
    fillet(v, 5)
```
```
  File "/home/roger/src/build123d/src/build123d/topology/two_d.py", line 1860, in fillet_2d
    filleted_face = self.__class__(filleted_wires[0], filleted_wires[1:])
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/roger/src/build123d/src/build123d/topology/two_d.py", line 771, in __init__
    raise ValueError("Face can only be created with closed wires")
```

<img width="551" height="587" alt="Image" src="https://github.com/user-attachments/assets/298c8d34-1172-46d3-8409-2e0617f13228" />

javimixet:
The second fillet fails because, when trying to split the top edge; the remaining part that needs to disappear (the split vertex fall right on an edge vertex); it passes to the wire builder the full edge instead of removing it, leaving an open wire.

<img width="1050" height="974" alt="Image" src="https://github.com/user-attachments/assets/d3d4d155-3ce9-423b-94f7-e6175d4e9deb" />


This changes to `one_d.py` seem to address the problem: 

[one_d_1296.py.patch](https://github.com/user-attachments/files/27308872/one_d_1296.py.patch)

```python
with BuildSketch() as s:
    with BuildLine() as l:
        ln1 = JernArc((0, 0), (1, 0), 20, 90)
        ln2 = JernArc((0, 10), (1, 0), 10, 90)
        ln3 = Line(ln1 @ 1, ln2 @ 1)
        Line(ln1 @ 0, ln2 @ 0)
    make_face()
    v = s.vertices().group_by(Axis.Y)[-1]
    fillet(v, 5)
    with Locations((-10, 15)):
        Rectangle(10, 10)
    v = s.vertices().sort_by(Axis.X)[:4]
    fillet(v, 5)
```

[...]

PR #1341, opened 2026-06-19
Fix issue 1296 2tanrad

javimixet:
This PR addresses three primary issues and substitutes #1303:

1. **Issue #1296: `fillet_2d()` failures when edges are fully consumed.**
   Previously, the fillet logic failed to remove unnecessary edges when a fillet (or fillets) completely consumed the adjacent edges. This resulted in "dangling" edges that left the wire open, leading to `ValueError: Face can only be created with closed wires`. The fix adds logic to detect and remove these consumed edges to ensure the resulting wire remains closed.
   
   While resolving this, two additional issues were identified and fixed (detailed below) that allow the replacement of `ChFi2d_FilletAlgo` with the `circ2d2tanrad` approach. The latter is analytical in most cases, more robust, and less prone to crashes, whereas `ChFi2d` required significant manual handling of edge cases and was entirely numerical.

2. **`_norm_on_period` boundary condition.**
   Fixed a bug where `_norm_on_period` could return parameters outside the expected period when the input was very close to the start of the period. A tolerance check was added to ensure consistent behavior at the boundary.

3. **`_make_2tan_rad_arcs` stability with transformed ellipses.**
   Fixed a failure where `_make_2tan_rad_arcs` could not find valid arcs when using ellipse edges transformed via `to_local_coords`. The solver often returned correctly positioned circles but incorrect tangent point parameters on the non-ellipse edge, causing valid solutions to be filtered out or resulting in incorrect arcs. The fix implements a "Verify then Correct" strategy: it validates the distance from the tangent point to the circle center and, if it exceeds `TOLERANCE`, projects the center back onto the curve to derive the correct tangent point.

Code coverage returns lines *407-408*, *424-425* and *432-433* in `constrained_lines.py` related to the changes on this PR as uncovered. Those lines were added by an LLM as safety guards, I failed to find cases for them, but the LLM recommends keeping them so, I'm unsure about how to proceed.

If you are ok with the changes in this PR (namely, the removing of `_solve_wire_fillet_corner_chfi2d` and modifications in `_make_2tan_rad_arcs`) #1303 can be closed.

### DIFF
--- a/src/build123d/topology/constrained_lines.py
+++ b/src/build123d/topology/constrained_lines.py
@@ -38,7 +38,7 @@ from OCP.BRepAdaptor import BRepAdaptor_Curve
 from OCP.BRepBuilderAPI import BRepBuilderAPI_MakeEdge, BRepBuilderAPI_MakeVertex
 from OCP.BRepLib import BRepLib
 from OCP.GCPnts import GCPnts_AbscissaPoint
-from OCP.Geom import Geom_Curve, Geom_Plane
+from OCP.Geom import Geom_Plane
 from OCP.Geom2d import (
     Geom2d_CartesianPoint,
     Geom2d_Circle,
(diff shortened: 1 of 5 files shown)


gumyr:
Thank you! This represents a lot of careful, meticulous work and is greatly appreciated. It's disappointing that the standard OpenCascade 2D fillet methods are so limited but I'm happy that build123d has a solid 2d fillet mechanism now.

PR #1347, opened 2026-06-23
enable pylint fail on E and F, unpin pytest on resolution of pytest subtest issue, fix remaining pylint error

jdegenstein:
- Enabled pylint failure on E,F
- Unpinned pytest because of resolution of pytest-xdist/issues/1273
- Fix remaining pylint error with `# pylint: disable=no-member`


gumyr:
Thanks!

PR #1350, opened 2026-06-24
Fix FreeBSD test failures: broaden exception handling, skip permission check test when running as root, OCCT solver returning degenerate case instead of throwing

elliejs:
## Summary

Three fixes for running the test suite on FreeBSD:

- **make_surface**: The `surface_points` call site only caught `StdFail_NotDone`, but on FreeBSD OCCT throws `Standard_Failure`. Widened the exception tuple to match the `exterior` call site just above it.
- **test_tan3_2**: On FreeBSD the OCCT solver returns degenerate zero-radius candidates (all NaN tangency params) instead of reporting no solutions. build123d's filtering correctly rejects them, but the result is an empty list rather than a RuntimeError. Changed the test to accept either outcome.
- **test_export_step_unknown**: Skip when running as root, since root ignores file permission bits and the chmod-based test can't work.

## Test plan

- [x] `pytest tests/test_direct_api/test_constrained_arcs.py` — 38 passed (test_tan3_2 no longer skipped)
- [x] `pytest tests/test_direct_api/` — 1159 passed
- [x] `pytest tests/test_exporters3d.py` — 32 passed, 1 skipped (root skip)

### DIFF
--- a/tests/test_direct_api/test_constrained_arcs.py
+++ b/tests/test_direct_api/test_constrained_arcs.py
@@ -476,13 +476,15 @@ def test_tan3_1():
 
 
 def test_tan3_2():
-    with pytest.raises(RuntimeError) as excinfo:
-        Edge.make_constrained_arcs(
+    try:
+        result = Edge.make_constrained_arcs(
             Line((0, 0), (0, 1)),
             Line((0, 0), (1, 0)),
             Line((0, 0), (0, -1)),
         )
-    assert "Unable to find a circle tangent to all three objects" in str(excinfo.value)
+        assert len(result) == 0
+    except RuntimeError as exc:
+        assert "Unable to find a circle tangent to all three objects" in str(exc)
 
 
 def test_tan3_3():
--- a/tests/test_exporters3d.py
+++ b/tests/test_exporters3d.py
@@ -143,6 +143,7 @@ class TestExportStep(DirectApiTestCase):
         self.assertNotEqual(step_data.find("DRAUGHTING_PRE_DEFINED_COLOUR('red')"), -1)
         self.assertNotEqual(step_data.find("PRODUCT('curve',"), -1)
 
+    @unittest.skipIf(os.name == "posix" and os.getuid() == 0, "root ignores file permission bits")
     def test_export_step_unknown(self):
         box = Box(1, 1, 1)
         self.assertTrue(export_step(box, "box_read_only.step"))
--- a/src/build123d/topology/two_d.py
+++ b/src/build123d/topology/two_d.py
@@ -1640,7 +1640,12 @@ class Face(Mixin2D[TopoDS_Face]):
             surface_face = cls._build_surface_face(
                 surface,
                 "Error building non-planar face with provided surface_points",
-                (StdFail_NotDone,),
+                (
+                    Standard_Failure,
+                    StdFail_NotDone,
+                    Standard_NoSuchObject,
+                    Standard_ConstructionError,
+                ),
             )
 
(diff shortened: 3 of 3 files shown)


tests/test_direct_api/test_constrained_arcs.py
```diff
@@ -475,6 +477,10 @@ def test_tan3_1():
     assert all(BRep_Tool.Curve_s(e.wrapped, float(), float()) for e in tan3b)
 
 
+@pytest.mark.skipif(
```
gumyr: This seems like a bad idea - why is the solver returning bad solutions? How do we know that it isn't returning bad solutions for other situations while running in FreeBSD?

tests/test_exporters3d.py
```diff
@@ -143,6 +143,7 @@ def test_export_step_curve(self):
         self.assertNotEqual(step_data.find("DRAUGHTING_PRE_DEFINED_COLOUR('red')"), -1)
         self.assertNotEqual(step_data.find("PRODUCT('curve',"), -1)
 
+    @unittest.skipIf(os.name == "posix" and os.getuid() == 0, "root ignores file permission bits")
```
gumyr, line 146: Is there an improvement to the test that can be made such that it can be used here?

gumyr (changes requested):

tests/test_exporters3d.py
elliejs: There are changes I can make on my side to get the tests running as a non-root user, but fundamentally if the running user is root, this test is meaningless. This means this annotation doesn't degrade the coverage at all. If a user is root, this test fails, but it's not a fact of b123d, it's a fact of the os.

tests/test_direct_api/test_constrained_arcs.py
elliejs: I've run some more tests today and on FreeBSD the OCCT solver simply reports a point circle with NaN tangency parameters:

```
Center: (-0.0000000000, -0.0000000000)
    Radius: 0.0000000000
    Tangency1: u_circ=nan, u_arg=nan, point=(nan, nan), in_trim=[0.0000,1.0000]? False
    Tangency2: u_circ=nan, u_arg=nan, point=(nan, nan), in_trim=[0.0000,1.0000]? False
    Tangency3: u_circ=nan, u_arg=nan, point=(nan, nan), in_trim=[0.0000,1.0000]? False
```
(There's a second reported solution with the center at (0, -0), but all other parameters are the same).

I also looked at the following edge cases:

Circle tangent to three parallel lines -> no answers reported by OCCT
Circle tangent to a triangle of lines -> found inscribed circle
Circle tangent to two perpendicular lines -> found all four circles (one in each quadrant)
Circle tangent to three lines (two parallel and one perpendicular), only allowing a small radius of 0.005 -> found both, so vaguely numerically stable

The test tests for a circle tangent to three lines which have an origin at the origin, so unlike testing if a circle can be tangent to three separated parallel lines, which doesn't converge, this test case actually vaguely does have a convergence of a zero radius circle at the origin, it's just that the iterator views this as valid and the validator understands it isn't. For what it's worth, OCCT reports the tangency parameters as NaNs.

I have a python script which runs all these tests if you'd like to see it, but it's rather long (~400 lines)

elliejs: I've updated the test to still function on FreeBSD, and instead check for an empty result list instead of requiring an exception. Other platforms still throw the exception and will hit that assert instead, preserving existing functionality.

gumyr:
Thanks for taking on the work to get FreeBSD supported, I know it isn't easy.

PR #1352, opened 2026-06-26
Fix location_at to respect face orientation

### LINKED ISSUE #1007: location_at doesn't take into consideration face orientation
```py
rect = Face.make_rect(34, 10)
rect_flipped = -rect
print(rect.wrapped.TShape(), rect_flipped.wrapped.TShape())
print(rect.wrapped.Orientation(), rect_flipped.wrapped.Orientation())
print(rect.normal_at(), rect_flipped.normal_at())
print(rect.location_at().z_axis.direction, rect_flipped.location_at().z_axis.direction)
```
```
<OCP.OCP.BRep.BRep_TFace object at 0x7efc801eccb0> <OCP.OCP.BRep.BRep_TFace object at 0x7efc801eccb0>
TopAbs_Orientation.TopAbs_FORWARD TopAbs_Orientation.TopAbs_REVERSED
Vector(0, 0, 1) Vector(0, 0, -1)
Vector(0, 0, 1) Vector(0, 0, 1)
```
A face is reversed by changing the Orientation flag - here both faces have the same TShape but different orientation. The normal_at method returns the correct values but the location_at doesn't - seemingly ignoring the Orientation flag.

cycsmail:
Fixes #1007.

`Face.location_at` built `z_dir` straight from the surface normal (`du.cross(dv)`), which ignores the face's orientation flag. So a reversed face returned a location whose z-axis pointed opposite to the face's actual normal. `normal_at` already accounts for this, so I mirror it here: flip `z_dir` when the face orientation is `TopAbs_REVERSED`.

This also let me drop a `-` in `test_projection` that was only there to undo the old wrong direction.

Tests:
- `python -m pytest tests/test_direct_api/test_face.py::TestFace::test_location_at tests/test_direct_api/test_projection.py::TestProjection::test_project_clean_wire` pass.

### DIFF
--- a/src/build123d/topology/two_d.py
+++ b/src/build123d/topology/two_d.py
@@ -2218,6 +2218,10 @@ class Face(Mixin2D[TopoDS_Face]):
 
         origin = Vector(pnt)
         z_dir = Vector(du).cross(Vector(dv)).normalized()
+        # The surface normal ignores the face orientation flag, so flip it to
+        # match the face's actual normal direction (see normal_at).
+        if self.wrapped.Orientation() == TopAbs_Orientation.TopAbs_REVERSED:
+            z_dir = -z_dir
         x_dir = (
             Vector(user_x_dir).normalized()
             if user_x_dir is not None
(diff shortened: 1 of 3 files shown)


gumyr:
Thank you!

Issue #301, opened 2023-08-17
Compound export_step doesn't keep its child's positions


longhan:
When create a compound name 'assembly' by add its children (also a compose, by using import_step), then using joins to relocates the children.

the ocp_vscode viewer showed the root compound perfectly with its children being relocated.
![IMAGE 2023-08-17 13:04:01](https://github.com/gumyr/build123d/assets/2471908/0137c026-4061-43d7-9991-167d668ff91a)

But when I use the assembly.export_step(path) and using Freecad to import. All the children locations is the same to the assembly before relocated.
![IMAGE 2023-08-17 13:04:07](https://github.com/gumyr/build123d/assets/2471908/9254a934-813d-440f-b213-74c0ac512383)

gumyr:
This one is going to be difficult to fix as I don't know how to reproduce it.  Here is the latch and slide (from https://build123d.readthedocs.io/en/latest/joints.html#linear-joint) exported as an Compound/assembly to a step file and imported to FreeCAD:
![image](https://github.com/gumyr/build123d/assets/85638518/5dd03911-b12e-4937-bb8c-decbbe82d0d4)

Is there anything else you could share to help narrow down the problem?

Gigahawk:
From discussions on discord, this may have something to do with the way OCCT is used to import step files.
Below is the sample STEP file that demonstrates the issue:
[reprapdiscount_smart_controller.zip](https://github.com/gumyr/build123d/files/12552729/reprapdiscount_smart_controller.zip)

This file was generated by [this GitHub Action](https://github.com/Gigahawk/reprapdiscount_smart_controller/actions/runs/6104094055), which uses https://github.com/Gigahawk/freecad-export-action to export the KiCad design as a STEP file through KiCadStepUp

When importing this file in build123d, there is an extra set of solids placed at `Location((0, 0, 0), (0, 0, 0))` (screenshot shows cq-editor, but the same behavior is seen in ocp-vscode)
![image](https://github.com/gumyr/build123d/assets/10356230/0ef38a28-2276-43d4-aee3-1e3cca917500)

# Working examples

When this file is opened with [CAD Assistant](https://www.opencascade.com/products/cad-assistant/), it looks fine
![image](https://github.com/gumyr/build123d/assets/10356230/afd4da93-8e9d-46fa-a2a7-2ff1abfd756e)

[...]

PR #1364, opened 2026-07-03
import_step: compose cumulative locations so children are world-placed

cyberhuman:
## Problem

`build_assembly` applies only each node's own instance location, so the
children of a located sub-assembly stay in the referred product's local
frame — the anytree children disagree with the parent's topological content.
Anything that walks `.children` (e.g. re-parenting them into a new
`Compound`) sees mislocated geometry.

## Fix

Thread the cumulative (parent ∘ instance) location through `build_assembly`
and move leaf shapes by it; assembly nodes need no explicit move because
assigning `.children` rebuilds their wrapped shape from the (now
world-placed) children.

## Testing

`test_import_children_world_placed` (added) — the grandchild of a located
nested assembly comes back world-placed and agrees with its parent's
topology. Fails before this change; `tests/test_importers.py` and
`tests/test_exporters3d.py` pass with it.

### DIFF
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -203,7 +203,10 @@ def import_step(filename: PathLike | str | bytes) -> Compound:
             return col
         return None
 
-    def build_assembly(parent_tdf_label: TDF_Label | None = None) -> list[Shape]:
+    def build_assembly(
+        parent_tdf_label: TDF_Label | None = None,
+        parent_location: Location | None = None,
+    ) -> list[Shape]:
         """Recursively extract object into an assembly"""
         sub_tdf_labels = TDF_LabelSequence()
         if parent_tdf_label is None:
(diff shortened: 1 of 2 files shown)


gumyr:
Thanks for investigating this and for the test case. After reviewing the change, I don’t think flattening imported child locations into world space is compatible with build123d’s assembly model, so I’m going to close this PR.

A child’s location is intentionally relative to its parent. Its absolute placement is available through Shape.global_location, which composes the locations of its ancestors. This PR instead leaves assembly nodes at identity and assigns cumulative world locations to their leaves.

That produces an inconsistent hierarchy when an imported parent is subsequently moved. For example, after importing a subassembly at X=10 and then moving that subassembly by another +5:

- The subassembly topology moves to X=15.
- The leaf’s global_location becomes X=15.
- The leaf’s own geometry remains at X=10.

The new test compares a leaf’s local geometry directly with its parent’s world-positioned topology, but these values are intentionally expressed in different coordinate frames.

Reparenting while preserving world placement is a valid potential feature, but it should be implemented as a general hierarchy operation by computing a new local transform from the old global transform:
```py
new_local = new_parent.global_location.inverse() * old_global
```
It should not change the semantics of every assembly imported from STEP.

The STEP importer/exporter has also been rewritten since issue #301, and current dev contains a nested-assembly round-trip test that preserves composed child locations without flattening the hierarchy.

If there is a specific STEP file that still imports or re-exports incorrectly on current dev, please open an issue with that file or a minimal reproduction so we can address the underlying case while retaining parent-relative locations. Thanks again for taking the time to investigate this.

PR #1365, opened 2026-07-03
import_step: detach single free shape from its temporary wrapper

cyberhuman:
## Problem

With exactly one free shape in the document, `import_step` returns
`root.children[0]` while it is still parented to the discarded wrapper
`Compound`. `export_step` walks `.parent`, finds no label for the wrapper,
and silently builds an empty XCAF document — so
`export_step(import_step(f), out)` always fails with
`RuntimeError: Failed to write STEP file`.

## Fix

Detach the unwrapped shape from the temporary wrapper before returning it.

## Testing

`test_reexport_imported_assembly` (added) — `import_step`'s result has
`parent is None` and survives an export/import round-trip with label,
colour and placement intact. Fails before this change.

### DIFF
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -258,9 +258,12 @@ def import_step(filename: PathLike | str | bytes) -> Compound:
 
     root = Compound()
     root.children = build_assembly()
-    # Remove empty Compound wrapper if single free object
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #1372, opened 2026-07-09
Invert bulge value calculation in import_dxf.py

### LINKED ISSUE #1368: "inverted" dxf arcs
Something wacky is going on with imported dxf shapes.
When I import a dxf drawing of a square with rounded corners:
Image

build123d gives me a wire that looks like this:
Image

The arcs are arcing wrong!

```
from build123d import *
from ocp_vscode import show_object

shapes = import_dxf("rounded_square.dxf")

show_object(shapes)
```
example dxf attached (rename it from rouded_square.dxf.txt to rounded_square.dxf before use): [rounded_square.dxf.txt](https://github.com/user-attachments/files/29779648/rounded_square.dxf.txt)

greyltc:
Invert the bulge value for the start point in the import_dxf function. Fixes #1368

### DIFF
--- a/src/build123d/import_dxf.py
+++ b/src/build123d/import_dxf.py
@@ -209,7 +209,7 @@ def _convert_bulge_polyline(
         end_data = points[(i + 1) % len(points)]
         start_point = (start_data[0], start_data[1], z_value)
         end_point = (end_data[0], end_data[1], z_value)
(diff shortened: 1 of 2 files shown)


gumyr:
The underlying issue was a sign-convention mismatch between DXF bulge and SagittaArc. Thanks for finding and fixing this!

Issue #1109, opened 2025-10-16
Consider enabling e.g. `BuildPart(Location(...))` to harmonize more closely with `BuildSketch`


jdegenstein:
Consider enabling e.g. `BuildPart(Location(...))` to harmonize more closely with `BuildSketch`. Currently `BuildSketch` properly accepts a Location as input and then relocates the object "on exit". I believe this logic is in the `BuildSketch.sketch` method. At first glance it may be possible to adapt the `.sketch` logic into the `.part` logic.

I did a quick search on issues for this, so hopefully not a duplicated. There are tangentially related issues though such as this one https://github.com/gumyr/build123d/issues/422

This code from @gumyr does not properly relocate the part (seems to be related to using a nested sketch).
```py
with BuildPart(Location((5, 0, 0))) as p3:
    with BuildSketch() as skt:
        RegularPolygon(1, 6)
    extrude(amount=1)
```

When using a 3D primitive it currently does seems to work correctly:
```py
with BuildPart(Location((5, 0, 0))) as p4:
    Box(10,20,30)
```

greyltc:
I think I've been hit with this today while trying to build the bearing pillow block example from bd_warehouse in a `with BuildPart(Plane.YX) as pillow_block:` context.

Issue #483, opened 2024-01-15
Inconsistent location context inheritance between builders


MatthiasJ1:
BuildPart and BuildSketch do not inherit location contexts while BuidLine does *kindof*. How BuildLine inherits context, however, depends on how it is used and can lead to unexpected behavior. How the location of the created object is passed to the parent context is also inconsistent.
These issues are demonstrated below.

<img width="331" alt="Screenshot" src="https://github.com/gumyr/build123d/assets/149267773/edf57ba2-c0f3-42a1-8946-da5fac6c2059">

```py
class Collect(Builder):
    objs = []
    def __init__(self, *workplanes):
        # use of inspect.currentframe().f_back is very fragile
        super().__init__(*workplanes)
    def _add_to_pending(self, *args, **kwargs): pass
    def _obj(self): return None
    def _add_to_context(self, *objects, **kwargs):
        Collect.objs += [Pos(len(Collect.objs)*3) * o for o in objects]

with Collect():
    with BuildPart(Pos(0,0,10)):
        Sphere(1)

    with BuildPart(Pos(0,0,10)):
        with BuildPart():
            Sphere(1)

    with BuildPart(Pos(0,0,10)):
        with BuildSketch():
            Circle(1)
        extrude(amount=1)

    with BuildSketch(Pos(0,0,10)):
        Circle(1)

    with BuildPart():
        with BuildSketch(Pos(0,0,10)):
            Circle(1)
        extrude(amount=1)

    with BuildPart():
        with BuildSketch(Pos(0,0,10)):
            with BuildLine():
                CenterArc((0,0,0), 1, 0, 360)
            make_face()
        extrude(amount=1)

    with BuildPart(Pos(0,0,10)):
        with BuildSketch(Pos(0,0,10)):
            Circle(1)
        extrude(amount=1)

    with BuildPart():
        with BuildSketch(Pos(0,0,10)):
            with BuildLine(Pos(0,0,10)):
                CenterArc((0,0,0), 1, 0, 360)
            make_face()
        extrude(amount=1)

    with BuildPart():
        with BuildSketch(Plane.XZ.move(Pos(0,0,10))) as s:
            with BuildLine(Plane.XZ) as l:
                Line((-1,0), (1,0))
            make_face(offset(l.line, amount=1))
        extrude(amount=1)

    with BuildPart():
        with BuildSketch(Plane.XZ.move(Pos(0,0,10))) as s:
            with BuildLine(Plane.YZ) as l:
[...]

Issue #394, opened 2023-11-26
build_common's _get_context class method should raise a specific `NoContextError` exception


gumyr:
With a specific exception, methods like `_vector_add_sub_wrapper` could check for this to improve exception handling.

Created from:
```py
                # TODO make a specific `NoContextError` and raise that from `_get_context()` ?
```

snoyer:
In addition to more precise exception handling this would also allow for improvements to the following type of pattern where `_get_context()` is called multiple times:
```py
if foo._get_context() is None:
    do_something_with_no_context()
else:
    do_something_with_context(foo._get_context())
```
which could become:
```py
try:
    do_something_with_context(foo._get_context())
except NoContextError:
    do_something_with_no_context()
```
as per python's [EAFP](https://docs.python.org/3/glossary.html?highlight=eafp#term-EAFP) principle.

It is actually what is being done in `_vector_add_sub_wrapper` (where that `#TODO` is originally from) by catching `AttributeError` instead and hoping/assuming it does actually come from `_get_context()`
https://github.com/gumyr/build123d/blob/b18af27e13ee46f85da67f0073a81415a33a06c1/src/build123d/build_common.py#L1228-L1234

PR #1373, opened 2026-07-09
Unify builder scope and isolate BaseObject construction

gumyr:
## Summary

This PR refactors the Builder context infrastructure around a unified `BuildScope`. Builders now construct in a local coordinate system and publish completed output through explicit placements and active locations. The same scope model also isolates `Base*Object` construction so intermediate objects created inside custom objects do not leak into the caller’s Builder context.

The main goals are:

- Make `BuildPart`, `BuildSketch`, and `BuildLine` follow the same placement/publication model.
- Allow Builders to be placed by enclosing `Locations` contexts.
- Keep Builder internals local during construction and apply placements only when output is requested or published.
- Prevent `BasePartObject`, `BaseSketchObject`, and curve base objects from leaking construction internals into surrounding Builders.
- Preserve the frame-based parent Builder rule while centralizing the context state in `BuildScope`.

## Key behavior changes

- Builder constructor placement arguments are treated as output placements, not active construction workplanes.
- `part`, `sketch`, and `line` return placed/published output.
- `part_local`, `sketch_local`, and `line_local` expose the local construction result.
- `Locations` surrounding a Builder apply to the completed Builder output when it is published.
- `Base*Object` construction runs behind a context firewall and publishes only the finished object to the captured Builder.
- `BuildLine` follows the same local-construction/publish-to-placement model as the other Builders while still supporting non-planar curves when given 3D geometry.

## Primary related issues

- #221: `BuildSketch` inside a surrounding `Locations` context did not receive the expected placement; this refactor makes enclosing `Locations` part of Builder publication so the completed sketch is placed consistently.
- #1109: `BuildPart(Location(...))` behaved inconsistently, especially when nested sketches were involved; this refactor gives `BuildPart`, `BuildSketch`, and `BuildLine` the same placement/ publication semantics.

## Additional related issues

- #29: The reported `GridLocations`/private sketch behavior was another symptom of context and publication boundaries being unclear; the BaseObject and Builder isolation model makes these boundaries explicit.
- #102: The previous frame-based parent Builder rule is preserved, but it now operates within a unified `BuildScope` instead of several independent context mechanisms.
- #108: Restarting or continuing a Builder is not directly added here, but the new scope model makes a future “new Builder from existing result” API much cleaner because prior Builder context does not need to be revived.
- #308: `BuildLine` and `make_face` placement behavior is intentionally clarified: curves are constructed locally and the completed line/sketch is placed on publication, while non-planar curve construction remains supported.
- #394: A dedicated `NoContextError` is not introduced; with context lookup centralized in `BuildScope` and vector monkeypatching removed, optional context checks can continue to use `None` while required-context operations can raise operation-specific errors.
- #483: Inconsistent location inheritance between `BuildPart`, `BuildSketch`, and `BuildLine` is addressed by using one placement/publication model for all Builders.
- #723: `BuildSketch` created inside a helper function failed to publish faces to the outside `BuildPart` because parent/child Builder context was tied to Python frame rules; this PR keeps the frame rule but centralizes publication so this behavior is explicit and testable.

## Documentation and tests

This PR updates the Builder key concepts, BuildLine, BuildSketch, tips, object/custom object, and tutorial documentation to describe placements and local construction. It also adds regression coverage for Builder placement, nested Builder publication, BaseObject isolation, and curve behavior affected by the refactor.


greyltc:
Should we expect what's now in the `main` branch of bd_warehouse to work with this? Seems like my fastener's `.hole_locations` are somehow no longer being registered.

The following few bd_warehouse tests done against this PR fail now too:
```
================================= short test summary info ==================================
FAILED test_fasteners.py::test_nuts[DomedCapNut-din1587-M5-0.8] - assert not True
FAILED test_fasteners.py::test_nuts[HexNut-iso4032-M5-0.8] - assert not True
FAILED test_fasteners.py::test_nuts[HexNut-iso4033-M5-0.8] - assert not True
FAILED test_fasteners.py::test_nuts[HexNut-iso4035-M5-0.8] - assert not True
FAILED test_fasteners.py::test_nuts[UnchamferedHexagonNut-iso4036-M5-0.8] - assert not True
FAILED test_fasteners.py::test_nuts[SquareNut-din557-M5-0.8] - assert not True
FAILED test_gears.py::test_spur_gear_plan - ValueError: Edges are disconnected
FAILED test_gears.py::test_spur_gear - ValueError: Edges are disconnected
============= 8 failed, 871 passed, 1 skipped, 21 warnings in 66.23s (0:01:06) ============
```

jdegenstein:
> Should we expect what's now in the main branch of bd_warehouse to work with this?

No. https://github.com/gumyr/bd_warehouse/blob/main/pyproject.toml#L39

jdegenstein:
Looks good to me.

PR #1374, opened 2026-07-15
Add build123d-vtk package variant to publish.yml workflow, add a variant test to testing workflow

jdegenstein:
- Patch pyproject.toml during runtime of publish.yml and test.yml workflows
  - Changes to cadquery-ocp as dep, change package name to build123d-vtk
- Considered the approach of creating a `build123d-vtk/` subfolder with separate `pyproject.toml` but using `../src` is not recommended and not compatible with windows
- Manually sets setuptools_scm version of build123d-vtk variant to match build123d to override dirty git history caused by sed changes to pyproject.toml
- Both packages are built then published simultaneously on tag creation
- Will need to separately create a build123d-vtk package on PyPI

Motivation:

1. Makes it easier to install build123d alongside cadquery again because of shared `cadquery-ocp` dependency instead of clash caused by `cadquery-ocp-novtk`
2. Makes it easier to use build123d's `vtk_tools.py`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,6 +36,15 @@ jobs:
       - name: Build sdist and wheel (build123d-vtk)
         shell: bash
         run: |
+          WHEEL_FILE=$(ls wheelhouse/build123d-*.whl | head -n 1)
+
+          # Use unzip to stream the METADATA file, then grep the Version line
+          CLEAN_VERSION=$(unzip -p "$WHEEL_FILE" "*.dist-info/METADATA" | grep "^Version: " | cut -d' ' -f2)
+
+          # Force setuptools_scm to use this exact version
+          export SETUPTOOLS_SCM_PRETEND_VERSION="$CLEAN_VERSION"
+          echo "Pretending version for VTK build: $CLEAN_VERSION"
+
           # Patch pyproject.toml for the VTK variant
           sed -i 's/name = "build123d"/name = "build123d-vtk"/' pyproject.toml
           sed -i 's/cadquery-ocp-novtk/cadquery-ocp/' pyproject.toml
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,7 +20,7 @@ jobs:
           persist-credentials: false
           fetch-depth: 0 # get all history for setuptools_scm
         
-      - name: Build sdist and wheel
+      - name: Build sdist and wheel (main build123d)
         shell: bash
         run: |
           pwd
@@ -33,9 +33,21 @@ jobs:
           python3 -m pip freeze
           ls -lR
 
+      - name: Build sdist and wheel (build123d-vtk)
+        shell: bash
+        run: |
+          # Patch pyproject.toml for the VTK variant
+          sed -i 's/name = "build123d"/name = "build123d-vtk"/' pyproject.toml
+          sed -i 's/cadquery-ocp-novtk/cadquery-ocp/' pyproject.toml
+
+          # Build the new variant into the existing wheelhouse dir
+          python3 -m build --outdir wheelhouse
+
+          ls -lR wheelhouse
+
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a #v7.0.1
         with:
-          path: ./wheelhouse/build123d*.* # store the build123d wheel and sdist
+          path: ./wheelhouse/*.* # store the build123d wheel and sdist and build123d-vtk variant
       
   upload_pypi:
     needs: [build_wheel]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,19 +18,33 @@ jobs:
           windows-latest, 
           ubuntu-24.04-arm
         ]
+        vtk: [false]
         include:
           - python-version: "3.10"
             os: ubuntu-latest
+            vtk: false
+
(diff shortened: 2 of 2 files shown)


gumyr:
It will be nice to move further away from vtk. I'm not an expert is this stuff so I asked for help from CoPilot - here are its suggestions but I'm not asking for any changes:
1) Avoid in-place edits of the checked-in pyproject.toml
   - Instead of sed -i on the repo file, build the vtk variant from a temporary copy of the repo (e.g., copy the workspace to /tmp or a new directory, patch the copy, and run the build there). This prevents leaving the workspace mutated and reduces risk of surprising side effects later in the job.

2) Use a structured edit rather than a blind sed
   - The sed replacements (name and dependency) may inadvertently change other matches. Prefer a small Python script using tomllib/toml or a toml-cli tool to programmatically update the [project] name and only the specific dependency entry.

3) Confirm portability and runner assumptions
   - sed -i is GNU-sed-specific; while publish.yml likely runs on ubuntu, call out this assumption or prefer the temporary-copy approach which avoids cross-shell sed behavior differences.

4) Make METADATA extraction slightly more robust
   - unzip -p "*.dist-info/METADATA" works, but consider explicitly targeting the wheel's internal dist-info path or using a Python snippet (zipfile) to read METADATA to avoid unexpected matches if wheelhouse contains other wheels.

5) Build isolation / debugging
   - Add a short echo of the modified pyproject.toml (or its [project] name + dependencies) before building so CI logs show what was used to build the vtk wheel. Also ensure the build runs in a clean venv so the variant build won't be affected by the prior build step.

6) Publishing considerations
   - Ensure a separate PyPI project exists for build123d-vtk before the upload step (you noted this). Confirm your upload step(s) will publish both wheels and that twine/pypi credentials permit publishing multiple distributions in one job.

7) Tests
   - Ensure the corresponding test workflow installs and imports the correct wheel name and validates the vtk-specific bits (e.g., that vtk_tools.py is importable) in an isolated environment.

jdegenstein:
Thanks for the feedback, I am converting to draft while I work through some of these suggestions.

.github/workflows/publish.yml
```diff
@@ -33,9 +33,30 @@ jobs:
...
+
+          # Use unzip to stream the METADATA file, then grep the Version line
+          CLEAN_VERSION=$(unzip -p "$WHEEL_FILE" "*.dist-info/METADATA" | grep "^Version: " | cut -d' ' -f2)
+
+          # Force setuptools_scm to use this exact version
+          export SETUPTOOLS_SCM_PRETEND_VERSION="$CLEAN_VERSION"
+          echo "Pretending version for VTK build: $CLEAN_VERSION"
+
+          # Patch pyproject.toml for the VTK variant
+          sed -i 's/name = "build123d"/name = "build123d-vtk"/' pyproject.toml
```
bernhard-42, line 49: `sed -i` doesn't work on MacOS. One needs to use `sed -i.bak` which works on both, but leaves a backup behind. For a toml file a Python solution could also be considered.

.github/workflows/publish.yml
```diff
@@ -33,9 +33,30 @@ jobs:
           python3 -m pip freeze
           ls -lR
 
+      - name: Build sdist and wheel (build123d-vtk)
+        shell: bash
+        run: |
+          WHEEL_FILE=$(ls wheelhouse/build123d-*.whl | head -n 1)
+
+          # Use unzip to stream the METADATA file, then grep the Version line
+          CLEAN_VERSION=$(unzip -p "$WHEEL_FILE" "*.dist-info/METADATA" | grep "^Version: " | cut -d' ' -f2)
```
bernhard-42, line 42: Just curious, why does the non-vtk (standard) build not need the `SETUPTOOLS_SCM_PRETEND_VERSION`?
(no need for change)

bernhard-42 (commented):
Plus , I don't fully understand the upload to pypi, since I think it'll be two different pypi projects (build123d and build123d-vtk, isn'tit?

I would maybe have written an action with parameters that does all the wheel building and pypi upload and a workflow that calls this action twice with different parameters (`name` as build123d/build123-vtk, and `ocp` as cadquery_ocp_novt/cadquery_ocp).

But there are for sure more than one way to approach this.

.github/workflows/publish.yml
jdegenstein: Yeah I am going to add a python file in tools/ to do this patch instead of this sed stuff. Users will still need to clone -> patch -> install but it will be cross platform and automatic (when installing from github). When installing from PyPI it will be just `pip install build123d-vtk` of course.

.github/workflows/publish.yml
jdegenstein: > Manually sets setuptools_scm version of build123d-vtk variant to match build123d to override dirty git history caused by sed changes to pyproject.toml

^ should be answered by this explanation in the post above

jdegenstein:
>Plus , I don't fully understand the upload to pypi, since I think it'll be two different pypi projects (build123d and build123d-vtk, isn'tit? 

As far as I know PyPI upload does not depend on this at all, which is why I can simultaneously upload to multiple different PyPI projects and that is sorted out by the wheels/sdists themselves. I haven't been able to run a complete test on this yet, might be worth trying with test.pypi.org come to think of it.

.github/workflows/publish.yml
bernhard-42: thanks

Issue #1269, opened 2026-04-06
`EllipticalCenterArc` raises UnboundLocalError if only one depreciated parameter used


jwagenet:
Legacy examples which use constructor with single depreciated parameter, eg. `EllipticalCenterArc(..., end_angle=360)`, raise UnboundLocalError.

[Example](https://github.com/gumyr/build123d/blob/e8cae0660e7ec0f2cc795229966fe47e24f66fbf/docs/spitfire_wing_gordon.py)

In this case `direction` is unset when only `end_angle` is unused:
https://github.com/gumyr/build123d/blob/e8cae0660e7ec0f2cc795229966fe47e24f66fbf/src/build123d/objects_curve.py#L1126-L1140

The depreciation branch sets the flag and intermediate variable for the used parameter, but not the unused depreciated parameter:
https://github.com/gumyr/build123d/blob/e8cae0660e7ec0f2cc795229966fe47e24f66fbf/src/build123d/objects_curve.py#L1094-L1112

jdegenstein:
Closed as fixed by a1d4f223ea05a0e149e46ea2a2fd4448a9e33d68 which removed the deprecated parameters. The example was also fixed in the docs as linked above.

PR #1376, opened 2026-07-16
Test Docs Examples

jdegenstein:
- create new test file for testing examples from docs subfolder
  - skip ttt as they are tested in existing `test_examples.py`
  - skip those that depend on `bd_warehouse`
  - does not test python code embedded in .rst files (non-goal of this PR)
  - for the most part this is testing that examples can run, not for correctness
  - correctness can be tested by simply adding assert statements as I did below with spitfire_gordon.py
- fix failing tests
  - spitfire_gordon.py had two problems:
    - affected by bug in EllipticalCenterArc arc_size handling
    - affected by change to .translate where it does not reconstruct anymore, decided to switch to .move and scale about global origin. This resulted in incorrect curve placement, causing gordon surface construction to fail.
    - also added an assert statement with expected volume
  - work around bug in EllipticalCenterArc, HyperbolicCenterArc, and ParabolicCenterArc by manually setting arc_size
  
Motivation:
  
Add another guardrail to prevent the docs from not being updated when the API is. Even if this is an incomplete solution, I feel like it will help catch issues in the future.

related to this issue https://github.com/gumyr/build123d/issues/1269 but does not fix the underlying problem

### DIFF
--- a/docs/constraint_examples.py
+++ b/docs/constraint_examples.py
@@ -79,7 +79,7 @@ svg.write("assets/perpendicular_ex.svg")
 # Intersection
 #
 with BuildLine() as intersect_ex:
-    c_l1 = EllipticalCenterArc((0, 0), 1.2, 1.8, 0, 90, mode=Mode.PRIVATE)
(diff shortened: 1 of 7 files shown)


gumyr:
Thank you!

PR #1377, opened 2026-07-16
switch from codecov to covdefaults, coverage.py, diff-cover

jdegenstein:
1. This does not depend on an cloud provider like codecov did
2. covdefaults provides the default config as a plugin to coverage.py. This step is analogous to codecov/project.
3. diff-cover takes the report from step 2 and analyzes it in context of the patch. This step is analogous to codecov/patch.
4. lower project fail target to 94% which is slightly below current (94.21)
5. add html reports for coverage.py and diff-cover
6. switch to editable installs to ensure paths match correctly, this affects all workflows because it was changed in our setup workflow.

Motivation:
Eliminate constant flaky behavior in codecov resulting in failing CI especially for codecov/project being out of date.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -21,6 +21,6 @@ runs:
     - name: Install Requirements
       shell: bash
       run: |
-        uv pip install .["$OPTIONAL_DEPS"]
+        uv pip install -e .["$OPTIONAL_DEPS"]
       env:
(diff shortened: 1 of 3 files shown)


jdegenstein:
Here's an example screenshot of the coverage.py html report. The diff-cover html report is uninteresting for this PR because there were no changes to relevant files it checks.

image

This is ready for review @gumyr

PR #1380, opened 2026-07-21
Load .ttf files that actually contain a TrueType Collection

### LINKED ISSUE #1367: TTLibFileIsCollectionError when importing .ttf fonts containing a TrueType Collection on Windows
## Description

`build123d` still fails to import on Windows when a font file has a `.ttf` extension but actually contains a TrueType Collection.

This is a different scenario from #1333, which addressed `.TTC` files with uppercase extensions.

During `FontManager.register_system_fonts()`, `register_font()` decides whether to use `TTFont` or `TTCollection` based solely on the filename extension:

```python
_, ext = os.path.splitext(path)

if ext.strip(".").lower() == "ttc":
    fonts = ttCollection.TTCollection(path)
else:
    fonts = [TTFont(path)]
```

If a font file has a `.ttf` extension but internally contains a TrueType Collection, `TTFont(path)` raises:

```text
fontTools.ttLib.TTLibFileIsCollectionError
```

which aborts the entire `build123d` import.

## Example

On my Windows installation, the following font reproduces the issue:

```text
C:\Windows\Fonts\arplukai.ttf
```

Testing it directly with `fontTools` shows:

```python
from fontTools.ttLib import TTFont

TTFont(r"C:\Windows\Fonts\arplukai.ttf")
```

raises

```text
TTLibFileIsCollectionError
```

while

```python
from fontTools.ttLib import TTCollection

collection = TTCollection(r"C:\Windows\Fonts\arplukai.ttf")
print(len(collection.fonts))
```

loads successfully and reports **4 fonts**.

This indicates that the file is a valid TrueType Collection despite having a `.ttf` extension.

## Expected behavior

`build123d` should not rely exclusively on the filename extension to determine whether a file is a single font or a font collection.

One possible approach would be:

* attempt to load the file using `TTFont()`;
* if `TTLibFileIsCollectionError` is raised, fall back to `TTCollection()`.

This would make the font scanner more robust against font files whose internal format does not match their extension.

## Environment

* build123d 0.11.1
* fontTools 4.63.0
* Python 3.13.3
* Windows 11

cycsmail:
Some system fonts ship a TrueType Collection under a `.ttf` extension (Windows `arplukai.ttf` is the one in the issue), so `register_font` can't decide by extension alone. `TTFont(path)` raises `TTLibFileIsCollectionError` on those, which bubbles all the way out. I catch it and fall back to loading the file as a `TTCollection`, same as the `.ttc` branch does.

Added a test that saves a collection with a `.ttf` name and registers it, it fails without the change. `tests/test_text.py` passes (9 tests).

Fixes #1367.

### DIFF
--- a/src/build123d/text.py
+++ b/src/build123d/text.py
@@ -16,7 +16,7 @@ import platform
 import sys
 from dataclasses import dataclass
 
-from fontTools.ttLib import TTFont, ttCollection  # type: ignore
+from fontTools.ttLib import TTFont, TTLibFileIsCollectionError, ttCollection  # type: ignore
 from OCP.Font import (
     Font_FA_Bold,
     Font_FA_BoldItalic,
@@ -137,7 +137,12 @@ class FontManager:
         if ext.strip(".").lower() == "ttc":  # pragma: no cover
             fonts = ttCollection.TTCollection(path)
         else:
-            fonts = [TTFont(path)]
+            try:
+                fonts = [TTFont(path)]
+            except TTLibFileIsCollectionError:
+                # Some files carry a .ttf extension but actually contain a
+                # TrueType Collection; fall back to loading them as one.
+                fonts = ttCollection.TTCollection(path)
 
         font_faces = []
         for font in fonts:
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #1381, opened 2026-07-21
Sheet metal POC: BuildSheet builder with flange and hem operations

GabrielJMS:
## Summary

This is a **proof-of-concept** for construction-side sheet metal support in
build123d: a `BuildSheet` builder context plus `flange` and `hem`
operations, following the direction gumyr sketched in #305.

`BuildSheet` treats closed sketch regions as an auto-padded base sheet of
constant `thickness`; `flange` folds a wall up from selected sheet edges
with a cylindrical bend, and `hem` folds an edge back onto itself (flat,
open, teardrop, or rolled), matching the corresponding operations in the
FreeCAD SheetMetal workbench.

```python
from build123d import *

with BuildSheet(thickness=1, bend_radius=2) as box:
    with BuildSketch():
        Rectangle(100, 60)

    bottom = box.faces().sort_by(Axis.Z)[0]
    flange(
        bottom.edges().filter_by(GeomType.LINE),
        length=20,
        gap1=3.1,
        gap2=3.1,
    )

    long_walls = box.faces().filter_by(Axis.Y).sort_by(Axis.Y)
    rims = [
        long_walls[0].edges().sort_by(Axis.Z)[-1],
        long_walls[-1].edges().sort_by(Axis.Z)[-1],
    ]
    hem(rims, hem_type=HemType.OPEN, width=6, opening=2)
```

`box.sheet` is a plain `Part` afterwards — export it, measure it, combine
it with the rest of build123d as usual.

## The modeling technique

Each bend is built as two solids and fused onto the sheet without face
unification:

1. A **thickness cross-section rectangle** (the bend's radial profile,
   `p0, p1, p1 + thk_dir*thickness, p0 + thk_dir*thickness`) is **revolved**
   about the bend axis by the bend angle — this becomes the bend's
   cylindrical surface plus two flat "fan"-shaped end caps closing the
   revolve.
2. A flat wall rectangle is extruded by `thickness` and rotated to the
   bend's end angle to form the flange/hem leg beyond the bend.

The two solids are fused (and, for inward bend positions, a slab is first
cut from the sheet to make room) with `SkipClean` active throughout, so
the `ShapeUpgrade_UnifySameDomain` pass that `fuse`/`cut` normally run is
skipped, and the bend's cylindrical face and its two fan end caps survive
as distinct faces in the result instead of being merged into the
adjoining flats.

This matters because a **future unfold tool** needs exactly this
topology to find bends: it can walk the part's faces, recognize a
cylindrical face bounded by two fan-shaped end caps as one bend, and use
its radius/angle (plus the builder's stored `k_factor`) to compute the
flat-pattern length. Once `clean()` is called the cylindrical and flat
faces merge and that information is gone — which is why `BuildSheet`
parts are deliberately never auto-cleaned, and both `flange`/`hem`
default `clean=False` with an explicit warning against calling
`clean()` on the result.

Credit: this modeling technique (bend-as-revolved-cross-section, fan-face
preservation for unfolding, the hem geometry equations) is translated
from the FreeCAD SheetMetal workbench
(https://github.com/shaise/FreeCAD_SheetMetal), particularly the `smBend`
engine in `SheetMetalCmd.py` and the hem generators in `SheetMetalHem.py`. While porting `generateTeardropHem`'s
`opening == 2 * radius` branch, this port also fixes an upstream bug: the
FreeCAD code calls the open-hem helper with 3 arguments where it takes 4,
which would raise a `TypeError` if that branch were ever hit. Our port
(`_hem_parameters`, `HemType.TEARDROP` / `opening == 2 * radius` case)
calls it correctly.

## Relationship to #305

This POC implements the **construction API** direction from gumyr's
ideation comment on #305 (`BuildSheet` + operations), scoped down to a
minimal but real vertical slice:

- `BuildSheet` (the base sheet builder, analogous to `BuildPart`)
- `flange` (≈ the issue's "Flange"/"Bend" operation)
- `hem` (≈ the issue's "Hem" operation, all 4 FreeCAD hem styles)

`unfold` is explicitly **out of scope** for this POC — the topology
choices above are made specifically so that a later unfold tool has what
it needs, but no flattening logic is implemented here.

## Out of scope / contribution roadmap

Deliberately left for follow-up PRs (not attempted here):

- Bend between two sketch regions (bend without an existing solid edge)
- `BendLine`
- `Tab`
- Bend reliefs (corner reliefs where flanges meet)
- Automatic / manual miters
- Perforation
- `base_sheet()` for open profiles built from a `BuildLine` (as opposed to
  the existing `make_brake_formed`), including auto-filleting of corners
- `smCreateBaseShape`-style parametric starter shapes
- `Material` integration (see open question on `k_factor` below)
- Unfold / flat-pattern generation

## Open questions for maintainers

- **Operation functions vs. capitalized operation classes** — `flange`
  and `hem` are plain functions (like `fillet`/`chamfer`), consistent with
  existing `operations_*` modules. Given sheet metal operations carry
  more per-call state (bend frame, gap trimming, hem parameter solving),
  is that still the right shape, or should these be objects/context
  managers akin to `GridLocations`?
- **Dedicated `Sheet` composite type vs. plain `Part`** — `BuildSheet.sheet`
  is currently a plain `Part`. Would a `Sheet` subclass (carrying
  `thickness`/`bend_radius`/`k_factor` with the geometry, so it survives
  outside the builder context) be preferable, e.g. for a future `unfold()`
  free function to consume?
- **`Material` integration for `k_factor`** — `k_factor` is stored on
  `BuildSheet` but never consumed (needed once unfold lands). Should it
  instead live on/derive from a `Material` object once that concept
  exists in build123d, rather than being a raw builder parameter?
- **Defaulting `make_brake_formed`'s `thickness` from `BuildSheet`** —
  `make_brake_formed` currently requires an explicit `thickness` argument
  even inside a `BuildSheet` context that already has one. Should it fall
  back to `context.thickness` when omitted, the way `flange`/`hem` do?
- (Minor, noted for completeness) `hem`'s `roll_angle` for `HemType.ROLLED`
  defaults to `None` → the physical maximum roll angle (matching FreeCAD's
  behavior), rather than a fixed `270`. This seemed like better default
  behavior than a magic constant, but flagging it as a deviation from the
  original brainstorm in case a fixed default is preferred for
  predictability.

---


gumyr:
I've been working on another prototype based on this - see https://github.com/gumyr/build123d/pull/1434

PR #1388, opened 2026-07-29
Add persistent Onshape-style mates and AP242 kinematics

Divide-By-0:
## Summary

Adds a persistent assembly constraint system alongside the existing one-shot Joint positioning API, then makes it portable through AP242 STEP.

- introduces `Assembly`, an exportable `Compound` that retains mates, relations, grounding, and solve diagnostics
- uses `MateConnector` as a `RigidJoint`-compatible local coordinate frame (Z primary, X secondary)
- solves all active constraints simultaneously with explicit/implicit grounding, underconstraint regularization, limits, driven values, suppression, and clear inconsistency errors
- preserves mate design intent and connector rebinding through deep copies

## Onshape-compatible feature surface

Connector/entity mates:

- Fastened
- Revolute
- Slider
- Planar
- Cylindrical
- Pin Slot
- Ball, including the single conical swing limit exposed by Onshape as `limitEulerConeAngleMax`
- Parallel
- Tangent, including face/edge/vertex selections, swept/analytic-face validation, flip behavior, and tangent propagation
- Width, with one/two tab connectors and two width connectors
- Group, capturing component-origin relationships without geometry dependencies

Mate relations:

- Gear
- Rack and Pinion
- Screw
- Linear

Raw Onshape API parameter objects can be attached to every mate/relation with `onshape_parameters`; order, parameter IDs, expressions, values, and extra fields are copied and retained exactly.

## AP242 STEP kinematics

`export_step()` now detects an `Assembly` with mates and writes AP242 by default. The file contains:

- standard `KINEMATIC_LINK`, `KINEMATIC_JOINT`, pair, range, mechanism, state, and product/geometry association entities
- native Fastened, Revolute, Slider, Planar, Cylindrical, Ball, and generic low-order Pin Slot/Parallel pairs
- offsets plus primary/secondary alignment incorporated into the native pair frames
- Group expanded into fully constrained pairs
- current pair values and finite/unbounded ranges in the appropriate STEP units
- a deterministic, versioned, compressed build123d payload in a legal Part 21 comment for exact round trips

`import_step_assembly()` restores the geometry, mate objects, raw Onshape fields, limits, values, suppression, connectors, offsets, selected entities, Group transforms, relations, ratio/travel/reverse settings, and relation phase. `write_kinematics=False` keeps the prior geometry-only behavior.

### Explicit interoperability boundary

No build123d/Onshape information is discarded. Some feature intent has no faithful one-to-one native AP242 entity, so it is retained in the versioned payload:

- Onshape Ball uses one cone-angle limit; AP242 spherical ranges use independent angular coordinates
- Tangent propagation/selections and Width centering are not single AP242 low-order pairs
- Pin Slot and Parallel have motion-compatible generic low-order pairs but no native feature identity
- Gear lacks the two absolute pitch radii required by AP242 when only an Onshape ratio is available
- mate-coordinate Gear/Rack-and-Pinion/Screw/Linear relations do not always describe one adjacent-link AP242 motion-coupling pair

Generic AP242 readers therefore get the faithful standard subset; build123d gets a lossless round trip including every extension detail.

## Validation

- full suite: `2202 passed, 2 skipped, 64 subtests passed`
- focused mate/export/import suite: `120 passed`
- new AP242 tests cover all connector pair families, Ball cone semantics, offset/alignment frames, Group expansion, raw API payloads, relation round-trip, byte streams, opt-out, and OCCT STEP parsing
- touched-module pylint: `10.00/10`, no errors/fatals
- Black and `git diff --check`: clean

## Documentation

Updates the persistent-mates guide with AP242 behavior, lossless round-trip usage, and the explicit standards boundary.

### DIFF
--- a/docs/assemblies.rst
+++ b/docs/assemblies.rst
@@ -8,6 +8,9 @@ Most CAD designs consist of more than one part which are naturally arranged in
 some type of assembly. Once parts have been assembled in a :class:`~topology.Compound` object
 they can be treated as a unit - i.e. :meth:`~topology.Shape.moved` or exported.
 
+For assemblies that must preserve and re-solve their mechanical constraints,
+see :ref:`assembly_mates`.
+
 To create an assembly in build123d, one needs to
 create a tree of parts by simply assigning either a :class:`~topology.Compound` object's ``parent`` or
 ``children`` attributes. To illustrate the process, we'll extend the
--- a/docs/assembly_mates.rst
+++ b/docs/assembly_mates.rst
@@ -1,0 +1,253 @@
+.. _assembly_mates:
+
+#########################
+Persistent Assembly Mates
+#########################
+
+Build123d joints position two parts when ``connect_to`` is called.  An
+:class:`~assembly.Assembly` instead retains its mates and mate relations as
+design intent.  Calling :meth:`~assembly.Assembly.solve` recalculates every
+active constraint simultaneously, so moving a fixed component, changing a
+mate value, or suppressing a feature updates the complete assembly.
+
+Mate connectors use the Onshape coordinate convention: Z is the primary axis
+and X is the secondary axis.  :class:`~assembly.MateConnector` derives from
+:class:`~joints.RigidJoint`, so existing rigid joints can be used directly as
+mate connectors and retain the usual part-local storage and copy behavior.
+
+Quick Start
+***********
+
+.. code-block:: python
+
+    from build123d import *
+
+    frame = Box(40, 20, 4)
+    frame.label = "frame"
+    arm = Pos(0, 0, 20) * Box(4, 4, 30)
+    arm.label = "arm"
+
+    frame_axis = MateConnector("frame axis", frame, Location((0, 0, 4)))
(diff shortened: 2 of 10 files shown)


gumyr:
Thank you for taking the time to put this PR together. I'm very interested in the ability to save joints into STEP files, and I appreciate the effort that went into exploring this capability.

That said, this PR introduces a major new architectural subsystem that I'm not ready to add to build123d at this point. It adds a substantial amount of new infrastructure, and the implementation doesn't appear to be tightly integrated with the existing build123d architecture. As a result, I'm unlikely to merge it in its current form.

I'm also hesitant to directly adopt another project's joint system without first considering what best fits build123d's long-term design goals. When joint persistence is added, I'd prefer to start with a discussion of the architecture and API before implementing it.

For changes of this scope, I'd strongly encourage opening an issue or discussion first so we can agree on the overall design before anyone invests significant effort in an implementation.

Even though I won't be merging this PR as-is, it provides a useful reference that I expect to revisit when this area of the project is explored in the future. Thank you again for the contribution.

PR #1389, opened 2026-08-02
Raise error for missing STL export directory

### LINKED ISSUE #1379: Exporting silently fails if destination directory does not exist
`export_stl(Box(10, 10, 10), "nonsense/path")` should throw an error rather than silently failing with no indication why

kforris:
## Summary

- validate the STL destination parent directory before meshing
- raise `FileNotFoundError` when the directory is missing
- add a focused regression test for the reported case

Closes #1379

## Validation

- `.venv/bin/python -m pytest tests/test_exporters3d.py -q` — 34 passed
- diff coverage — 100% (4/4 changed source lines)
- Black — changed source file and test hunk clean
- pylint — 10.00/10 for `src/build123d/exporters3d.py`
- mypy — no issues in the changed source and test files
- full `pytest -n auto --benchmark-disable` — 2151 passed, 2 skipped, with one unrelated local macOS font-database failure in `TestFontManager.test_register_system_fonts`; the failing test persists in isolation

### DIFF
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -469,7 +469,14 @@ def export_stl(
 
     Returns:
         bool: Success
+
+    Raises:
+        FileNotFoundError: The destination directory does not exist.
     """
+    output_path = Path(fsdecode(file_path))
+    if not output_path.parent.exists():
+        raise FileNotFoundError(output_path.parent)
+
     mesh = BRepMesh_IncrementalMesh(
         to_export.wrapped, tolerance, True, angular_tolerance, True
     )
@@ -478,7 +485,7 @@ def export_stl(
     writer = StlAPI_Writer()
 
     writer.ASCIIMode = ascii_format
-    return writer.Write(to_export.wrapped, fsdecode(file_path))
+    return writer.Write(to_export.wrapped, str(output_path))
 
 
 def export_to_pcbway(
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -474,7 +474,7 @@ def export_stl(
         FileNotFoundError: The destination directory does not exist.
     """
     output_path = Path(fsdecode(file_path))
-    if not output_path.parent.exists():
+    if not output_path.parent.is_dir():
         raise FileNotFoundError(output_path.parent)
 
     mesh = BRepMesh_IncrementalMesh(
--- a/tests/test_exporters3d.py
+++ b/tests/test_exporters3d.py
@@ -406,6 +406,14 @@ def test_export_stl_missing_destination_directory(tmp_path):
         export_stl(Box(1, 1, 1), output_path)
 
 
+def test_export_stl_destination_parent_is_file(tmp_path):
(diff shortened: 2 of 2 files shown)


tests/test_exporters3d.py
```diff
@@ -399,6 +399,13 @@ def test_pathlike_exporters(tmp_path, format, exporter):
...
 
 
+def test_export_stl_missing_destination_directory(tmp_path):
+    output_path = tmp_path / "missing" / "box.stl"
+
+    with pytest.raises(FileNotFoundError, match="missing"):
+        export_stl(Box(1, 1, 1), output_path)
+
+
 @pytest.mark.parametrize("exporter", (export_step, export_brep))
```
BlueDrink9, line 417: Have we checked whether the other exporters also have this issue and should also be tested?

src/build123d/exporters3d.py
```diff
@@ -469,7 +469,14 @@ def export_stl(
 
     Returns:
         bool: Success
+
+    Raises:
+        FileNotFoundError: The destination directory does not exist.
     """
+    output_path = Path(fsdecode(file_path))
+    if not output_path.parent.exists():
```
gumyr: How about adding a `output_path.parent.is_dir()` check in there - not sure what would happen if there was a file with that name?

gumyr (changes requested):

kforris:
Addressed both review points in f84e92bb:

- Switched the STL parent check to Path.is_dir() and added a regression for the case where the parent path exists as a file.
- Audited the other 3D file exporters with a missing parent directory: export_brep returns False, export_gltf emits the native writer error and returns False, and export_step raises RuntimeError. I kept this PR scoped to STL because the issue reproduction and requested code change are STL-specific; I can follow up separately if you want consistent behavior across BREP and glTF too.

Validation: the focused exporter suite passes 35 tests; source Black, Pylint 10.00/10, Mypy, git diff --check, and changed-line coverage at 100% all pass. The full parallel suite reached 2,152 passed and 2 skipped with one unrelated glTF PBR test failure; that exact test passed when rerun alone.

kforris:
Thanks again for the review. The requested `is_dir()` guard and regression are on `f84e92bb`, and the refreshed GitHub matrix is green (15 passed, with 2 publish-only jobs skipped). When convenient, could you take another look? I am happy to adjust if you would prefer broader exporter consistency to be handled in this PR rather than separately.

tests/test_exporters3d.py
kforris: Checked the other 3D exporters with the same missing-parent case: export_brep returns False, export_gltf reports the native writer error and returns False, and export_step raises RuntimeError. I kept this fix scoped to STL because #1379 and the failing STL writer behavior are STL-specific; happy to follow up separately if consistent behavior is desired.

src/build123d/exporters3d.py
kforris: Addressed in f84e92bb: the guard now uses output_path.parent.is_dir(), and the added regression covers a parent path that exists as a file.

gumyr:
Thank you!

PR #1396, opened 2026-08-06
Accept iterators in Locations

### LINKED ISSUE #1359: Locations rejects generators/map despite docstring saying "iterable of same"
### Description

`Locations`' docstring documents `pts` as `Union[...] or iterable of same`, but passing a generator/iterator (e.g. the result of `map`) raises `ValueError`. In practice only `list`, `tuple`, `filter`, and `set` are flattened — other iterables are not.

### Reproduction

```python
from build123d import Locations, BuildSketch, Rectangle, Plane

pts = map(lambda p: (p, 0), [10, 20, 30])  # an iterable of VectorLike

with BuildSketch(Plane.XY):
    with Locations(pts):        # docstring: "iterable of same" should be accepted
        Rectangle(1, 1)
```

Result:

```
ValueError: Locations doesn't accept type <class 'map'>
```

Splatting or wrapping in a list works fine:

```python
Locations(*pts)        # ok
Locations(list(pts))   # ok
```

### Cause

`Locations.__init__` runs its args through `flatten_sequence` (`build_common.py`), which only recurses into a hard-coded set of container types:

```python
if isinstance(item, (list, tuple, filter, set)) and not _is_point(item):
    flat_list.extend(flatten_sequence(*item))
else:
    flat_list.append(item)
```

A `map`/generator isn't in that set, so it's appended as a single opaque item and then rejected by the type checks in `Locations.__init__`.

The `# Note:` there explains why a blanket `isinstance(item, Iterable)` check isn't used (`Vector`/`Vertex` are iterable and would be broken into floats), so the allow-list is intentional — but it omits the most common lazy iterables (`map`, generators), which contradicts the documented "iterable of same".

### Suggested fix

Either:

- update the docstring to name the actually-supported container types (`list`, `tuple`, `filter`, `set`), or
- honor the documented contract by materializing generic iterators while still excluding the point-like iterables (`Vector`, `Vertex`, etc.).

### Workaround

`Locations(*gen)` or `Locations(list(gen))`.

### Environment

- build123d 0.11.0
- Python 3.14
- macOS

kforris:
## Summary

- flatten iterator inputs such as `map`, generators, `filter`, and list iterators
- keep the narrower iterator check instead of accepting every `Iterable`, so iterable geometry types such as `Vector` and `Vertex` remain atomic
- add regression coverage for both `flatten_sequence` and `Locations`

## Validation

- `uv run --no-sync python -m pytest tests/test_build_common.py::TestFlattenSequence tests/test_build_common.py::TestLocations -q` (24 passed)
- `uv run --no-sync python -m pytest -n auto` (2182 passed, 2 skipped)
- Black checks passed for the changed ranges
- `uv run --no-sync pylint src/build123d/build_common.py` (10.00/10)
- `uv run --no-sync mypy src/build123d/build_common.py` (no issues)

Closes #1359

### DIFF
--- a/src/build123d/build_common.py
+++ b/src/build123d/build_common.py
@@ -134,8 +134,8 @@ def flatten_sequence(*obj: T) -> ShapeList[Any]:
     flat_list: ShapeList[Any] = ShapeList()
     for item in obj:
         # Note: an Iterable can't be used here as it will match with Vector & Vertex
-        # and break them into a list of floats.
-        if isinstance(item, (list, tuple, filter, set)) and not _is_point(item):
+        # and break them into a list of floats. Iterators are safe to consume.
+        if isinstance(item, (list, tuple, set, Iterator)) and not _is_point(item):
             flat_list.extend(flatten_sequence(*item))
         else:
             flat_list.append(item)
--- a/tests/test_build_common.py
+++ b/tests/test_build_common.py
@@ -87,6 +87,12 @@ class TestFlattenSequence(unittest.TestCase):
             flatten_sequence("a", ("b", "c", "d"), "e"), ["a", "b", "c", "d", "e"]
         )
 
+    def test_iterators(self):
+        self.assertListEqual(flatten_sequence(map(str, range(3))), ["0", "1", "2"])
+        self.assertListEqual(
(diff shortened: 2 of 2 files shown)


gumyr:
Thank you!

PR #1397, opened 2026-08-06
Preserve labels for 3MF assembly children

### LINKED ISSUE #1375: Exporting 3MF with object labels
Hello,

I'm trying generate a 3MF with meaningful object names, using something along the lines of:

```python
from build123d import Mesher, Compound, Sphere, Pos

a = Pos(20, 0, 0) * Sphere(5)
b = Pos(40, 0, 0) * Sphere(5)
c = Pos(60, 0, 0) * Sphere(5)

a.label, b.label, c.label = "A", "B", "C"

assembly = Compound(children=[a, b, c])
exporter = Mesher()
exporter.add_shape(assembly) # or just [a, b, c]
exporter.write("example.3mf")
```

I was hoping to end up with a 3MF with three objects named A, B, C: 

```xml
<!-- unzip -p example.3mf 3D/3dmodel.model | grep -F '<object' | grep -F 'name=' | sed 's,^\s*,,'  -->
<object id="1" name="A" type="model" p:UUID="e5e9352e-5cfc-407a-a81c-6008f49d0c20">
<object id="3" name="B" type="model" p:UUID="3be08777-9022-4ca1-8420-57dc00ab39d0">
<object id="5" name="C" type="model" p:UUID="1ec96b70-d595-45c0-a71a-1429203f538d">
```

However, running this code results in no named objects (i.e. there is no `name` attribute on any of the `object` tags).

From a cursory read of the code, it looks like `Mesher.add_shape()` always ends up working on sub-shapes (e.g. `Solid`s), since everything you pass to it is both a sub-type of`Compound` and iterable. Since the sub-shape doesn't inherit the label of its parent, we don't get any labels in the resulting 3MF.

https://github.com/gumyr/build123d/blob/72cecc0f8950507e6cf4d150b0413ca4c7561d6d/src/build123d/mesher.py#L396-L403

It's worth noting that `export_step()` handles this rather well. Perhaps 3MF (and STL) can use the same means of determining the appropriate labels as the one in `_create_xde()`:

https://github.com/gumyr/build123d/blob/72cecc0f8950507e6cf4d150b0413ca4c7561d6d/src/build123d/exporters3d.py#L206-L211

Kind regards,
Georgi

kforris:
## Summary

- treat a single `Shape` as a shape before considering its iterable behavior, so labeled Part objects are not prematurely flattened
- export semantic assembly children directly and preserve their labels
- retain the existing behavior of splitting an unlabeled topological `Compound` into separate meshes
- add 3MF write/read regression coverage for both a list of labeled Part objects and `Compound(children=...)`

## Validation

- reproduced the issue before the change: both input forms produced three unnamed meshes
- verified the original labeled-sphere assembly now produces `A`, `B`, and `C`
- `uv run --no-sync python -m pytest tests/test_mesher.py -q` (21 passed, 5 subtests passed)
- `uv run --no-sync python -m pytest -n auto` (2181 passed, 2 skipped)
- Black checks passed for the changed ranges
- `uv run --no-sync pylint src/build123d/mesher.py` (10.00/10)
- `uv run --no-sync mypy src/build123d/mesher.py` (no issues)

Closes #1375

### DIFF
--- a/src/build123d/mesher.py
+++ b/src/build123d/mesher.py
@@ -393,10 +393,16 @@ class Mesher:
             Warning: Degenerate shape skipped
             Warning: 3mf mesh is not manifold
         """
-        shapes = []
-        for input_shape in shape if isinstance(shape, Iterable) else [shape]:
+        shapes: list[Shape] = []
+        input_shapes = [shape] if isinstance(shape, Shape) else shape
+        for input_shape in input_shapes:
             if isinstance(input_shape, Compound):
-                shapes.extend(list(input_shape))
+                if input_shape.children:
+                    shapes.extend(input_shape.children)
+                elif input_shape.label:
+                    shapes.append(input_shape)
+                else:
+                    shapes.extend(list(input_shape))
             else:
                 shapes.append(input_shape)
 
--- a/tests/test_mesher.py
+++ b/tests/test_mesher.py
@@ -179,6 +179,33 @@ class TestAddShape(DirectApiTestCase):
         shapes = importer.read(filename)
         self.assertEqual(importer.mesh_count, 2)
 
+    def test_add_labeled_compounds(self):
+        labels = ["A", "B", "C"]
+        for use_assembly in (False, True):
+            with self.subTest(use_assembly=use_assembly):
+                parts = [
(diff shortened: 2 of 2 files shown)


gumyr:
Thank you!

PR #1398, opened 2026-08-06
Add Superellipse to objects_sketch.

Silcantar:
A [superellipse](https://en.wikipedia.org/wiki/Superellipse) is a mathematical generalization of a normal ellipse that can be used to generate star-like shapes and approximate rectangles with rounded corners. The latter includes the "squircle", which is [used in industrial design](https://99percentinvisible.org/article/circling-square-designing-squircles-instead-rounded-rectangles/) to create objects with smooth shadow gradients.


gumyr:
I find the generation of a single spline when order is 1 or less problematic as there should be four vertices not a continuous curve. How about something like this?
```py
        points: list[VectorLike] = []
        for i, t in enumerate(linspace(0, pi / 2, point_count)):
            # To avoid gaps between the edges ensure the ends are on the axis
            if i == 0:
                x, y = width / 2, 0.0
            elif i == point_count - 1:
                x, y = 0.0, height / 2
            else:
                cos_of_t, sin_of_t = cos(t), sin(t)
                x = abs(cos_of_t) ** (2 / order) * width * copysign(1, cos_of_t) / 2
                y = abs(sin_of_t) ** (2 / order) * height * copysign(1, sin_of_t) / 2
            points.append(Vector(x, y))

        top_right_edge = Edge.make_spline(points)

        perimeter_edges = [
            top_right_edge,
            top_left_edge := top_right_edge.mirror(Plane.YZ).reversed(),
            top_left_edge.mirror(Plane.XZ).reversed(),
            top_right_edge.mirror(Plane.XZ).reversed(),
        ]
        perimeter = Wire(perimeter_edges)
        face = Face(perimeter)
        super().__init__(face, rotation, align, mode)
```
with the vertices as shown here:
image

gumyr:
I'm thinking about the order 1 case; maybe splines shouldn't be used as they generally are much more problematic when surfaces are created from them. A simple `Polyline((width/2,0),(height/2,0),(-width/2,0),(-height/2), close=True)` would create the diamond shape with straight lines.

Silcantar:
Very good point. With the single spline the the shape somehow *looks* fine for n < 1 even when zooming way in, but I'm sure it's a very weird unstable spline. For n = 1 the corners are definitely rounded over.

I'll start implementing and testing your suggestions.

Silcantar:
- All cases of superellipse (order < 1, order == 1, 1 < order < 2, order == 2, order > 2) now use 4 mirrored segments as suggested by @gumyr.
- Added special cases for order == 1 (rhombus) and order == 2 (ellipse) that produce the exact shape rather than a spline approximation.
- Added tests for all cases.

gumyr:
The implementation looks good but no one will know about it unless it's in the documentation: https://build123d.readthedocs.io/en/latest/objects.html#id1. As this object generates a wide variety of shapes it might be best to generate a number of them with different order on top of each other for the picture.

Silcantar:
Of course. Would ArrowHead be a good example of what you have in mind?

gumyr:
Yes, that would be great.

jdegenstein:
Also reminder to add `Superellipse` to https://build123d.readthedocs.io/en/latest/cheat_sheet.html

jdegenstein:
Looks like you missed coverage for this line:

```py
  545         align: Align | tuple[Align, Align] | None = (Align.CENTER, Align.CENTER),
  546         mode: Mode = Mode.ADD,
  547     ):
  548         if order <= 0:
! 549             raise ValueError("Order must be greater than zero.")
  550         self.width = width
  551         self.height_ = height
  552         self.order = order
  553         self.point_count = point_count
```

Silcantar:
Thanks y'all. Let me know if I missed anything else.

gumyr:
I don't see the svg file for the docs, am I missing something?

Silcantar:
*.svg is in the gitignore, so I assumed that the images were generated by GitHub Actions or similar. Is that not the case?

gumyr:
No, the images are not automagically generated for you, use git add -f my.svg. A recent PR for autogeneration showed just how complex that is.

gumyr:
Thank you!

PR #1399, opened 2026-08-07
Warn when Plane ignores origin with a Face

### LINKED ISSUE #1371: Warn when passing both face and origin into Plane constructor
I wanted to place a plane on a face and move it to one of that faces vertices (i.e., specify the origin). My intuition told me to try this:

```python
x = Box(1,1,1)
f = x.faces().sort_by(Axis.X)[1]
v = x.vertices().sort_by(Axis.X)[1]
Plane(f, origin=v)
> Plane((-0.5, 0, 0), (0, 0, 1), (-1, 0, 0))
v
> Vertex(-0.5, -0.5, -0.5)
```

This places the plane at the centre of the face, not at the vertex. The origin parameter is ignored.

Ideally, both of these parameters would be used, keeping the orientation as if just face had been passed in, while keeping the origin of the vertex, to give the equivalent behaviour of 
```python
Plane(f).shift_origin(v)
```

However, in the absence of that behavior, I believe a warning should be emitted about the dropped parameter.

kforris:
## Summary

- warn when `origin` is ignored while constructing a `Plane` from a `Face`
- preserve the existing face-center behavior
- cover positional and keyword `face` construction paths

## Validation

- `python -m pytest tests/test_direct_api/test_plane.py -q`
- `python -m pytest -n auto --benchmark-disable`
- `pylint --rcfile=.pylintrc --fail-under=9.5 --fail-on=E,F src/build123d`
- `mypy --config-file mypy.ini src/build123d`

Closes #1371

### DIFF
--- a/src/build123d/geometry.py
+++ b/src/build123d/geometry.py
@@ -2901,6 +2901,11 @@ class Plane(metaclass=PlaneMeta):
             surface = BRep_Tool.Surface_s(arg_face.wrapped)
             if not arg_face.is_planar:
                 raise ValueError("Planes can only be created from planar faces")
+            if arg_origin is not None:
+                warnings.warn(
+                    "The origin parameter is ignored when creating a Plane from a Face",
+                    stacklevel=2,
+                )
             properties = GProp_GProps()
             BRepGProp.SurfaceProperties_s(arg_face.wrapped, properties)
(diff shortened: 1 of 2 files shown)


gumyr:
Thank you!

PR #1400, opened 2026-08-09
docs: clarify constrained geometry terminology

### LINKED ISSUE #1382: docs: review use of terminology "contact" in docs/docstrings of ConstrainedArcs and ConstrainedLines
The docs and docstrings for ConstrainedArcs and ConstrainedLines make repeated references to tangent/contact objects or entities. To me the terminology "contact" is confusing and should be dropped where it is accurately described by tangent. As far as I can tell this is true for 4 signatures of `ConstrainedArcs ` excluding `Signature D: One constraint + fixed center` which appears to support using a point as tangency_one. 

I may be missing some nuance here though which is why I am raising this issue to document this.

ConstrainedLines `Signature B: One constraint + through point` and `Signature C: One constraint + fixed orientation` both support contact objects without tangency.

kforris:
## Summary

Clarify the constraint terminology used by `ConstrainedArcs`, `ConstrainedLines`, and their direct-API counterparts.

The documentation now distinguishes:

- tangency to curve inputs;
- incidence through point inputs;
- center-locus and fixed-orientation constraints.

This avoids describing every supported overload as a generic tangent/contact problem while preserving the point-input cases that are not tangency constraints.

Closes #1382.

## Root cause

The overloads accept both curve and point inputs, but the tutorial and docstrings used umbrella phrases such as tangent/contact and contacted/touched. Those phrases blurred two different geometric conditions. Replacing contact with tangent everywhere would also be inaccurate because several arc overloads accept point-incidence targets.

The patch is limited to terminology in the public tutorial, public constructor docstrings, direct API docstrings, and matching internal solver docstrings. Solver behavior, signatures, dependencies, and generated geometry are unchanged.

## Scope and risk

- Risk: low; documentation and docstrings only.
- No API or runtime behavior changes.
- No dependency, solver, qualification, or selection changes.
- No broad documentation rewrite or unrelated formatting.

## Reproduction and evidence

Before the patch, the affected files repeatedly used `tangent/contact`, `tangent/contacting`, and `contacted/touched` across overloads that support both curves and points. A focused post-change search finds none of those ambiguous phrases in the four affected files.

## Validation

- `git diff --check`: passed.
- Focused constrained-geometry tests: 58 passed.
- Full suite: 2,209 passed, 2 skipped, 1 unrelated local system-font registry failure. The failing test was rerun alone and reproduced without involving any changed file.
- `mypy` on the three changed Python modules: passed.
- `pylint` on the three changed Python modules: no errors; 9.97/10 with existing warnings outside the changed lines.
- Sphinx HTML build: passed; existing warnings remain outside the changed lines.
- `black --check`: the current upstream `objects_curve.py` would be reformatted in unrelated sections, and the local Python 3.11 runner also warns about the configured Python 3.14 target. I left those unrelated formatting differences untouched.

## Documentation impact

Updated the tutorial and API docstrings because this issue is specifically a terminology mismatch in public documentation. No changelog entry is needed because behavior and API are unchanged.

## Remaining gates

- Repository CI on the submitted head.
- Maintainer review and merge decision.

### DIFF
--- a/docs/tutorial_constraints.rst
+++ b/docs/tutorial_constraints.rst
@@ -500,9 +500,12 @@ Overview
 
 1. Two curve-tangency or point-incidence targets + fixed radius
 2. Two curve-tangency or point-incidence targets + center constrained on a locus
-3. Three curve-tangency or point-incidence targets
+3. Three curve-tangency targets
 4. One curve-tangency or point-incidence target + fixed center
 5. One curve-tangency or point-incidence target + fixed radius + center constrained on a locus
+
+Point inputs are supported by signatures 1, 2, 4, and 5. The three-target
+signature requires curve inputs.
 
 ``sagitta`` selects short/long/both arc branches:
 
@@ -566,8 +569,8 @@ Signature C: Three constraints
 .. figure:: ./assets/tan3_ex.svg
    :align: center
 
-Use for an arc constrained by three curve-tangency or point-incidence targets. This can
-produce several branches; always consider using ``selector``.
+Use for an arc tangent to three curve targets. This can produce several branches;
+always consider using ``selector``.
 
 Signature D: One constraint + fixed ``center``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/tutorial_constraints.rst
+++ b/docs/tutorial_constraints.rst
@@ -410,8 +410,9 @@ Tangency Constraints
 Both :class:`~objects_curve.ConstrainedArcs` and :class:`~objects_curve.ConstrainedLines` 
 return a :class:`~topology.Curve` containing one or more :class:`~topology.Edge` objects.
 
-These constructors solve tangent/contact problems from mixed numeric and geometric inputs.
-Because tangency is often ambiguous, multiple valid branches are expected.
+These constructors combine curve-tangency constraints with point-incidence, center-locus,
+and orientation constraints. Because tangency is often ambiguous, multiple valid branches
+are expected.
 
 
 Multiple solutions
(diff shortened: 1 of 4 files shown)


gumyr:
Looks good to me. @jdegenstein does this PR clear up the confusing descriptions sufficiently?

docs/tutorial_constraints.rst
```diff
@@ -497,11 +498,11 @@ Overview
...
-1. Two tangency/contact objects + fixed radius
-2. Two tangency/contact objects + center constrained on a locus
-3. Three tangency/contact objects
-4. One tangency/contact object + fixed center
-5. One tangency/contact object + fixed radius + center constrained on a locus
+1. Two curve-tangency or point-incidence targets + fixed radius
+2. Two curve-tangency or point-incidence targets + center constrained on a locus
+3. Three curve-tangency or point-incidence targets
+4. One curve-tangency or point-incidence target + fixed center
+5. One curve-tangency or point-incidence target + fixed radius + center constrained on a locus
```
jdegenstein, line 505: can you review the accuracy of these 5 signature descriptions in light of the issue you linked to? I was under the impression that only no. 4 supported point incidence.

jdegenstein (commented):
Added a review comment. Depending on the outcome of that it may need some more changes to each signature.

docs/tutorial_constraints.rst
kforris: Thanks for flagging this. I checked all five signatures against current behavior: signatures 1, 2, 4, and 5 accept point inputs, while the three-target signature does not. I updated the overview and Signature C accordingly and narrowed the two public overload annotations in `e2166719` so they no longer advertise point inputs for signature 3. Validation: the 38 focused constrained-arc tests passed, the full suite passed with 2,210 tests and 2 skipped, Mypy passed, and the Sphinx build succeeded.

gumyr:
Thank you!

PR #1404, opened 2026-08-11
docs: add rotation methods example

### LINKED ISSUE #512: Add Rotation Example to Intro Docs
Currently not enough ways of applying rotation are shown in the intro docs here https://build123d.readthedocs.io/en/latest/introductory_examples.html

Need to add an example to illustrate this important concept more clearly.

kforris:
## Summary

- add Introductory Example 38 with builder-mode and algebra-mode variants
- compare object `rotation=`, a `Pos * Rot` placement, and `Shape.rotate(Axis, angle)`
- add the generated SVG that visually separates the three rotated boxes

Closes #512.

## Root cause

The introductory examples mention rotation and show a rotated placement plane, but do not compare the main ways users can apply rotation to an object. The new example keeps that comparison in one small, executable section instead of changing runtime behavior.

## Scope and risk

- Risk: low; documentation and executable documentation examples only
- Scope: one RST section, the matching builder/algebra snippets, and one generated SVG
- Non-goals: no geometry API, rotation semantics, existing example, packaging, workflow, or dependency changes

## Reproduction

On current `dev`, the Introductory Examples page has no dedicated comparison of constructor rotation, placement rotation, and rotation around an arbitrary axis. Example 38 now executes all three paths and produces three solids with total volume `2304.0`.

## Validation

- `python docs/general_examples_algebra.py` — passed
- `cd docs && python general_examples.py` — passed and generated `general_ex38.svg`
- focused assertions — 3 solids, positive volume, expected RST include markers, and valid SVG passed
- visual SVG review — three separated boxes with progressively larger rotations
- `make -C docs html` — passed; 8 existing warnings outside the changed RST section
- `pytest tests/test_importers.py tests/test_materials.py::TestMaterialGltfExport::test_gltf_json_pbr_values` with the environment CA bundle — 38 passed
- full serial `pytest` with the environment CA bundle — 2221 passed, 2 skipped
- parallel `pytest -n auto` — 2212 passed, 2 skipped; 8 setup errors came from the local Python CA fixture download and one failure came from the existing xdist fixed-filename `test.gltf` race; both affected scopes passed serially above
- `git diff --check` — passed
- whole-file Black, Pylint, and Mypy checks still report pre-existing issues before Example 38 (Ex. 8 formatting/import spacing, wildcard/example-script lint, and 13 earlier type errors); Black proposes no change in the added block and Mypy reports no error in it

## Documentation impact

This PR is the documentation update requested by the Issue. No additional changelog or API-reference update is needed because behavior is unchanged.

## Remaining gates

- maintainer review
- repository CI
- merge decision


docs/general_examples.py
```diff
@@ -762,3 +762,24 @@ def square(rad, loc):
...
 # show_object(ex37.part)
+
+##########################################
+# 38. Applying Rotations
+# [Ex. 38]
+length, width, height, spacing = 24, 8, 4, 30
+
+with BuildPart() as ex38:
+    Box(length, width, height, rotation=(0, 0, 20))
+    with Locations(Pos(Y=spacing) * Rot(Z=45)):
```
jdegenstein: Prefer these are nested location contexts as using algebra mode inside of  builder mode is not supported everywhere (which would be confusing for users).

docs/general_examples.py
```diff
@@ -762,3 +762,24 @@ def square(rad, loc):
...
+        Box(length, width, height)
+    rotated_box = Pos(Y=2 * spacing) * Box(length, width, height, mode=Mode.PRIVATE)
```
jdegenstein: See above, do not use algebra mode inside of builder mode

docs/general_examples_algebra.py
```diff
@@ -596,3 +596,22 @@ def square(rad, loc):
...
 # show_object(ex36)
+
+##########################################
+# 38. Applying Rotations
+# [Ex. 38]
+length, width, height, spacing = 24, 8, 4, 30
+
+rotated_with_constructor = Box(length, width, height, rotation=(0, 0, 20))
+rotated_with_location = Pos(Y=spacing) * Rot(Z=45) * Box(length, width, height)
+rotated_about_axis = (Pos(Y=2 * spacing) * Box(length, width, height)).rotate(
```
jdegenstein: I will let Roger weigh in on this one, not sure what approach is appropriate given my above comments.

docs/introductory_examples.rst
```diff
@@ -1144,3 +1144,38 @@ coordinates.
...
+shows the same three approaches with progressively larger angles so their
+effects are easy to compare.
+
+.. image:: assets/general_ex38.svg
+    :align: center
+
+* **Builder mode**
+
+    The first box uses the object's ``rotation`` parameter. The second box is
+    created inside a :class:`~build_common.Locations` context that combines
```
jdegenstein: update descriptions as appropriate

jdegenstein (commented):
Added a few review comments.

docs/general_examples.py
kforris: Addressed in 4ff8c575. The second box now uses nested `Locations` contexts: the outer context translates it and the inner `Locations(Rot(Z=45))` applies the rotation. There is no longer a `Pos * Rot` algebra expression in the builder example.

docs/general_examples.py
kforris: Addressed in 4ff8c575. The private `Box` is now created without algebra placement; it is rotated with `Shape.rotate(Axis.Z, 70)` and added under a `Locations((0, 2 * spacing))` context. The geometry signature remains identical: 3 solids, total volume 2304.0, and centers at Y=0/30/60.

docs/general_examples_algebra.py
kforris: Understood. I left the algebra-mode example unchanged and limited this update to removing algebra expressions from builder mode. I can adjust the algebra example after Roger weighs in.

docs/introductory_examples.rst
kforris: Updated in 4ff8c575. The text now describes the nested `Locations` contexts and the private `Shape.rotate` plus `add` flow, and no longer says that `Pos` and `Rot` are combined in builder mode.

kforris:
Thank you very much for taking the time to review this and for the helpful guidance.

I updated the builder-mode example in `4ff8c575` to use nested `Locations` contexts throughout, removing the algebra-style placement expressions from the builder flow. I also updated the accompanying description to match. The resulting geometry is unchanged, and the full repository CI is green.

I left the algebra-mode example unchanged for now, as suggested, so that Roger can advise on the preferred approach there.

Thanks again — please let me know if you would like any further adjustments.

docs/general_examples.py
```diff
@@ -762,3 +762,21 @@ def square(rad, loc):
     write_svg()
 
 # show_object(ex37.part)
+
+##########################################
+# 38. Applying Rotations
+# [Ex. 38]
+length, width, height, spacing = 24, 8, 4, 30
+
+with BuildPart() as ex38:
```
gumyr, line 771: How about this instead?

```py
with BuildPart() as ex38:
    with Locations(Rot(Z=20)):
        Box(length, width, height)
    with Locations((0, spacing)):
        Box(length, width, height, rotation=(0, 0, 40))
    box = Box(length, width, height, mode=Mode.PRIVATE)
    with Locations((0, 2 * spacing)):
        add(box.rotate(Axis.Z, 60))
    with BuildPart(Location((0, 3 * spacing), (0, 0, 80))):
        Box(length, width, height)
```
This version also shows the new ability to place (i.e. rotate) a part by specifying a location as a placement parameter to BuildPart.

docs/general_examples_algebra.py
gumyr: Maybe changing the order and doing the rotate about axis first without a position component would make these more clear?
```py
rotated_about_axis = Box(length, width, height).rotate(
    Axis((-length / 2, -width / 2), (0, 0, 1)), 20
)

rotated_with_constructor = Pos(Y=spacing) * Box(
    length, width, height, rotation=(0, 0, 40)
)
rotated_with_location = Pos(Y=2 * spacing) * Rot(Z=60) * Box(length, width, height)
```

docs/introductory_examples.rst
```diff
@@ -1144,3 +1144,39 @@ coordinates.
         :start-after: [Ex. 37]
         :end-before: [Ex. 37]
 
+
+.. _ex 38:
+
+38. Applying Rotations
+---------------------------------------------------
+
+Rotation can be applied when an object is created, through nested location
```
gumyr: ... and now as a placement parameter to the Builders.

gumyr (changes requested):
It's tricky getting maximum clarity in the examples.

kforris:
Thanks for the concrete suggestions. I updated the builder example to cover `Locations(Rot)`, constructor rotation, `Shape.rotate`, and the new `BuildPart(Location)` placement path; reordered the algebra example so the offset-axis rotation comes first; and regenerated the SVG in `7130fdae`.

Validation: both focused examples execute with the expected 4 builder solids / 3 algebra solids, the SVG was visually checked, Sphinx completes with the existing 8 unrelated warnings, Mypy passes, and the full serial suite passes with 2,221 tests plus 68 subtests (2 skipped). The parallel suite’s only failure was the known shared `test.glb` filename race; that test passes in isolation.

gumyr:
Thank you!

PR #1406, opened 2026-08-13
Add a reference to the ExportToPython project in the external tools

RanVisionware:
Add a reference to the ExportToPython project in the external tools and libraries section in the documentation.


gumyr:
This is an interesting project, good luck with it.

PR #1415, opened 2026-08-16
Fix Text resolved font attributes

### LINKED ISSUE #1235: Fix `Text.font` attribute to point at actual font
Currently creating a text object does not always give back the correct "resolved" font:
```py
label = Text("asdf", font_size=20, align=Align.CENTER)
print(label.font) # Returns Arial even though it is actually using e.g. Liberation Sans
```
font resolution happens inside `Compound.make_text`, maybe it should pass as properties to `Text` class?
```py
def make_text():
   ...
   text_flat.font = system_font.FontName().ToCString()
   text_flat.font_path = system_font.FontPath(FONT_ASPECT[font_style]).ToCString()
   return text_flat
```
then in `Text` class:
```py
class Text..
    ...
    text_string = Compound.make_text(
        txt=txt,
        font_size=font_size,
        font=font,
        font_path=font_path,
        font_style=font_style,
        text_align=text_align,
        align=align,
        position_on_path=position_on_path,
        text_path=path,
        single_line_width=single_line_width,
    )
    self.font = text_string.font # and delete earlier set attributes
    self.font_path = text_string.font_path 
    super().__init__(text_string, rotation, None, mode)
```

kforris:
## Summary

- expose the font name and file path that OpenCascade actually resolved on `Text.font` and `Text.font_path`
- centralize font lookup in the owner-proposed `Compound.resolve_font()` helper
- preserve the existing `Compound.make_text()` return type and avoid adding attributes to `Compound`
- add a regression for a missing requested font that falls back to a system font

Fixes #1235.

## Root cause

`Text` stored the caller-provided `font` and `font_path` values before `Compound.make_text()` resolved them through `FontManager`. When OpenCascade substituted a fallback font, the geometry used the fallback while the public attributes still reported the request.

The revision follows the maintainer's proposed design: `Compound.resolve_font()` returns the resolved name, path, and system-font object. `make_text()` uses it without changing its output, while `Text` uses the same helper to populate its existing public attributes. No dynamic metadata is attached to `Compound` or `Shape`.

## Scope and risk

Risk: low. The change is limited to resolving and reporting text font metadata. Glyph generation, alignment, paths, single-line outlining, and the `Compound.make_text()` signature and return type are unchanged.

Non-goals: changing fallback selection, font registration, font discovery, OpenCascade rendering, or introducing a generic metadata API.

Reproduced on current `dev` at `b8f85510840712ab63b2eb3c12b004f0ef4ad0a3`: requesting a deliberately missing font rendered with the OpenCascade fallback but left `Text.font` set to the missing name.

- focused text tests: 4 passed, 49 deselected
- required parallel suite: 2,225 passed, 2 skipped, with one `TestFontManager.test_register_system_fonts` failure
- the exact font-inventory failure reproduces in isolation on an unmodified `upstream/dev` worktree, so it is not caused by this PR
- full serial suite excluding that exact baseline test: 2,225 passed, 2 skipped, 1 deselected, 68 subtests passed
- Mypy on both changed source files: passed
- `git diff --check`: passed
- Sphinx HTML: passed with 7 pre-existing warnings
- Black reports only pre-existing formatting in `objects_sketch.py` and `test_build_sketch.py`; the changed `composite.py` is clean
- Pylint reports only two pre-existing `wrapped` attribute warnings in `composite.py`; the new lines add no warnings

## Documentation impact

The `Text` docstring states that `font` and `font_path` contain the resolved values after construction. No separate guide change is needed because the constructor inputs and user workflow are unchanged.

## Remaining gates

Repository CI, maintainer review, and merge remain external.

### DIFF
--- a/src/build123d/objects_sketch.py
+++ b/src/build123d/objects_sketch.py
@@ -681,8 +681,9 @@ class Text(BaseSketchObject):
             text_path=path,
             single_line_width=single_line_width,
         )
-        self.font = text_string.font
-        self.font_path = text_string.font_path
+        self.font, self.font_path, _ = Compound.resolve_font(
+            font, font_path, font_style
+        )
         super().__init__(text_string, rotation, None, mode)
 
 
--- a/src/build123d/objects_sketch.py
+++ b/src/build123d/objects_sketch.py
@@ -604,8 +604,10 @@ class Text(BaseSketchObject):
     Args:
         txt (str): text to render
         font_size (float): size of the font in model units
-        font (str, optional): font name. Defaults to "Arial"
-        font_path (PathLike | str, optional): system path to font file. Defaults to None
+        font (str, optional): requested font name. After construction, this attribute
+            contains the resolved font name. Defaults to "Arial"
+        font_path (PathLike | str, optional): system path to font file. After
+            construction, this attribute contains the resolved font path. Defaults to None
         font_style (Font_Style, optional): font style, REGULAR, BOLD, BOLDITALIC, or
             ITALIC. Defaults to Font_Style.REGULAR
         text_align (tuple[TextAlign, TextAlign], optional): horizontal text align
@@ -679,6 +681,8 @@ class Text(BaseSketchObject):
             text_path=path,
             single_line_width=single_line_width,
         )
+        self.font = text_string.font
+        self.font_path = text_string.font_path
         super().__init__(text_string, rotation, None, mode)
 
 
--- a/src/build123d/topology/composite.py
+++ b/src/build123d/topology/composite.py
@@ -65,6 +65,7 @@ from bd_materials import FinishedMaterial
 
 import OCP.TopAbs as ta
 from OCP.BRepAlgoAPI import BRepAlgoAPI_Fuse
+from OCP.Font import Font_SystemFont
 from OCP.gp import gp_Ax3
 from OCP.Graphic3d import (
     Graphic3d_HTA_LEFT,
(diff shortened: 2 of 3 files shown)


gumyr:
I don't really like the idea of dynamically adding `font` and `font_path` attributes to the `Compound` class in the `make_text` method - exactly what `pylint: disable-next=attribute-defined-outside-init` is complaining about and being disabled here. It seems like there are a couple ways to improve this:

1. Add the `font` and `font_path` attributes to the `Compound` classes in the constructor and initialize them to `None`.
2. Add a more general `metadata: dict` attribute to `Shape` or even just to `Compound` that could be used to carry `font` and `font_path` in this case but potentially other values in the future.

Any other ideas?

kforris:
I agree that dynamically attaching these attributes to `Compound` is not ideal. I think there is a narrower third option for this PR: extract the existing resolution/build path into an internal helper that returns both the generated `Compound` and the resolved font object (or resolved name/path). Public `Compound.make_text()` can keep returning only the `Compound`, while `Text.__init__` consumes both results and sets its existing `font` / `font_path` attributes.

That keeps text-specific fields off every `Compound`, and avoids introducing a general `metadata` API before its copy/transform/serialization semantics are defined. If that internal split is too invasive, option 1 seems safer than adding general metadata, but the helper approach looks like the narrowest scope to me.

gumyr:
I think there is a cleaner option: create a new `Compound.resolve_font` method - something like this:
```py
@staticmethod
def resolve_font(
    font: str = "Arial",
    font_path: PathLike[str] | str | None = None,
    font_style: FontStyle = FontStyle.REGULAR,
) -> tuple[str, str, Font_SystemFont]:

    requested_path = fspath(font_path) if font_path is not None else None
    manager = FontManager()

    if requested_path and manager.check_font(requested_path):
        face_names = manager.register_font(requested_path, True, False)
        selected_name = font if font in face_names else face_names[0]
    else:
        selected_name = font

    system_font = manager.find_font(selected_name, font_style)

    aspect = FONT_ASPECT[font_style]
    return tuple(
        system_font.FontName().ToCString(),
        system_font.FontPath(aspect).ToCString(),
        system_font,
    )
```
which would be called by `make_text` and again in the `Text` constructor - so there is a bit of overhead but this call isn't too computationally expensive. Therefore the `make_text` method doesn't change outputs, no new attributes are required on `Compound` or `Shape` and `Text` gets the information it needs.

kforris:
Implemented the proposed `Compound.resolve_font()` direction in 30c62c2d. `make_text()` still returns a plain `Compound`, no dynamic font attributes remain, and `Text` resolves/stores its name and path through the shared helper. Focused tests, Mypy, and docs pass; the full suite has only the same macOS FontManager inventory-count failure reproduced on unmodified upstream `dev`. Please re-review when convenient.

gumyr:
Thank you!

PR #1426, opened 2026-08-24
Report offset_2d builder failures clearly

### LINKED ISSUE #1425: offset_2d: BRepOffsetAPI_MakeOffset result is used without checking IsDone(), so a failed offset surfaces as a misleading "Null TopoDS_Shape object"
Authored by AI (Claude - Opus 5 w. ultracode), approved by me - a human :)

---

`Wire.offset_2d()` builds its result with `BRepOffsetAPI_MakeOffset` and reads `Shape()` without first checking `IsDone()`:

```python
# build123d/topology/one_d.py, 0.11.1
offset_builder = BRepOffsetAPI_MakeOffset()
offset_builder.Init(kind_dict[kind])
offset_builder.AddWire(topods_wire)
offset_builder.Perform(distance)

obj = downcast(offset_builder.Shape())      # <- no IsDone() check
```

When the offset fails to converge, `Perform()` leaves the builder not-done and `Shape()` returns a null shape. That null then propagates until something asks for its type, and the user gets:

```
ValueError: Null TopoDS_Shape object
```

raised from `topology/shape_core.py`. The message says nothing about offsetting, so the failure looks like a corrupt input shape rather than a solver that didn't converge. I spent a while looking for a malformed wire before finding the real cause.

**This is a departure from the file's own convention rather than an oversight in style.** `IsDone()` is checked 20 times across `build123d/topology/`, seven of them in this same file — `one_d.py` lines 1333, 2467, 2594, 3674, 4202, 4260 and 4580 all guard their builders before using the result. The offset builder is the exception.

I confirmed the method is available: `BRepOffsetAPI_MakeOffset` inherits `IsDone()` from `BRepBuilderAPI_Command`, present in cadquery-ocp 7.9.3.1.

Suggested change:

```python
offset_builder.Perform(distance)
if not offset_builder.IsDone():
    raise RuntimeError(f"offset_2d failed to converge at distance {distance}")
obj = downcast(offset_builder.Shape())
```

**Scope of what I'm reporting.** This is a code-correctness and diagnosability issue argued from the source. I hit it in practice — 5 of 30 faces imported from SVG raise the null-shape error at a small positive offset, while the other 25 offset fine in well under 0.1 s — but **I have not been able to build a synthetic reproducer.** Fourteen synthetic cases covering holes, slots, arcs, mirrored geometry and Bezier segments all offset without error, so I can't hand you a small file that triggers it. The missing guard is verifiable by reading regardless of which inputs cause the non-convergence.

Possibly related, though I don't think either is the same bug:

- #844 reports the identical `Null TopoDS_Shape object` message from `Shape.split`, which suggests this is a generic symptom of a null OCC result surfacing late rather than anything specific to offsetting.
- #739 is also `offset_2d`, but fails differently — a `TypeError` in the `side != Side.BOTH` post-processing branch rather than at the builder.
- #1224 is doing something similar in spirit for fillets: reporting which edges fail instead of failing opaquely.

Environment: build123d 0.11.1, cadquery-ocp-novtk 7.9.3.1.1, Python 3.13, Windows 11.

kforris:
## Summary

- check `BRepOffsetAPI_MakeOffset.IsDone()` before reading its result shape
- raise a contextual `RuntimeError` that includes the requested offset distance
- add a focused regression proving that a failed builder never reaches `Shape()`

Fixes #1425

## Root cause

`Mixin1D.offset_2d()` called `Shape()` immediately after `Perform()`. When OCCT reports the operation as not done, that result can be null, so the generic `ValueError: Null TopoDS_Shape object` escaped before build123d could identify the failed offset operation.

## Scope and risk

Risk is low. The patch adds one guard on the existing OCCT completion state and leaves the method signature, successful path, return types, join behavior, and side-selection logic unchanged.

Non-goals: changing offset algorithms, retrying failed offsets, handling multiple generated wires, or changing any other topology operation.

## Reproduction

On current `dev` (`2851b011`), a deterministic builder-failure regression showed that `IsDone()` was never called and `Shape()` was called once, surfacing `ValueError: Null TopoDS_Shape object`. With this patch, the same failure state raises `RuntimeError: 2D offset failed with distance 0.1`, and `Shape()` is not called.

## Validation

- `python -m pytest tests/test_direct_api/test_mixin1_d.py -q`: 41 passed
- focused failure regression: passed; diff coverage 100% for the two new production lines
- `black --check` on both changed files: unchanged
- `mypy src/build123d`: passed for 41 source files
- `pylint src/build123d/topology/one_d.py`: 9.98/10; three pre-existing warnings
- `python -m pytest -n auto`: 2,226 passed, 2 skipped, 1 failed

The sole full-suite failure is the existing environment-sensitive `TestFontManager.test_register_system_fonts` count assertion. It also fails serially on the untouched `2851b011` baseline with the same 1,828-after versus 2,382-before counts, so it is unrelated to these two changed files.

## Documentation impact

The public method docstring now documents the offset-calculation `RuntimeError`. No separate guide or example changes are needed because the successful API and workflow are unchanged.

## Remaining gates

Repository CI, maintainer review, and merge remain external gates.

### DIFF
--- a/src/build123d/topology/one_d.py
+++ b/src/build123d/topology/one_d.py
@@ -1156,6 +1156,7 @@ class Mixin1D(Shape[TOPODS]):
             closed (bool, optional): if Side!=BOTH, close the LEFT or RIGHT
                 offset. Defaults to True.
         Raises:
+            RuntimeError: 2D offset calculation failed
             RuntimeError: Multiple Wires generated
             RuntimeError: Unexpected result type
 
@@ -1185,6 +1186,8 @@ class Mixin1D(Shape[TOPODS]):
         # offset_builder.SetApprox(True)
         offset_builder.AddWire(topods_wire)
         offset_builder.Perform(distance)
+        if not offset_builder.IsDone():
+            raise RuntimeError(f"2D offset failed with distance {distance}")
 
         obj = downcast(offset_builder.Shape())
         if isinstance(obj, TopoDS_Compound):
--- a/tests/test_direct_api/test_mixin1_d.py
+++ b/tests/test_direct_api/test_mixin1_d.py
@@ -400,6 +400,18 @@ class TestMixin1D(unittest.TestCase):
         # self.assertTrue(offset_edge.geom_type == GeomType.LINE)
         # self.assertAlmostEqual(offset_edge.position_at(0).X, 3)
 
+    def test_offset_2d_builder_failure(self):
+        """Verify that OCCT offset failures are reported before reading Shape()."""
+        with patch("build123d.topology.one_d.BRepOffsetAPI_MakeOffset") as builder:
+            offset_builder = builder.return_value
+            offset_builder.IsDone.return_value = False
+            offset_builder.Shape.side_effect = ValueError("Null TopoDS_Shape object")
+
+            with self.assertRaisesRegex(RuntimeError, "2D offset failed.*0.1"):
+                Wire.make_rect(1, 1).offset_2d(0.1)
+
+            offset_builder.Shape.assert_not_called()
+
     def test_common_plane(self):
         # Straight and circular lines
         l = Edge.make_line((0, 0, 0), (5, 0, 0))


gumyr:
Thank you!

PR #1429, opened 2026-08-27
Fix through holes on offset parts

### LINKED ISSUE #996: `Hole` not taking part size/position into account properly
```py
with BuildPart() as ps:
    with BuildSketch(Plane.XY.offset(745-12.5-15)) as s:
        RectangleRounded(355,390,50)
    extrude(amount=-90)
    fillet(faces().sort_by(Axis.Z)[0].edges(),20)

    with GridLocations(50,430-25,2,2):
        Hole(26/2,mode=Mode.ADD)
```

Image

kforris:
## Summary

- size through-hole cutters from the current part bounds in each requested location's local Z frame
- keep explicit `depth` behavior unchanged
- add a regression for a part positioned away from the global origin

Fixes #996.

## Root cause

`Hole(depth=None)` used twice the part bounding-box diagonal but always centered the cutter on the local origin. A part translated far enough from that origin could therefore sit completely outside the cutter. The patch projects the current part's bounding-box corners into each requested location frame, then uses the combined local Z extent for the cutter center and length. This is limited to automatic through-hole sizing.

## Scope and risk

Risk: medium-low. The change is confined to `Hole` construction and one focused regression. Explicit-depth holes, other hole types, public signatures, and general builder placement behavior are unchanged. No geometry abstraction, dependency, workflow, or compatibility policy is added.

On current `dev` (`2851b011`), the issue scenario produced hole geometry at Z `-535..535` while the part occupied Z `627.5..717.5`. After the patch, the hole geometry occupies Z `627.5..717.5`.

- `python -m pytest tests/test_build_part.py -q` — 77 passed
- `python -m pytest tests/test_build_part.py --cov=src/build123d --cov-report=xml --cov-fail-under=0 -q` — 77 passed
- `diff-cover coverage.xml --compare-branch=origin/dev --fail-under=95` — 100% diff coverage
- `black --check --fast` on changed ranges and the changed test file — passed
- `pylint src/build123d/objects_part.py` — 10.00/10
- `mypy src/build123d` — passed for 41 source files
- `git diff --check` — passed
- `python -m pytest -n auto` — 2,226 passed, 2 skipped, with the existing macOS `TestFontManager.test_register_system_fonts` inventory assertion as the only failure; the same `1828 < 2382` failure reproduced on untouched `dev`

## Documentation impact

None. This restores the documented through-part behavior for `depth=None` without changing the API or user workflow.

## Remaining gates

Repository CI, maintainer review, and the maintainer merge decision remain external. The unrelated macOS system-font inventory assertion remains an environment gate.


jdegenstein:
I believe this problem is likely to occur across all `*Hole` types including also `CounterSinkHole` and `CounterBoreHole`. Can you confirm that? With that in mind maybe the size logic should be abstracted out of `Hole` itself and into somewhere re-usable?

src/build123d/objects_part.py
```diff
@@ -416,8 +418,29 @@ def __init__(
         self.radius = radius
         if depth is not None:
             self.hole_depth = 2 * depth
+            hole_center = 0.0
         elif depth is None and context is not None:
-            self.hole_depth = 2 * context.max_dimension
+            part_bbox = context.part_local.bounding_box()
+            part_corners = [
```
gumyr: I like this version better:
```py
                # A bounding box is contained by a sphere centered on the box with
                # its diagonal as diameter. Projecting that sphere onto each hole's
                # local Z axis therefore bounds the part in every rotated frame.
                half_diagonal = part_bbox.diagonal / 2
                local_centers = [
                    Vector(
                        part_bbox.center()
                        .to_pnt()
                        .Transformed(location.inverse().wrapped.Transformation())
                    ).Z
                    for location in self._get_object_locations()
                ]
                min_z = min(center - half_diagonal for center in local_centers)
                max_z = max(center + half_diagonal for center in local_centers)
```
The minimal length hole that you appear to be creating here is only minimal for a rectangular shape aligned to the global axes so just using the diagonal length is generally the same in functionality and less complex. Also, the `product` import isn't required.

tests/test_build_part.py
```diff
@@ -500,6 +500,17 @@ def test_through_hole(self):
                 Hole(2)
         self.assertAlmostEqual(test.part.volume, 1000 - 4 * 10 * pi, 5)
 
+    def test_through_hole_offset_part(self):
+        with BuildPart() as test:
+            with Locations((0, 0, 100)):
+                Box(10, 10, 10)
+            hole = Hole(2, mode=Mode.ADD)
```
gumyr, line 566: This is kinda a strange test for a hole that isn't a hole. If `hole = Hole(2)` is used instead both the checks on the placement of `hole` can be checked and that the volume of the part has be reduced.

gumyr (changes requested):

gumyr:
Thank you. @jdegenstein comment made me realize that the problem does apply beyond Hole so I've updated the PR to make it more general.

PR #1430, opened 2026-08-28
Add surface patch and solid creation operations

### LINKED ISSUE #1077: Create a new patch_surface or patch_hole operation
```py
def patch_surface(
    hole: Wire | ShapeList[Edge],
    continuity: ContinuityLevel,
    surface_pnts: Iterable[VectorLike] | None = None,
) -> Face:
    # Extract edge/face pairs & build constraints
    constraints = []
    for hole_edge in hole.edges():
        connected_topods_faces = topo_explore_connected_faces(
            hole_edge, hole_edge.parent
        )
        if len(connected_topods_faces) != 1:
            raise ValueError("Each edge of the hole must be connected to just one face")

        constraints.append((hole_edge, Face(connected_topods_faces[0]), continuity))

    patch = Face.make_surface_patch(constraints, point_constraints=surface_pnts)
    return patch
```

The constraints to `make_surface_patch` method can automated for the user to make the creation of surface patches easy.  Note that `topo_explore_connected_faces` needs to be added to the `__init__.py`.  As with all new content, tests are required and the documentation need to be updated.

Here is a complete example:
```py
length = 100 * MM
width = 60 * MM
thickness = 20 * MM
corner_radius = 10 * MM

perimeter = RectangleRounded(width, length, corner_radius).face().outer_wire()
rim_profile_pln = Plane(perimeter.location_at(0, x_dir=(-1, 0, 0)))
rim_profile = (
    rim_profile_pln
    * Pos(X=-width / 2)
    * EllipticalCenterArc(
        (0, 0), width / 2, thickness / 2, start_angle=320, end_angle=40
    )
)
rim = Shell.sweep(rim_profile, perimeter)
top_hole = rim.edges().group_by(Axis.Z)[-1]

top = patch_surface(top_hole, ContinuityLevel.C1, [(0, 0, thickness / 2)])
bottom = mirror(top, Plane.XY)
soap = Solid(Shell(rim.faces() + [top, bottom]))
```

Image

kforris:
## Summary

Closing a non-planar hole currently requires manually pairing boundary edges with their support faces. This PR adds `patch_surface` to perform that mapping and create a continuity-constrained face, plus `make_solid` to turn a closed set of faces into a solid.

- Support both Algebra mode and `BuildPart`: surface patches become pending faces, and `make_solid()` consumes pending faces after successful solid creation.
- Use edge orientation to select a unique support face when the topology parent is a solid.
- Add Algebra and Builder examples, operation references, and tests for successful construction and invalid inputs.

Closes #1077.

## Implementation and scope

`patch_surface` uses the existing topology exploration helpers and delegates the surface calculation to `Face.make_surface_patch`. `make_solid` sews its input faces into one closed shell, validates the resulting solid, and then applies the requested builder combination mode or returns an Algebra part.

The operations are additive. Existing surface-construction APIs and continuity semantics are preserved. Detached or ambiguous boundary edges and faces that cannot define a single valid solid are rejected.

The feature includes new geometry operations and Builder integration. The latest follow-up is low risk: it adds only three tests (18 lines) for empty input, mixed face/edge input, and disconnected closed shells; it changes no production code or public behavior.

## Validation

Validated from the previous PR head `e98698a`, with current upstream `dev` at `5b1b4b5` used for the diff-coverage comparison.

- Before the test additions: the 164 focused tests passed, but diff coverage was 89% (42/47 lines), below the 95% threshold. This reproduced the existing CI failure.
- After the additions: `python -m pytest tests/test_build_part.py tests/test_build_generic.py -q --cov=src/build123d --cov-report=xml --cov-fail-under=0`: 167 passed.
- Full parallel suite: `python -m pytest -n auto --dist loadscope --benchmark-disable -q`: 2,250 passed, 2 skipped, 68 subtests passed. Class-scoped scheduling keeps the shared importer fixtures together.
- Full serial coverage run: 2,242 passed, 2 skipped, with 8 importer setup errors caused by the local Python CA configuration. After setting `SSL_CERT_FILE` to the installed Certifi bundle, all 8 importer tests passed on a serial rerun with `--cov-append`. Combined coverage is 94.89%, above the 94% repository threshold.
- `diff-cover coverage.xml --config-file pyproject.toml --compare-branch=upstream/dev`: passed, 97% diff coverage (46/47 lines).
- `black --check tests/test_build_part.py`: passed.
- `pylint --rcfile=.pylintrc --fail-under=9.5 --fail-on=E,F src/build123d`: passed, 9.97/10.
- `mypy --config-file mypy.ini src/build123d`: passed for 41 source files.
- `git diff --check`: passed.

## Documentation impact

The operation reference, cheat sheet, and introductory examples cover the new operations and both usage modes. The test-only follow-up needs no additional documentation changes because it verifies existing validation behavior.

## Remaining gates

CI on the updated PR head, maintainer review, and the merge decision remain pending.


docs/cheat_sheet.rst
```diff
@@ -87,6 +87,7 @@ Cheat Sheet
             | :func:`~operations_generic.fillet`
             | :func:`~operations_generic.mirror`
             | :func:`~operations_generic.offset`
+            | :func:`~operations_generic.patch_surface`
```
jdegenstein: If patch_surface returns a non-planar face then it definitely is not a 1D operation. It is also not applicable to 2D - BuildSketch either as those currently all focus on planar 2D objects. Unfortunately this patch_surface object ideally requires an entirely new builder (e.g. BuildSurface). It is usable in algebra mode as is of course.

gumyr:
I've pushed a commit that adds to this feature:
- `patch_surface` now publishes its output to the `BuildPart.pending_faces`
- a new `make_solid` operation is added that converts 2D faces (the builder's pending_faces) into a 3d solid in the same way that the `make_face` operation converts 1D edges/wires into a 2D face

When `BuildSurface` is added it will also use these same mechanisms - the 2D faces it creates will be published to `BuildPart` where `make_solid` will convert to a solid.

@jdegenstein please review this. 

P.S. I'll create an example to illustrate these new capabilities in the Builder context...

gumyr:
I've pushed two examples, one each for Algebra and Builder mode. When developing these I found a limitation in how face selection works so there is a fix for that here as well. The fix avoids the problem of a user explicitly providing faces but having the topo_parent set to a solid which causes the built in protection to reject those faces.

kforris:
Following up on the remaining coverage failure: `b8cfa566` adds three tests for `make_solid` with empty input, mixed face/edge input, and two disconnected closed shells. The previous head reproduced the 89% diff-coverage failure; these tests raise it to 97% (46/47 lines), above the 95% threshold. Production code is unchanged.

Validation: 167 focused tests passed; the full parallel suite passed with 2,250 tests, 2 skips, and 68 subtests. The PR description records the full serial coverage run and successful rerun of the eight importer tests after correcting the local CA path, along with lint/type/format checks.

I also updated the description to reflect the Builder support, `make_solid`, and examples now in the branch. CI is running on the new head. @jdegenstein, could you take another look when convenient? Please let me know if there is any remaining author-side work.

Issue #943, opened 2025-03-19
Fillet Not Handling Slicing Correctly


barrettluke:
Description
When applying fillets using Build123d, methods 1-4 fillet all four Z-axis edges as expected but unexpectedly fillet two edges along the X-axis and two along the Y-axis at the bottom. However, methods 5 and 6 behave as expected when slicing edges.

Steps to Reproduce
	1.	Use the following script with Build123d and ocp_vscode:
```python
from build123d import *
from ocp_vscode import show

# Dimensions
length, width, thickness = 120.0, 110.0, 30.0
fillet_radius = 3.0
with BuildPart() as ex1:
    # Create the box
    Box(length, width, thickness)

    # Method 1 
    # fillet(ex1.edges().filter_by(Axis.Z), radius=fillet_radius)
    # fillet(ex1.edges().filter_by(Axis.X)[-1], radius=fillet_radius)
    #-----------------------------------
    # Method 2
    # fillet(ex1.edges().filter_by(Axis.Z), radius=fillet_radius)
    # test = ex1.edges().filter_by(Axis.X)[-1]
    # fillet(test, radius=fillet_radius)
    #-----------------------------------
    # Method 3
    # fillet(ex1.edges().filter_by(Axis.Z), radius=fillet_radius)
    # fillet(test := ex1.edges().filter_by(Axis.X)[2], radius=fillet_radius)
    #-----------------------------------
    # Method 4
    # fillet(ex1.edges().filter_by(Axis.Z), radius=fillet_radius)
    # fillet(test := ex1.edges().filter_by(Axis.X)[-1], radius=fillet_radius)
    #-----------------------------------
    # Method 5 (Expected Behavior)
    # axis_z = ex1.edges().filter_by(Axis.Z) + ex1.edges().filter_by(Axis.X)[::2] + ex1.edges().filter_by(Axis.Y)[1::2]
    # fillet(axis_z, radius=fillet_radius)
    #-----------------------------------
    # Method 6 (Expected Behavior)
    # b_half = ex1.edges().filter_by(Axis.Z) + [ex1.edges().filter_by(Axis.X)[-1]]
    # fillet(b_half, radius=fillet_radius)

# Show the result
show(ex1.part)
```
	2.	Uncomment methods 1-4 one at a time and observe that they fillet the four Z-axis edges as expected, but also fillet unexpected edges on the X and Y axes at the bottom.
	3.	Uncomment methods 5 and 6 to see that they behave as expected.

Expected Behavior
[...]

Issue #1390, opened 2026-08-03
fuse missing geometry with partially coincident surfaces


trayracing:
Fusing two parts sometime loses a portion of geometry, probably due to partially coincident shells? In the image below, fusing to overlapping elbows loses geometry. The left column shows the two elbows, the bottom right the error from fusing the two, and the top right is a partial workaround by rotating the circles one of the elbows. Both e0.fuse(e1) and e1.fuse(e0) fail.

This suggested by Jern, on the premise that the seam from circle at the coincidence of the two parts was triggering the error. And that seemed to work. However, I rotate circles of both parts, the problem remains, even though the seam don't line up. So that conjecture is only part of the story.

I also tried tweaking the tolerance parameter of fuse, which didn't help.

Code attached. Seen in Version 0.11.1.

<img width="680" height="657" alt="Image" src="https://github.com/user-attachments/assets/0c9a6638-36d2-4c22-9b41-a028aa639083" />

[fuseFlaw.py](https://github.com/user-attachments/files/30641688/fuseFlaw.py)

gumyr:
OCCT (the OpenCascade CAD kernel) has known problems with non-planar geometry as you've shown here. Unfortunately, there is little that build123d can do about these types of problems. I do have a prototype of a `clean` mechanism for spheres that can in some situations remove the seam which seems to help but that doesn't address the fundamental OCCT limitations.

trayracing:
Is there a corresponding OCCT bug to link to? Seems giving OCCT simple cases like this would help track down the problem(s) or at least be part of a test set to verify any such fix.

trayracing:
Also, it turns out the rotation workaround is iffy. If I only loft 8 of the 10 sections, the fuse fails for the workaround (top right in the image below): 

<img width="652" height="635" alt="Image" src="https://github.com/user-attachments/assets/aac7da11-aa0f-458c-b786-ef8e171e18e6" />

Issue #1332, opened 2026-06-11
Subtracting a helical-sweep tool from a cylinder silently returns an empty shape


riosbotchen-source:
## Description

A V-groove profile swept along a `Helix` produces a valid solid (positive volume, exportable). Subtracting that tool from a plain cylinder via `add(tool, mode=Mode.SUBTRACT)` (or `cyl - tool`) does not raise, but returns an **empty result: volume 0, 0 faces** — instead of either a grooved cylinder or an error.

The helix is not self-intersecting (pitch 2.0, groove depth 1.0, well below the self-contact limit mentioned in the docs tips page). The expected result is the cylinder minus a shallow helical groove.

If this is a known OCCT boolean limitation with helical sweeps, a loud failure (exception) instead of a silent empty shape would already be a big improvement — the empty result propagates and is only noticed much later.

## Reproduction

```python
from build123d import *

pitch, height, radius = 2.0, 10.0, 8.0
groove_depth = 1.0

# Helical path slightly longer than the cylinder so the groove runs off both ends.
path = Helix(pitch=pitch, height=height + 2 * pitch, radius=radius, center=(0, 0, -pitch))
start, tangent = path @ 0, path % 0
radial = Vector(start.X, start.Y, 0).normalized()
section_plane = Plane(origin=start, x_dir=radial, z_dir=tangent)

# V-groove tool swept along the helix: this alone is a valid solid.
with BuildPart() as tool:
    with BuildSketch(section_plane):
        with BuildLine():
            w = pitch * 0.5
            Polyline([(0.05, -w / 2), (0.05, w / 2), (-groove_depth, 0)], close=True)
        make_face()
    sweep(path=path, is_frenet=True)
print(f"groove tool volume: {tool.part.volume:.2f}")  # 177.95 (valid, positive)

with BuildPart() as part:
    Cylinder(radius=radius, height=height, align=(Align.CENTER, Align.CENTER, Align.MIN))
    add(tool.part, mode=Mode.SUBTRACT)

print(f"after subtract: volume={part.part.volume:.2f}, faces={len(part.part.faces())}")
```

Output:

```
groove tool volume: 177.95
after subtract: volume=0.00, faces=0
```

Expected: volume slightly below the full cylinder (2010.62), many faces.
Actual: silent empty shape.

## Environment

Reproduced identically on both:

- build123d 0.10.0 (latest release) + cadquery-ocp 7.8.1.1.post1 (OCCT 7.8)
[...]

Issue #1064, opened 2025-08-07
Error in helix rendering


cornim:
I'm on build123d version 0.9.1 with Python 3.12.8.

The following script
```Python
r = 1
rm = 10
h = 40

ph = Helix(30, h, rm)

circle = Plane(ph@0, z_dir=ph%0) * Circle(r)
h1 = sweep(circle, ph)
bottom = extrude(Plane.XY.offset(-r) * Circle(rm + 2*r), 2*r)
bottom = fillet(bottom.edges(), 0.9)

res = h1  + bottom

show(res)
```

results in the following picture:

<img width="673" height="889" alt="Image" src="https://github.com/user-attachments/assets/3ceef6bd-0613-45f5-a6ab-6ecdcb128941" />

The missing parts (circled in red) are also missing in an exported STEP file.

Noteable is that for smaller heights (<30) the issues dissapears.
I also noticed that `bottom = extrude(Plane.XY.offset(-1.097*r) * Circle(rm + 2*r), 2*r)` produces the error while `bottom = extrude(Plane.XY.offset(-1.098*r) * Circle(rm + 2*r), 2*r)` does not.

If there are any infos missing, please let me know.

gumyr:
This is a known bug with OCCT (OpenCascade). The good news is that the OCCT 8.0.0 release is coming soon which has new helix functionality so there is some hope that this will be fixed before long.

PR #1435, opened 2026-09-03
Improve OCCT Robustness and Fix Misc. Bugs

### LINKED ISSUE #590: Basic operation results in invalid geometry
```py
p = Box(2,2,2) - Pos(-1.5) * Sphere(1)
```
Screenshot

### LINKED ISSUE #745: STEP Missing Surfaces?
I'm not sure where to post this and I'm not sure if this is a bug in build123d, FreeCAD, or something else.  So I'm looking for some guidance in debugging I guess.

I have a part that seems to build find and also displays as expected in Codium using [ocp_vscode](https://github.com/bernhard-42/vscode-ocp-cad-viewer).  However, when I export the file to STEP and import it into FreeCAD, a bunch of surfaces appear to be missing.

Has anyone run into this?  Does anyone have any recommendations to debug this?

For reference, here is the part in ocp_vscode:

image

And here it is imported into FreeCAD (latest version, 0.21.2):

image

Notice that the top surfaces are missing.

### LINKED ISSUE #902: mirror() Creates ghost geometry
`mirror(Plane.XY)` creates ghost geometry in my example, which add volume to the part and render as selectable faces in ocp_vscode. As discussed on discord, this may be due to https://github.com/Open-Cascade-SAS/OCCT/pull/26

Originally, this was to report where the combination of the previous `fillet(ppp0110... , 8)` and the `mirror(Plane.XY)` caused the entire operation to silently fail, but I was unable to reproduce in 0.9.0.

```python
with BuildPart() as ppp0110:
    with BuildSketch(Plane.YZ) as s:
        with BuildLine() as l:
            JernArc((0, 0), (1, 0), 40, 90)
            Line((0, 46), (42, 46))
        make_hull()

    with BuildSketch(Plane.YZ) as s2:
        arc = s.edges().sort_by(Axis.Y)[:-1].sort_by(Axis.X)[1:]
        make_face(offset(arc, amount=8, side=Side.LEFT))
        Rectangle(46, 16, align=Align.MIN)
        add(s, mode=Mode.INTERSECT)

    extrude(s2.sketch, amount=100 / 2)
    fillet(ppp0110.faces().filter_by(Plane.XY).sort_by(Axis.Y)[0].edges().sort_by(Axis.Y)[-1], 8)
    revolve(ppp0110.faces().sort_by(Axis.X)[0], axis=Axis.Z, revolution_arc=90)
    mirror(about=Plane.YZ.offset(100 / 2))
    # This fails but it should work
    mirror(about=Plane.XZ)

show(ppp0110)
```

Image

### LINKED ISSUE #987: Solid.extrude_taper might flip the direction or extrusion

### LINKED ISSUE #1123: Cutout not performed completely
I have just tried to cut out a rounded hole in a cylindrical shell. The program below is a 2D-design with an extrusion to produce the desired 3D-shape.

```
from build123d import *
from ocp_vscode import *

def build_part1():
    part1 = Part()
    l1 = CenterArc((0, 0),50,-30,60)
    l2 = CenterArc((0, 0),60,-30,60)
    l3 = Line(l1 @ 0, l2 @ 0)
    l4 = Line(l1 @ 1, l2 @ 1)
    part1_2d = make_face([l1,l2,l3,l4])
    part1_3d = extrude(part1_2d,amount=50)
    c3 = Location((50,0,25),(0,+90,0)) * Cylinder(10,100)
    part1 += part1_3d
    part1 -= c3
    part1 = Pos(0,0,50) * part1
    export_stl(part1,"part1.stl")
    return part1

if __name__ == "__main__":
    part1 = build_part1()

    show(part1)
```
In the OCP-CAD-Viewer part1 has not the expected appearance, instead of the circle in the middle, it should be a cylindrical hole, all material should be completely removed. Here is what I got:

Image

The generation of the STL-file was without any errors. However most other tools that load this file do very strange things and emit some errors. PrusaSlicer finds 337 open edges. Or on another code-example PrusaSlicer finds degenerated areas which are auto-repaired during loading, what has the final result, that the hole disappears completely!

Image

PrusaSlices renders part1 very simialr to OCP-CAD-Viewer.

I have tried to narrow down the problem. It seems that the extrusion works fine, but the input for the extrusion is somehow wrong. My lines are really closed. And it should work. Please help. What is wrong with the code, how can this problem be avoided?

### LINKED ISSUE #1205: Combining rotated lines bug
```py
a = Line(0, 30)
a = a + Rot(Z=-120) * a + Rot(Z=120) * a
b = Rot(Z=60) * a
c = a + b
```

Image
Image
Image

### LINKED ISSUE #1216: ExportSVG misses subtracted shape when using chamfer
I'm using the following code:

```
from build123d import *

with BuildSketch() as s:
    Rectangle(20, 20)
    Rectangle(5,5,mode=Mode.SUBTRACT)
    chamfer(s.vertices().group_by(Axis.Y)[-1], length=2)

for (exporter, suffix) in [(ExportSVG, "svg"), (ExportDXF, "dxf")]:
    exporter = exporter(unit=Unit.MM, line_weight=0.5)
    exporter.add_layer("Layer 1")
    exporter.add_shape(s.sketch, layer="Layer 1")
    exporter.write("chamfer_example."+suffix)
```

The subtracted 5x5 rectangle is not present in the exported SVG. It is present in the exported DXF, and can be seen if using ops-vscode.

If you remove the chamfer line then the subtracted rectangle does show up in the SVG.

I am using 0.10.1.dev176+g4d4622f91 on Ubuntu 24.04.3 LTS, x86_64.

I can use ExportDXF as a workaround for now, but it's a lot easier to use SVG for our laser cutting software.

### LINKED ISSUE #1271: Invalid faces
```sh
~/.venv/build123d/bin/pip install ocp_vscode~=3.3.4 git+https://github.com/gumyr/build123d
```

Following script produces object with invalid faces. If I save to step, the file can't be even opened.

```python
from build123d import *
from math import *

r_outer = 40
r_inner = 35
bump_r = 3
rows = 8

r_bump = r_outer - 1

with BuildPart() as nest:
    for lat in range(rows):
        angle_lat = 10 + lat * (80 / rows)
        rl = r_outer * sin(radians(angle_lat))
        cols = round(2 * pi * rl / (bump_r * 2.8))
        for j in range(cols):
            angle_lon = j * 360 / cols
            x = r_bump * sin(radians(angle_lat)) * cos(radians(angle_lon))
            y = r_bump * sin(radians(angle_lat)) * sin(radians(angle_lon))
            z = r_bump * cos(radians(angle_lat))
            with Locations(Location((x, y, z))):
                Sphere(bump_r, mode=Mode.ADD)

    Sphere(r_outer, arc_size1=0, arc_size2=90, align=(Align.CENTER, Align.CENTER, Align.MIN))  # upper half
    Sphere(r_inner, arc_size1=-90, arc_size2=90, mode=Mode.SUBTRACT)

export_stl(nest.part, "nest.stl")
```

Image

Image

### LINKED ISSUE #1351: Rounded loft failing to offset in 0.11
In 0.10, I was creating two sketches. The bottom one is smaller than the top one.

Image

As a loft, they create this body:

Image

Doing an offset with `-1` and the top and bottom faces open resulted in this:

Image

With 0.11, the `offset` call fails with:

```
  File ".venv/lib/python3.13/site-packages/build123d/topology/three_d.py", line 680, in offset_3d
    offset_occt_solid = offset_builder.Shape()
OCP.OCP.StdFail.StdFail_NotDone: BRep_API: command not done

The above exception was the direct cause of the following exception:

  File "run.py", line 43, in <module>
    b.offset(
    ~~~~~~~~^
        amount=-1,
        ^^^^^^^^^^
        openings=part.faces().group_by(b.Axis.Z)[-1]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        + part.faces().group_by(b.Axis.Z)[0],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File ".venv/lib/python3.13/site-packages/build123d/operations_generic.py", line 678, in offset
    solid.offset_3d(openings_in_this_solid, amount, kind=kind).fix()
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.13/site-packages/build123d/topology/three_d.py", line 682, in offset_3d
    raise RuntimeError(
        "offset Error, an alternative kind may resolve this error"
    ) from err
RuntimeError: offset Error, an alternative kind may resolve this error
```

I didn't find any `kind` that works. Do you have any idea? 

I could create a new left with two offsetted sketches and do a subtracting loft, but offset was so nice.

Here is a reproducing script:

```python
# build123d

import build123d as b
import ocp_vscode

WIDTH = 41.2
length = 24.4 * 3

with b.BuildPart() as part:
    with b.BuildSketch(b.Plane.XY.offset(40)) as skt:
        with b.BuildLine():
            b.Polyline(
                [
                    (0, 0),
                    (0, WIDTH),
                    (length, WIDTH),
                    (length, 0),
                    ((length / 2 + 20), 0),
                    ((length / 2 + 20), 10),
                    ((length / 2 - 20), 10),
                    ((length / 2 - 20), 0),
                    (0, 0),
                ]
            )
        b.make_face()

        vtouter = (
            skt.vertices().group_by(b.Axis.X)[0] + skt.vertices().group_by(b.Axis.X)[-1]
        )
        b.fillet(vtouter, radius=7)

        vtinner = (
            skt.vertices().group_by(b.Axis.X)[2] + skt.vertices().group_by(b.Axis.X)[3]
        )
        b.fillet(vtinner, radius=2.5)

    with b.BuildSketch(b.Plane.XY) as skb:
        b.add(skt)
        b.offset(amount=-1)

    b.loft([skb.sketch, skt.sketch])

    b.offset(
        amount=-1,
        openings=part.faces().group_by(b.Axis.Z)[-1]
        + part.faces().group_by(b.Axis.Z)[0],
    )

ocp_vscode.show(part)
```

### LINKED ISSUE #1363: Boolean subtraction fails for specific hole orientations and positions
Hi,

I found that a boolean subtraction fails under certain geometric configurations. Instead of creating a hole, the preview only displays the intersection edge where the cut should occur. The exported model (either step or stl) matches the preview and also contains no hole.

The issue is configuration dependent. It only occurs for certain combinations of the hole's position and rotation. Small changes to either the hole location or its orientation can cause the subtraction to succeed again.

Here is the simplified code from my sample, with comments where the bug appears. I wrote it via OCP CAD on VSCodium. Many thanks.

build123d version: 0.11.1
Python version: 3.14
Operating system: macOS 15.6

```python
# build123d
from build123d import *
from ocp_vscode import *

blocks = [
    ("shoulderL",1,18,-90),
    ("shoulderR",1,18, 90), # if the blocks are not matched, the hole will subtract    
]

holes = [
    ("hole", 10, 4), # if the hole is half way out of the block, it will subtract
]

def add_blocks(cylind, blockPos, blockLength, blockRotation):
    with BuildPart() as bp_main:
        with BuildSketch(Plane.XZ):
            with Locations((0, blockPos)):
                Rectangle(9, blockLength, align=Align.MIN)
        revolve(axis=Axis.Z)
        with BuildPart(mode=Mode.SUBTRACT):
            with BuildSketch(Plane.XZ):
                with Locations((6, blockPos)):
                    Rectangle(3, blockLength, align=Align.MIN)
            revolve(axis=Axis((-25, 0, 0), (0, 0, 1)))
    cylind.part += bp_main.part

def add_hole(model, z, d):
    model.part -= (
        Pos(0, 0, z)
        * Rot(X=-90, Y=90) #when X is between 0 and 180, it won't subtract 
        * extrude(Circle(radius=d/2),amount=10)
    )

with BuildPart() as cylind:
    with BuildSketch(Plane.XZ) as profile:
        Rectangle(6, 20, align=Align.MIN)
    revolve(axis=Axis.Z)

for _, blockPos, blockLength, blockRotation in blocks:
    add_blocks(cylind, blockPos, blockLength, blockRotation)

for _, z, d in holes:
    add_hole(cylind, z, d)

show_all()
``` 

Image
Image

### LINKED ISSUE #1428: clean() after boolean operations deletes faces crossed by a curved surface's parametric seam
`Sphere(5)` centered in `Box(8, 8, 8)` pokes a 14.66 mm³ spherical cap out of each box face. The boolean operators (and the `.cut()`/`.fuse()` methods) lose one of them:

| operation | volume | expected |
|---|---|---|
| `sphere - box` (operator) | 73.3038 | 87.9646 |
| `sphere + box` (operator) | 551.6439 | 599.9646 |
| raw `BRepAlgoAPI_Cut` | **87.9646** | 87.9646 |
| raw `BRepAlgoAPI_Fuse` | **599.9646** | 599.9646 |
| raw cut → `.clean()` | 73.3038 | (loss reproduced) |
| raw fuse → `.clean()` | 551.6439 | (loss reproduced) |

The raw booleans are correct, and `.clean()` alone reproduces the loss on their results — isolating the problem to the `ShapeUpgrade_UnifySameDomain` pass that `clean()` wraps and that the operators run automatically after every boolean.

The lost cap is the one crossed by the sphere's parametric seam meridian. Rotating the sphere about Z before cutting confirms it — the degenerate cap follows the seam, and when the seam points at a box *edge* (crossing no cap face) the result is fully correct:

```
  0 deg: volume  73.3038  degenerate cap at (+4.0,+0.0,-0.8)
 90 deg: volume  73.3038  degenerate cap at (+0.0,+4.0,-0.8)
180 deg: volume  73.3038  degenerate cap at (-4.0,+0.0,-0.7)
 45 deg: volume  87.9646  correct (seam aimed at a box edge)
```

A further consequence: the broken fuse result exports to STEP as faces only (no closed solid), so downstream tools (e.g. PrusaSlicer) read the model as zero volume.

## Workaround / proposed fix

Run the boolean raw, then keep `clean()` only when it conserves volume. Note `clean()` mutates `self.wrapped` in addition to returning, so the guard must clean a copy:

```python
def guarded_clean(shape: Shape) -> Shape:
    before = shape.volume
    cleaned = copy.deepcopy(shape).clean()
    ok = abs(cleaned.volume - before) < 1e-9 * max(1.0, before)
    return cleaned if ok else shape
```

In our benchmarks the guarded pair (raw boolean + guarded clean) is *faster* than the current unconditional clean (16.3 ms vs 22.9 ms per fuse on this model). Caveat: a volume guard protects solids; a hypothetical clean() failure that preserved volume while mangling faces would slip past it — so this is offered as a mitigation, with a real fix presumably belonging in/around the UnifySameDomain call.

## Reproduction

Self-contained script and the four exported STEP files (broken vs workaround pairs, for visual comparison — the cut pair shows 5 caps vs 6) are in this gist:
https://gist.github.com/etjones/c68e633b07dc7184504b9b303a93555e

<details>
<summary>repro_clean_bug.py (same as in the gist)</summary>

```python
"""Reproduction: the automatic clean() after boolean operations deletes
real geometry when the result contains a face crossed by a curved
surface's parametric seam.

Model: Sphere(5) and Box(8, 8, 8), both centered. The sphere pokes a
14.66 mm^3 spherical cap out of each of the six box faces.

    expected sphere - box = 6 caps           =  87.9646
    expected sphere + box = 523.60 + 512 - intersection = 599.9646

Observed (build123d 0.11.1, cadquery-ocp 7.9.3.1.1, macOS arm64):
`-` and `+` (and the .cut()/.fuse() methods) lose the cap crossed by the
sphere's parametric seam meridian; the fuse result additionally exports
to STEP as faces only (slicers read it as zero volume).

The boolean itself is sound: running the same operation through
BRepAlgoAPI directly gives the correct result, and calling .clean() on
that correct result reproduces the loss exactly -- isolating the bug to
the ShapeUpgrade_UnifySameDomain pass clean() wraps. Rotating the
sphere about Z moves the degenerate cap with the seam (0 deg -> +X,
90 deg -> +Y, 180 deg -> -X), and at 45 deg, where the seam points at a
box *edge* and crosses no cap face, the result is correct.

Workaround: run the raw boolean, then clean a *copy* and keep the clean
only when volume is conserved (note clean() mutates self.wrapped in
addition to returning, so the copy is required).
"""

import copy

from build123d import Box, Part, Rot, Shape, Sphere, export_step
from OCP.BRepAlgoAPI import BRepAlgoAPI_Cut, BRepAlgoAPI_Fuse
from OCP.TopTools import TopTools_ListOfShape

EXPECTED_CUT = 87.9646
EXPECTED_FUSE = 599.9646


def raw_bool(op, a: Shape, b: Shape) -> Part:
    """One OCCT boolean with no clean() afterwards."""
    args, tools = TopTools_ListOfShape(), TopTools_ListOfShape()
    args.Append(a.wrapped)
    tools.Append(b.wrapped)
    op.SetArguments(args)
    op.SetTools(tools)
    op.Build()
    return Part(op.Shape())


def guarded_clean(shape: Shape) -> Shape:
    """The proposed workaround/fix: keep clean() only when it conserves
    volume. clean() mutates in place, hence the copy."""
    before = shape.volume
    cleaned = copy.deepcopy(shape).clean()
    ok = abs(cleaned.volume - before) < 1e-9 * max(1.0, before)
    return cleaned if ok else shape


sphere, box = Sphere(5), Box(8, 8, 8)

print(f"{'operation':<26} {'volume':>10}   expected")
print(f"{'sphere - box (operator)':<26} {(sphere - box).volume:10.4f}   {EXPECTED_CUT}")
print(f"{'sphere + box (operator)':<26} {(sphere + box).volume:10.4f}   {EXPECTED_FUSE}")

raw_cut = raw_bool(BRepAlgoAPI_Cut(), sphere, box)
raw_fuse = raw_bool(BRepAlgoAPI_Fuse(), sphere, box)
print(f"{'raw BRepAlgoAPI_Cut':<26} {raw_cut.volume:10.4f}   {EXPECTED_CUT}")
print(f"{'raw BRepAlgoAPI_Fuse':<26} {raw_fuse.volume:10.4f}   {EXPECTED_FUSE}")

# clean() alone reproduces the loss on the correct raw results:
print(f"{'raw cut  -> .clean()':<26} {copy.deepcopy(raw_cut).clean().volume:10.4f}   (loss reproduced)")
print(f"{'raw fuse -> .clean()':<26} {copy.deepcopy(raw_fuse).clean().volume:10.4f}   (loss reproduced)")

# workaround: volume-guarded clean
print(f"{'guarded_clean(raw cut)':<26} {guarded_clean(raw_cut).volume:10.4f}   {EXPECTED_CUT}")
print(f"{'guarded_clean(raw fuse)':<26} {guarded_clean(raw_fuse).volume:10.4f}   {EXPECTED_FUSE}")

# the degenerate cap follows the seam as the sphere rotates:
print("\nseam location vs lost cap (sphere rotated about Z, then - box):")
for angle in (0, 90, 180, 45):
    d = (Rot(0, 0, angle) * Sphere(5)) - Box(8, 8, 8)
    dead = [x for x in d.solids() if x.volume < 1e-6]
    note = "correct (seam aimed at a box edge)" if not dead else (
        f"degenerate cap at ({dead[0].center().X:+.1f},"
        f"{dead[0].center().Y:+.1f},{dead[0].center().Z:+.1f})"
    )
    print(f"  {angle:3d} deg: volume {d.volume:8.4f}  {note}")

# STEP exports for visual comparison (broken vs workaround):
export_step(sphere - box, "cut_operator_broken.step")
export_step(guarded_clean(raw_cut), "cut_workaround.step")
export_step(sphere + box, "fuse_operator_broken.step")
export_step(guarded_clean(raw_fuse), "fuse_workaround.step")
print("\nwrote cut/fuse STEP pairs (operator vs workaround)")
```

</details>

**Environment**: build123d 0.11.1, cadquery-ocp-novtk 7.9.3.1.1, Python 3.13.5, macOS 15 arm64.

zalo:
@gumyr @jdegenstein

This PR adds build123d-side mitigations for the OpenCascade defects behind the open `occt`-labelled issues. Every issue was first reduced to a minimal build123d script and a pure OCCT C++ program and checked against OCCT 7.7.2, 7.8.1, 7.9.3 and 8.0.1; you can see the deep dive on each of these issues in the Discord message: https://discord.com/channels/964330484911972403/1275536240816558080/1544935794924585042

## `ShapeUpgrade_UnifySameDomain` guard (`clean()` and every boolean)

Root cause of #1428, #590, #1271, #1363 and #1123: the raw `BRepAlgoAPI_Cut/Fuse` is valid and has the correct volume; build123d's `clean()` (`UnifySameDomain(shape, True, True, True)`) then merges the two boundary arcs of a face that a boolean split along the seam of a periodic surface into one closed edge whose pcurve is rebuilt on the wrong branch of the period, producing an invalid face / zero-volume cap. `UnifyEdges=False` avoids it in every case. #902 is the same class run on a shape fused with its mirror image: same-domain sphere/torus faces with reflected frames make the face unification crash (7.8.1 throws, 7.9.3 segfaults, 8.0.1 returns +43 % ghost volume); `UnifyFaces=False` avoids it. All of these still reproduce on OCCT 8.0.1.

New `unify_same_domain()` in `shape_core.py` runs the full unification as before; when the shape contains periodic faces the result is validated with `BRepCheck_Analyzer` (1–10 ms on typical parts) and **only if rejected** recomputed with `UnifyEdges=False`, then `UnifyFaces=False`, else the un-unified shape is kept with a warning. The one up-front decision is the crash case, which cannot be observed after the fact: faces on one sphere/torus with reflected (or, for spheres, re-oriented) axis systems disable `UnifyFaces`. Torus patches from fillets share a torus with different x-directions and are not flagged.

Behaviour for shapes without periodic surfaces is unchanged (no extra check).

## Other fixes

* **#987** `extrude_taper`: `LocOpe_DPrism` extrudes along the plane axis and ignores a `TopAbs_REVERSED` face, so a +Z profile on a −Z plane was extruded downwards. The profile is rebuilt as a FORWARD face on a plane whose axis is the extrusion direction.
* **#567** `extrude_taper`: a non-solid `LocOpe_DPrism` result (the drafted profile collapses before the full height) now raises a clear `ValueError` instead of `Standard_TypeMismatch`.
* **#1216** `Face.chamfer_2d`: `BRepFilletAPI_MakeFillet2d` merges all wires of a face with holes into one invalid wire; chamfer each wire separately (as `fillet_2d` already does).
* **#1205** `Wire.edges()`: `BRepTools_WireExplorer` skips edges at branch vertices of non-manifold wires; the missing edges are appended so `a + b` of "Y"-shaped wires keeps all edges. (OCCT builds the wire correctly; this was a build123d enumeration issue.)
* **#745** `export_step`: `STEPControl_Writer` silently drops edges and faces on `Geom_OffsetCurve` geometry. On an export copy, offsets of lines and circles (and extrusion surfaces built on them) are swapped in place for the equal line/circle, which has the same parametrisation so the topology and pcurves are untouched; only offsets of B-splines are approximated with `ShapeCustom::BSplineRestriction`.
* **#1351** lofts: `ThruSections` describes straight lateral edges as degree-1 `Geom_BezierCurve`s, on which `BRepOffsetAPI_MakeThickSolid` fails while the identical `Geom_Line` works; they are replaced by lines.
* **#545** `offset_3d`: raise when `MakeThickSolid` "succeeds" but returns the input solid unchanged (Kind.INTERSECTION did that silently).
* **#1332** booleans: warn when a cut returns an empty shape although the tools' total volume is smaller than the argument's (the OCCT classification failure with helical sweeps).
* **#1011 / #943**: documented in the `fillet()` docstring (OCCT tangent-chain propagation, no opt-out exists).

Not addressed (genuine OCCT defects without a safe build123d workaround): #955, #1390, #568, #693. #903 and #980 are fixed in OCCT 7.9.2+/7.9; #1064 is not reproducible on any OCCT version.

## Tests

`tests/test_occt_workarounds.py` covers each item above with the reduced geometry from the issues. The rest of the suite was run; the only remaining failures are the pre-existing font-dependent text tests, which fail identically on `dev` in this environment.

Fixes #1428, fixes #590, fixes #1271, fixes #1363, fixes #1123, fixes #902, fixes #987, fixes #1216, fixes #1205, fixes #745, fixes #1351.

### DIFF
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -384,9 +384,7 @@ def _offset_curve_edges(shape: TopoDS_Shape) -> list[TopoDS_Edge]:
     """Edges of shape whose 3D curve is a Geom_OffsetCurve"""
     edge_map = TopTools_IndexedMapOfShape()
     TopExp.MapShapes_s(shape, ta.TopAbs_EDGE, edge_map)
-    edges = [
-        TopoDS.Edge_s(edge_map.FindKey(i)) for i in range(1, edge_map.Extent() + 1)
-    ]
+    edges = [TopoDS.Edge(edge_map.FindKey(i)) for i in range(1, edge_map.Extent() + 1)]
     return [
         e
         for e in edges
@@ -394,8 +392,10 @@ def _offset_curve_edges(shape: TopoDS_Shape) -> list[TopoDS_Edge]:
     ]
 
 
-def _exact_offset_curve(curve: Geom_OffsetCurve) -> Geom_Curve | None:
+def _exact_offset_curve(curve: Geom_Curve) -> Geom_Curve | None:
     """The line or circle equal to an offset curve with a line or circle basis"""
+    if not isinstance(curve, Geom_OffsetCurve):
+        return None
     basis = curve.BasisCurve()
     while isinstance(basis, Geom_TrimmedCurve):
         basis = basis.BasisCurve()
@@ -440,7 +440,7 @@ def _without_offset_curves(shape: TopoDS_Shape) -> TopoDS_Shape:
             builder.UpdateEdge(edge, exact, location, BRep_Tool.Tolerance_s(edge))
     explorer = TopExp_Explorer(shape, ta.TopAbs_FACE)
     while explorer.More():
-        face = TopoDS.Face_s(explorer.Current())
+        face = TopoDS.Face(explorer.Current())
         explorer.Next()
         location = TopLoc_Location()
         surface = BRep_Tool.Surface_s(face, location)
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -425,13 +425,9 @@ def _without_offset_curves(shape: TopoDS_Shape) -> TopoDS_Shape:
     in place by the equal line or circle, which has the same parametrisation so
     the topology is unaffected. Other offset curves are approximated by B-splines.
     """
-    edges = _offset_curve_edges(shape)
-    if not edges:
-        return shape
-
     builder = BRep_Builder()
     approximate = False
-    for edge in edges:
+    for edge in _offset_curve_edges(shape):
         location = TopLoc_Location()
         exact = _exact_offset_curve(BRep_Tool.Curve_s(edge, location, 0.0, 0.0))
         if exact is None:
--- a/src/build123d/exporters3d.py
+++ b/src/build123d/exporters3d.py
@@ -29,6 +29,7 @@ license:
 # pylint has trouble with the OCP imports
 # pylint: disable=no-name-in-module, import-error
 
+import copy as copy_module
 import tempfile
 import warnings
 import webbrowser
@@ -42,8 +43,20 @@ import OCP.TopAbs as ta
 import requests
 from anytree import PreOrderIter
 from OCP.APIHeaderSection import APIHeaderSection_MakeHeader
+from OCP.BRep import BRep_Builder, BRep_Tool
+from OCP.BRepAdaptor import BRepAdaptor_Curve
 from OCP.BRepMesh import BRepMesh_IncrementalMesh
 from OCP.BRepTools import BRepTools
+from OCP.Geom import (
+    Geom_Circle,
+    Geom_Curve,
+    Geom_Line,
+    Geom_OffsetCurve,
+    Geom_SurfaceOfLinearExtrusion,
+    Geom_TrimmedCurve,
+)
+from OCP.GeomAbs import GeomAbs_C1, GeomAbs_CurveType
+from OCP.gp import gp_Vec
 from OCP.IFSelect import IFSelect_ReturnStatus
 from OCP.IGESControl import IGESControl_Controller
 from OCP.Interface import Interface_Static
@@ -51,6 +64,7 @@ from OCP.Message import Message, Message_Gravity, Message_ProgressRange
 from OCP.RWGltf import RWGltf_CafWriter
 from OCP.STEPCAFControl import STEPCAFControl_Controller, STEPCAFControl_Writer
 from OCP.STEPControl import STEPControl_Controller, STEPControl_StepModelType
+from OCP.ShapeCustom import ShapeCustom, ShapeCustom_RestrictionParameters
 from OCP.StlAPI import StlAPI_Writer
 from OCP.TCollection import (
     TCollection_AsciiString,
@@ -61,7 +75,10 @@ from OCP.TColStd import TColStd_IndexedDataMapOfStringString
 from OCP.TDataStd import TDataStd_Name
 from OCP.TDF import TDF_Label
 from OCP.TDocStd import TDocStd_Document
-from OCP.TopExp import TopExp_Explorer
+from OCP.TopExp import TopExp, TopExp_Explorer
+from OCP.TopLoc import TopLoc_Location
+from OCP.TopTools import TopTools_IndexedMapOfShape
+from OCP.TopoDS import TopoDS, TopoDS_Edge, TopoDS_Shape
 from OCP.XCAFApp import XCAFApp_Application
 from OCP.XCAFDoc import XCAFDoc_ColorType, XCAFDoc_DocumentTool, XCAFDoc_ShapeTool
 from OCP.XSControl import XSControl_WorkSession
@@ -70,8 +87,8 @@ from threejs_materials import PbrProperties, inject_materials
 
 from build123d.build_constants import UNITS_PER_METER
 from build123d.build_enums import PrecisionMode, Unit
-from build123d.geometry import Location
-from build123d.topology import Compound, Curve, Part, Shape, Sketch
+from build123d.geometry import TOLERANCE, Location
+from build123d.topology import Compound, Curve, Part, Shape, Sketch, downcast
 
 
 def _create_xde(
@@ -363,6 +380,107 @@ def export_gltf(
     return status
 
 
+def _offset_curve_edges(shape: TopoDS_Shape) -> list[TopoDS_Edge]:
+    """Edges of shape whose 3D curve is a Geom_OffsetCurve"""
+    edge_map = TopTools_IndexedMapOfShape()
+    TopExp.MapShapes_s(shape, ta.TopAbs_EDGE, edge_map)
+    edges = [
+        TopoDS.Edge_s(edge_map.FindKey(i)) for i in range(1, edge_map.Extent() + 1)
+    ]
+    return [
+        e
+        for e in edges
+        if BRepAdaptor_Curve(e).GetType() == GeomAbs_CurveType.GeomAbs_OffsetCurve
+    ]
+
+
+def _exact_offset_curve(curve: Geom_OffsetCurve) -> Geom_Curve | None:
+    """The line or circle equal to an offset curve with a line or circle basis"""
+    basis = curve.BasisCurve()
+    while isinstance(basis, Geom_TrimmedCurve):
+        basis = basis.BasisCurve()
+    normal = gp_Vec(curve.Direction())
+    if isinstance(basis, Geom_Line):
+        line = basis.Lin()
+        shift = gp_Vec(line.Direction()).Crossed(normal)
+        if shift.Magnitude() < TOLERANCE:
+            return None
+        shift.Normalize()
+        shift.Multiply(curve.Offset())
+        return Geom_Line(line.Location().Translated(shift), line.Direction())
+    if isinstance(basis, Geom_Circle):
+        circle = basis.Circ()
+        axis = gp_Vec(circle.Axis().Direction())
+        if not axis.IsParallel(normal, TOLERANCE):
+            return None
+        radius = circle.Radius() + curve.Offset() * (1 if axis.Dot(normal) > 0 else -1)
+        return Geom_Circle(circle.Position(), radius) if radius > TOLERANCE else None
+    return None
+
+
+def _without_offset_curves(shape: TopoDS_Shape) -> TopoDS_Shape:
+    """Replace Geom_OffsetCurve geometry, which the STEP writer silently drops
+
+    Offsets of lines and circles (and surfaces extruded from them) are replaced
+    in place by the equal line or circle, which has the same parametrisation so
+    the topology is unaffected. Other offset curves are approximated by B-splines.
+    """
+    edges = _offset_curve_edges(shape)
+    if not edges:
+        return shape
+
+    builder = BRep_Builder()
+    approximate = False
+    for edge in edges:
+        location = TopLoc_Location()
+        exact = _exact_offset_curve(BRep_Tool.Curve_s(edge, location, 0.0, 0.0))
+        if exact is None:
+            approximate = True
+        else:
+            builder.UpdateEdge(edge, exact, location, BRep_Tool.Tolerance_s(edge))
+    explorer = TopExp_Explorer(shape, ta.TopAbs_FACE)
+    while explorer.More():
+        face = TopoDS.Face_s(explorer.Current())
+        explorer.Next()
+        location = TopLoc_Location()
+        surface = BRep_Tool.Surface_s(face, location)
+        if isinstance(surface, Geom_SurfaceOfLinearExtrusion) and isinstance(
+            surface.BasisCurve(), Geom_OffsetCurve
+        ):
+            exact = _exact_offset_curve(surface.BasisCurve())
+            if exact is None:
+                approximate = True
+                continue
+            extrusion = Geom_SurfaceOfLinearExtrusion(exact, surface.Direction())
+            builder.UpdateFace(face, extrusion, location, BRep_Tool.Tolerance_s(face))
+
+    if approximate:
+        parameters = ShapeCustom_RestrictionParameters()
+        parameters.ConvertPlane = False
+        parameters.ConvertBezierSurf = False
+        parameters.ConvertRevolutionSurf = False
+        parameters.ConvertCurve3d = False
+        parameters.ConvertCurve2d = False
+        shape = ShapeCustom.BSplineRestriction_s(
+            shape, 1e-4, 1e-6, 9, 300, GeomAbs_C1, GeomAbs_C1, True, True, parameters
+        )
+    return shape
+
+
+def _prepare_for_step(to_export: Shape) -> Shape:
+    """Copy of to_export with geometry the STEP writer can't handle replaced"""
+    if not any(
+        node.wrapped is not None and _offset_curve_edges(node.wrapped)
+        for node in PreOrderIter(to_export)
+    ):
+        return to_export
+    exported = copy_module.deepcopy(to_export)
+    for node in PreOrderIter(exported):
+        if node.wrapped is not None:
+            node.wrapped = downcast(_without_offset_curves(node.wrapped))
+    return exported
+
+
 def export_step(
     to_export: Shape,
     file_path: PathLike | str | bytes | BytesIO | BinaryIO,
@@ -393,6 +511,8 @@ def export_step(
     Returns:
         bool: success
     """
+
+    to_export = _prepare_for_step(to_export)
 
     # Create the XCAF document
     doc = _create_xde(to_export, unit, auto_naming=True)
--- a/src/build123d/operations_generic.py
+++ b/src/build123d/operations_generic.py
@@ -413,6 +413,11 @@ def fillet(
     Fillet the given sequence of edges or vertices. Note that vertices on
     either end of an open line will be automatically skipped.
 
+    Note:
+        3D fillets propagate along chains of tangent edges (e.g. around a rounded
+        slot, or the top of a box whose vertical edges were rounded earlier);
+        OpenCascade offers no way to disable this.
+
     Args:
         objects (Edge | Vertex or Iterable of): edges or vertices to fillet
         radius (float): fillet size - must be less than 1/2 local width
--- a/src/build123d/topology/one_d.py
+++ b/src/build123d/topology/one_d.py
@@ -4036,14 +4036,25 @@ class Wire(Mixin1D[TopoDS_Wire]):
         explorer = BRepTools_WireExplorer(self.wrapped)
 
         edge_list: ShapeList[Edge] = ShapeList()
+        topo_parent = self if self.topo_parent is None else self.topo_parent
         while explorer.More():
             next_edge = Edge(explorer.Current())
             # pylint: disable=attribute-defined-outside-init
-            next_edge.topo_parent = (
-                self if self.topo_parent is None else self.topo_parent
-            )
+            next_edge.topo_parent = topo_parent
             edge_list.append(next_edge)
             explorer.Next()
+
+        # the WireExplorer skips edges at branch vertices of non-manifold wires
+        all_edges = TopTools_IndexedMapOfShape()
+        TopExp.MapShapes_s(self.wrapped, ta.TopAbs_EDGE, all_edges)
+        if all_edges.Extent() != len(edge_list):
+            for i in range(1, all_edges.Extent() + 1):
+                topods_edge = all_edges.FindKey(i)
+                if not any(topods_edge.IsSame(e.wrapped) for e in edge_list):
+                    missing_edge = Edge(TopoDS.Edge_s(topods_edge))
+                    # pylint: disable=attribute-defined-outside-init
+                    missing_edge.topo_parent = topo_parent
+                    edge_list.append(missing_edge)
         return edge_list
 
     def fillet_2d(self, radius: float, vertices: Iterable[Vertex]) -> Wire:
--- a/src/build123d/topology/one_d.py
+++ b/src/build123d/topology/one_d.py
@@ -4051,7 +4051,7 @@ class Wire(Mixin1D[TopoDS_Wire]):
             for i in range(1, all_edges.Extent() + 1):
                 topods_edge = all_edges.FindKey(i)
                 if not any(topods_edge.IsSame(e.wrapped) for e in edge_list):
-                    missing_edge = Edge(TopoDS.Edge_s(topods_edge))
+                    missing_edge = Edge(TopoDS.Edge(topods_edge))
                     # pylint: disable=attribute-defined-outside-init
                     missing_edge.topo_parent = topo_parent
                     edge_list.append(missing_edge)
--- a/src/build123d/topology/shape_core.py
+++ b/src/build123d/topology/shape_core.py
@@ -3855,7 +3855,7 @@ def _periodic_surfaces(shape: TopoDS_Shape) -> list[GeomAdaptor_Surface]:
     try:
         explorer = TopExp_Explorer(shape, TopAbs_ShapeEnum.TopAbs_FACE)
         while explorer.More():
-            face = TopoDS.Face_s(explorer.Current())
+            face = TopoDS.Face(explorer.Current())
             explorer.Next()
             adaptor = GeomAdaptor_Surface(BRep_Tool.Surface_s(face))
             if adaptor.IsUPeriodic() or adaptor.IsVPeriodic():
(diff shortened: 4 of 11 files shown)


zalo:
## Status of the `occt`-labelled issues after #1435

Every issue was reduced to a minimal build123d script and a clean OpenCascade C++ program and run against
OCCT 7.7.2, 7.8.1, 7.9.3 (the version behind OCP 7.9.3.1) and 8.0.1. Unless noted, the defects reproduce on
all four versions, so the upstream reports are still worth filing.

### Fixed in build123d by #1435 – close once merged

| issue | fix | suggested close comment |
|---|---|---|
| #1428, #590, #1271, #1363, #1123 | One shared root cause: `ShapeUpgrade_UnifySameDomain` merges the two halves of a seam-split periodic face into a single closed edge with a collapsed pcurve. `clean()` (used after every boolean) now validates the unified shape with `BRepCheck_Analyzer` whenever periodic faces are present and falls back to `UnifyEdges=False`, then `UnifyFaces=False`, else keeps the un-unified shape. | Fixed by #1435 (all five are the same OCCT defect). Reported upstream as an OCCT issue: `UnifySameDomain` corrupts seam-split periodic faces. |
| #902 | Not the fuse: `UnifySameDomain` crashes (7.9.3 segfault, 7.8.1 exception, 8.0.1 ghost volume) when merging same-domain sphere/torus faces whose frames differ by a reflection. Such faces are now detected before the call and `UnifyFaces` is disabled for them. | Fixed by #1435; reported upstream separately from the seam issue. |
| #987 | `LocOpe_DPrism` extrudes along the plane axis and ignores a `REVERSED` face. `extrude_taper` rebuilds the profile as a FORWARD face whose plane axis is the extrusion direction. | Fixed by #1435. |
| #1216 | `BRepFilletAPI_MakeFillet2d` dumps all edges of a face with holes into one wire. `Face.chamfer_2d` now processes each wire separately, as `fillet_2d` already did. Duplicate of #958, related to #720. | Fixed by #1435; duplicate of #958. |
| #1205 | build123d bug, not OCCT: `Wire.edges()` used `BRepTools_WireExplorer`, which skips edges at branch vertices. Missing edges are now added from `TopExp::MapShapes`. | Fixed by #1435; remove the `occt` label. |
| #745 | The STEP writer has no branch for `Geom_OffsetCurve` and silently drops every edge/face using one. `export_step` now replaces offset curves with a line/circle basis by the exact line/circle and approximates the remaining B-spline offsets with `ShapeCustom::BSplineRestriction`. | Fixed by #1435; reported upstream (writer should convert or at least warn). |
| #1351 | `MakeThickSolid` fails on `ThruSections` lofts because the lateral edges are degree-1 Bezier curves rather than lines. Lofts now replace those edges by `Geom_Line` edges, after which thickening succeeds. | Fixed by #1435. |

### Silent failure turned into an error or warning – close, defect stays upstream

| issue | change in #1435 | suggested close comment |
|---|---|---|
| #567 | `LocOpe_DPrism` returns an open shell with `IsDone()==true` when the draft collapses part of the profile. `extrude_taper` now raises `ValueError` explaining that the taper collapses the profile. | Now a clear error instead of `Standard_TypeMismatch`; the degenerate input cannot be built. |
| #545 | `MakeThickSolid` with `Join=Intersection` silently returns the input solid; with `Join=Arc` it returns an invalid shell. `offset_3d` now raises `RuntimeError` when the solid was not hollowed. | Now a clear error; OCCT defect reported upstream together with #1351. |
| #1332 | The Boolean kernel classifies the cylinder's split faces as inside the thin helical sweep, so `Cut` returns an empty shape with `HasErrors()==false`. Booleans now warn when a cut returns an empty shape although the tools are smaller than the argument. | Warning added; the cut itself needs an upstream fix (BOPAlgo classification). |
| #1011, #943 | Duplicates: `MakeFillet::Add(edge)` propagates over the whole G1-continuous edge chain by design and has no public opt-out. Documented in the `fillet()` docstring; workaround is to fillet all edges of the chain in one call. | Document as known OCCT behaviour and close #943 as a duplicate of #1011; feature request (opt-out flag) filed upstream. |

### Still open after the merge – genuine OCCT defects without a safe build123d workaround

| issue | finding | action |
|---|---|---|
| #955 | Two symmetric interior edges added to one `MakeFillet` builder (r ≥ 2) give `IsDone()==true` but an unorientable shape; each edge alone or sequentially is fine. | Keep open, link the upstream report. |
| #1390 | `Fuse` of two `ThruSections` elbows whose base circles are internally tangent misses the BSpline/BSpline face intersection when one seam lies on the tangency line; result equals one operand. | Keep open, link the upstream report. |
| #568 | Outward `BRepOffsetAPI_MakeOffset(GeomAbs_Intersection)` fails whenever a reflex vertex turns by more than 90° (interior angle > 270°). Arc/Tangent joins are exact. | Keep open, link the upstream report. |
| #693 | The STEP writer's `DirectFaces` step flips a `REVERSED` torus face and writes `FACE_BOUND .F.`; OCCT, FreeCAD and Mayo re-read it, OnShape and SolidWorks reject it. | Keep open as interoperability issue; suggest bound normalisation upstream. |

### Closeable independently of #1435

| issue | verdict | suggested close comment |
|---|---|---|
| #903 | `GeomFill_CorrectedFrenet` hang fixed upstream in OCCT 7.9.2 (OCCT #370, PR #630); does not occur with OCP 7.9.3. | Fixed upstream in OCCT 7.9.2. |
| #980 | `UnifySameDomain::UnifyFaces` infinite loop, fixed in OCCT 7.9; hangs only on 7.7.2/7.8.1. | Fixed upstream in OCCT 7.9. |
| #1064 | Not reproducible with a faithful construction on 7.7.2, 7.8.1, 7.9.3, 8.0.1 or in current build123d (fuse volume exact, shape valid). | Cannot reproduce; close unless the reporter can supply a failing environment. |

### Duplicate map

* #1428 ≡ #590 ≡ #1271 ≡ #1363 ≡ #1123 – one `UnifySameDomain` seam defect.
* #1011 ≡ #943 – tangent-chain fillet propagation.
* #1216 ≡ #958, related #720.
* #1351 ~ #545 – same `MakeThickSolid`-on-loft path, two failure modes.
* #902 and #980 – also `UnifySameDomain`, but distinct defects.

zalo:
Related OCCT issue: https://github.com/Open-Cascade-SAS/OCCT/issues/1541

PR #1436, opened 2026-09-03
Correct description of Solids and Parts to 3D Shapes

iparsrc:


jdegenstein:
Merged. Thank you for finding and fixing this!

PR #1442, opened 2026-09-05
Changes for build123d to run under OCP 8.0.1.0.0

bernhard-42:
I split the commits into the 4 groups of changes:

- Replace NCollection_Utf8String with NCollection_String
- New way to access BoundBox coords
- Use of the new OCP.collection classes
- bumping pyproject.toml

Note, I set an upper bound of 8.1 for `cadquery_ocp_novtk`, since it sounded like the API might change again


bernhard-42:
@gumyr I adapted the PR for the latest dev branch.

3 categories of changes:
1) `OCP.collection` related
2) A change in OCCT 8 that now requires for smoothing to C2 continuity degree 6 instead of 5.
3) `BRepTopAdaptor_FClass2d` still in OCCT 8, but dropped by OCP 8. Replaced with `IntTools_FClass2d` which claude assessed to be a drop-in replacement.

--- 

I think 2 is worth a check from your side, since you know best the intention of the code and whether the change is fine for you.

--- 

For 3 I additionally asked perplexity for a second opinion:

> They are two APIs for essentially the same job: classify a UV-space point relative to a trimmed TopoDS_Face, returning TopAbs_IN, TopAbs_OUT, or TopAbs_ON. In OCCT 8, the practical distinction is mostly historical/package/API shape, not a choice between fundamentally different classification semantics

Aspect | BRepTopAdaptor_FClass2d | IntTools_FClass2d
-- | -- | --
Header/package | BRepTopAdaptor_FClass2d.hxx / BRepTopAdaptor | IntTools_FClass2d.hxx / IntTools
Intended context | General B-Rep/topology-adaptor face classification | Intersection/Boolean-algorithm infrastructure
Primary input | TopoDS_Face, UV tolerance | Same
Main query | Perform(gp_Pnt2d, RecadreOnPeriodic) | Same
Periodic faces | Supports UV recadrage by default | Explicitly documented to adjust a point to the surface period by default
Other operations | PerformInfinitePoint(), TestOnRestriction() | Same, plus IsHole()
Initialization lifecycle | Construct with face and tolerance | Can default-construct, then call Init(face, tol); also has face/tolerance constructor
OCCT 8 recommendation | Prefer this one when you need this legacy classifier directly | Treat as legacy compatibility/intersection-package API unless you specifically need its lifecycle or IsHole()

However, perplexity continues:

> For new code, write the BRepTopAdaptor form.

But we simply don't have it any more in OCP 8

gumyr:
Here is Claude's (Fable) description - these seems like good changes: 

- Degree 6 for C2 smoothing is correct and the minimum. In OCCT 7.9 the Jacobi polynomial constructor computed its internal degree as WorkDegree minus 6 for C2 and never checked the sign, so degree 5 produced an internal degree of -1 that happened to work. OCCT 8 added an explicit check that throws "WorkDegree too small for given ConstraintOrder". The kernel passes DegMax straight through as the work degree, so 6 is the floor.
- IntTools_FClass2d: Two differences exist. IntTools has a more careful constructor, handling vertex tolerances, degenerate edges, and bad wires.  And in the fallback used when the polygon test is ambiguous, the old class passes the tolerance as a raw UV distance while IntTools converts it through the surface's U and V resolution, treating it as a 3D tolerance.

It looks like this PR is ready to go - thanks!

tests/test_direct_api/test_constrained_arcs.py
```diff
@@ -504,6 +506,11 @@ def test_tan3_4():
     assert len(tan3) == 0
 
 
+@pytest.mark.xfail(
+    platform.system() == "Linux" and platform.machine() == "aarch64",
+    reason="the RadiusArc case leaves two half circles, so there is no shorter "
+    "or longer sagitta to choose and the kernel's pick differs on Linux arm64",
+)
 def test_make_constrained_arcs_3tan():
```
jdegenstein, line 514: should the test be instead adjusted to reduce the ambiguity of the solution?

jdegenstein (commented):
Added one comment on the xfail test

PR #1443, opened 2026-09-07
Guard unnamed STEP component imports

### LINKED ISSUE #1433: import_step segfaults on STEP assemblies containing an unnamed component
## Summary

`import_step` **segfaults** (SIGSEGV, exit 139) on any STEP file containing an assembly component
whose label has no `TDataStd_Name` attribute. It is not an exception, so callers cannot defend
against it with `try`/`except` — the interpreter dies.

Reproduced on `build123d` 0.11.1 (current release, and current `main` has the same code).

## Reproducer

The public NIST MBE PMI conformance model `nist_ctc_02_asme1_ap242-e2.stp` (from the NIST "MBE PMI
Validation and Conformance Testing Project" AP242 test set) triggers it:

```python
from build123d import import_step
part = import_step("nist_ctc_02_asme1_ap242-e2.stp")
```

```
Fatal Python error: Segmentation fault

Current thread 0x00000001f3c4e180 (most recent call first):
  File ".../build123d/importers.py", line 152 in get_name
  File ".../build123d/importers.py", line 236 in build_assembly
  File ".../build123d/importers.py", line 226 in build_assembly
  File ".../build123d/importers.py", line 260 in import_step
```

The file itself is fine — a plain `STEPControl_Reader` reads it without complaint and yields 637
faces.

## Cause

`importers.py` (line numbers from 0.11.1; `main` is the same code at 146–150):

```python
def get_name(label: TDF_Label) -> str:
    """Extract name and format"""
    name = ""
    std_name = TDataStd_Name()
    if label.FindAttribute(TDataStd_Name.GetID_s(), std_name):   # <-- crashes here
        name = TCollection_AsciiString(std_name.Get()).ToCString()
```

`TDF_Label.FindAttribute` in OCP segfaults when the requested attribute is **absent**. I filed that
separately as CadQuery/OCP#225, with a six-line reproducer that needs no STEP file. It is the same
binding already reported as unreliable in CadQuery/OCP#153 (open) and CadQuery/OCP#55 (closed),
where the *attribute-present* path returns an unpopulated handle.

`import_step` calls `get_name` on every assembly component label, so it only takes one unnamed
component to kill the process. I walked the label tree of `ctc_02`: the offending component carries
`TNaming_NamedShape`, `TDataStd_UAttribute`, `TDF_TagSource` and `TDataStd_TreeNode` — but no
`TDataStd_Name`. Twelve of the seventeen NIST AP242 files are assemblies and load fine; they simply
happen to name every component. So this is not specific to that model, it is any STEP file with an
unnamed assembly component.

## Suggested fix

`label.IsAttribute(guid)` returns `False` safely on the exact label that crashes `FindAttribute`, so
guarding is enough:

```python
    std_name = TDataStd_Name()
    if label.IsAttribute(TDataStd_Name.GetID_s()) and label.FindAttribute(
        TDataStd_Name.GetID_s(), std_name
    ):
        name = TCollection_AsciiString(std_name.Get()).ToCString()
```

Alternatively, the `TDF_AttributeIterator` reimplementation suggested in CadQuery/OCP#153 avoids
`FindAttribute` altogether. I verified `TDF_AttributeIterator` enumerates the crashing label's four
attributes without incident, so it is safe on this input too.

Given that OCP#55 was closed in 2021 and the same method is still unreliable, guarding in
`build123d` rather than waiting on the binding seems worthwhile regardless of what happens upstream.

## Environment

- `build123d` 0.11.1
- `cadquery-ocp-novtk` 7.9.3.1 (OCCT 7.9.3)
- Python 3.14.7
- macOS 15 (Darwin 25.6.0), arm64

atirna:
## Description

I guard the OCP name lookup before accessing the attribute. Previously, an unnamed assembly component could terminate `import_step`; now it imports with the existing empty-label behavior.

Fixes #1433

## Why

STEP assemblies may omit component names, which should not terminate the interpreter.

## Verification

- Before: importing `nist_ctc_02_asme1_ap242-e2.stp` exited 139.
- After: the same file imports as a `Compound` and retains an empty label for the unnamed component.
- `python -m pytest tests/test_importers.py -q`

### DIFF
--- a/src/build123d/importers.py
+++ b/src/build123d/importers.py
@@ -147,7 +147,9 @@ def import_step(filename: PathLike | str | bytes) -> Compound:
         """Extract name and format"""
         name = ""
         std_name = TDataStd_Name()
-        if label.FindAttribute(TDataStd_Name.GetID_s(), std_name):
+        if label.IsAttribute(TDataStd_Name.GetID_s()) and label.FindAttribute(
+            TDataStd_Name.GetID_s(), std_name
+        ):
             name = TCollection_AsciiString(std_name.Get()).ToCString()
         # Remove characters that cause ocp_vscode to fail
         clean_name = "".join(ch for ch in name if unicodedata.category(ch)[0] != "C")
--- a/tests/test_importers.py
+++ b/tests/test_importers.py
@@ -1,9 +1,10 @@
-from io import StringIO
+from io import BytesIO, StringIO
 import os
 from os import fsencode, fsdecode
 import unittest
 import urllib.request
 import tempfile
+import zipfile
 from pathlib import Path
 from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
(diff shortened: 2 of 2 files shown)


atirna:
guarded the name-attribute lookup before calling `FindAttribute`; the AP242 reproducer now imports without terminating Python.

gumyr:
Thank you!

PR #1465, opened 2026-09-19
Write flat shapes into a face's parameter space, and rebuild wrap on it

gumyr:
## Write flat shapes into a face's parameter space, and rebuild wrap on it

This PR adds `UVFrame`, a flat coordinate system laid onto a face, and rebuilds `Face.wrap` and `wrap_faces` on top of it. A planar Edge, Wire or Face drawn on `Plane.XY` is written into the face's (u, v) domain and rebuilt as edges carrying pcurves on the face's own surface, so the result lies on the surface by construction rather than being fitted in 3D, and combines with the face's owner in booleans.

It is the foundation for writing sheet-metal features onto bends, which will follow on the `sheet_metal` branch.

Branched from `matrix_vector`, so it carries that commit (Matrix construction from any affine values and `Vector.transform`) until that PR merges; the frame needs `Matrix.svd()` and the copy fix added here.

### The frame

```python
frame = face.uv_frame(face.location_at(0.5, 0.5, x_dir=(0, 0, 1)))
edge  = frame.write_edge(Edge.make_line((0, 0), (3, 3)))
wire  = frame.write_wire(Rectangle(8, 3).face().outer_wire())
faces = frame.write_face(Text("Hi", 4).faces()[0])   # ShapeList, one per period reached
punched, *islands = frame.punch(Circle(2).face())    # the face with the outline cut out
```

The flat origin lands on the location's position and flat x runs along its x direction. The map behind the frame is chosen from the face's geometry, and is exact wherever the surface allows it:

| surface | map |
|---|---|
| plane, cylinder | a `Matrix`; lines stay lines, circles and ellipses stay conics, other curves are B-splines with transformed poles |
| cone | the exact sector unrolling about the apex |
| sphere | azimuthal equidistant projection about the origin |
| everything else | the exponential map, integrated along geodesics with scipy's RK45 |

Distances along straight lines from the origin are exact on all of them, and the geodesic map reproduces the closed forms on the surfaces that have one, which the tests check. A map of one's own, a `Matrix` or any callable from flat `(x, y)` to `(u, v)`, can be passed to the `UVFrame` constructor.

### Seams and poles

Seams are handled in parameter space, where they are known exactly. The mapped shape is laid out as a planar face on the uv plane, split along the seam lines it crosses, and each piece is moved by whole periods into the surface's range before it is written. So a shape across a seam comes back as one face per period, as the kernel's own faces do, and a wire gets a vertex where it crosses. A torus is split along both of its seams.

A shape spanning a whole turn or more, or one surrounding a pole of the surface, is refused with a `ValueError` rather than producing a self-overlapping face. Writing centred on a pole, such as a logo on top of a ball, would need a degenerate edge in the written face and is left for a follow-up.

### wrap and wrap_faces

`Face.wrap` now writes through the frame instead of marching across the surface in 3D. Closed wires close by construction, every surface type is supported, and the marching helpers, the spline extend-and-trim, and their imports are gone. Behaviour changes callers will notice:

- A face crossing a seam comes back as a `Shell` of one face per period, which thickens and fuses as a single solid; the face overload is typed `Face | Shell`.
- `extension_factor` is gone; `tolerance` is now the largest 3D error of an interpolated curve, default 1e-4.
- `wrap_faces` no longer mutates its input faces, accepts an empty list, lays the row out from the leftmost point of the set, and centres each face on its point of the path. It gains a `tolerance` parameter.
- Planes, cylinders and cones unroll without distortion. A sphere or torus cannot be flattened, so there a shape is distorted increasingly with distance from the origin; the old marching approach preserved edge lengths instead and paid for it with open wires. The docstrings say to keep the origin near the middle of the shape.

The bicycle tire example runs unchanged, in under two seconds, with its tread faces on the surface to 1e-6.

### Also in this PR

- `Matrix.svd()`, used to find the axes of a conic's affine image, and a fix so that copies of a `Matrix` no longer share the underlying transformation.
- `examples/wrap_gallery.py`: text embossed across a cylinder's seam, a star engraved into a sphere, a label on a cone, a logo on a torus, a pattern on a free-form surface, a punched window, and letters laid along a circle of latitude with `wrap_faces`.

### Tests

- `tests/test_direct_api/test_uv_write.py`: 33 tests over cylinder, plane, cone, sphere, torus and B-spline frames, seam splitting for faces and wires, in-range parameters of the pieces, booleans with thickened pieces, punching, the whole-turn and pole refusals, agreement of the geodesic map with the closed forms, and hand-written maps.
- `test_face.py`: `wrap` on cylinder, sphere, cone and torus with validity, surface gap and normal direction; exact lengths and areas on the developable surfaces; the seam case returning a shell that thickens and fuses; `wrap_faces` leaving its inputs alone.
- `Matrix.svd` and copy tests in `test_matrix.py`.

Direct-API suite, pylint, mypy and black are clean.

### Follow-ups

- Degenerate edges, so that a shape can cover a pole.
- Sheet metal: replace the flat frame in `operations_sheet.py` with `UVFrame` once `dev` is merged into `sheet_metal`.
- A docs page for wrapping.


seltzered:
Thank you! Will try to revisit & play with this hopefully in the coming weeks.

